PR #500, opened 2023-09-06
prune and inline styles for performance

gebhardtr:
TODO:
- [x] invoke at build time
- [ ] (future enhancement) can inject the style directly into `base.html` for even better performance
- [x] ensure `npx`  & `jq` are installed
- [x] prevent patternfly.css from being loaded in the image
- [x] confirm prune-css is better than Critical (and its variants); I like it so far
- [ ] (future enhancement) can optimize the `echo`s away in favor of strictly using `jq`
- [ ] (future enhancement) prune css for each page and include a unique `style` element
- [ ] ensure this doesn't break the new React pages

Reviewer's TODO:
- [ ] ensure all pages still render properly
 

CSS size reduction: 1.4MB -> 80KB

Lots of CSS being loaded that isn't critical/used:
image

And inlining reduces the render-blocking `link` to improve performance:
image

image


robinbobbitt (changes requested):
The icons on the home page don't load correctly with this change. It looks like a path issue, where it's now trying to load e.g. from http://localhost:8000/assets/fonts/webfonts/fa-solid-900.woff rather than http://localhost:8000/static/assets/fonts/webfonts/fa-solid-900.woff (note missing "static" in the former).

PR #512, opened 2023-09-07
SSO: remove our local version of get_username

goneri:
We won't expose anymore the internal username to the user. As a consequence,
we don't need this extra logic to try to reuse the one coming from Github.


manstis (approved):

robinbobbitt (commented):
I'm not so sure about removing this. Agree it feels less important now because we don't expose their django username, but it still makes sense to me that you are the same user whether you log in with GitHub Team identity provider or vanilla GitHub. I'm concerned we might regret removing this later if we end up needing to do something more with the GitHub users. Thoughts?

Edited to add: I converted this to a draft for now to avoid anybody merging til we decide on this.

PR #594, opened 2023-10-03
Upgrade dependencies to pull in newer version of urllib.

manstis:
```
--- | --- | --- | --- | ---
urllib3 | 1.26.16 | GHSA-v845-jxx5-vc9f | 1.26.17,2.0.6 | urllib3 doesn't treat the `Cookie` HTTP header special or provide any helpers for managing cookies over HTTP, that is the responsibility of the user. However, it is possible for a user to specify a `Cookie` header and unknowingly leak information via HTTP redirects to a different origin if that user doesn't disable redirects explicitly.  Users **must** handle redirects themselves instead of relying on urllib3's automatic redirects to achieve safe processing of the `Cookie` header, thus we decided to strip the header by default in order to further protect users who aren't using the correct approach.  ## Affected usages  We believe the number of usages affected by this advisory is low. It requires all of the following to be true to be exploited:  * Using an affected version of urllib3 (patched in v1.26.17 and v2.0.6) * Using the `Cookie` header on requests, which is mostly typical for impersonating a browser. * Not disabling HTTP redirects * Either not using HTTPS or for the origin server to redirect to a malicious origin.  ## Remediation  * Upgrading to at least urllib3 v1.26.17 or v2.0.6 * Disabling HTTP redirects using `redirects=False` when sending requests. * Not using the `Cookie` header.
```

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,30 +6,32 @@
 #
 black==22.10.0
     # via -r requirements-dev.in
-build==0.10.0
+build==1.0.3
     # via pip-tools
-cfgv==3.3.1
+cfgv==3.4.0
     # via pre-commit
-click==8.1.3
+click==8.1.7
     # via
     #   black
     #   pip-tools
 coverage==7.2.1
     # via -r requirements-dev.in
-distlib==0.3.6
+distlib==0.3.7
     # via virtualenv
 enum-compat==0.0.3
     # via torch-model-archiver
-filelock==3.12.0
+filelock==3.12.4
     # via virtualenv
 flake8==6.0.0
     # via -r requirements-dev.in
-grpcio==1.54.2
+grpcio==1.59.0
     # via grpcio-tools
 grpcio-tools==1.54.2
     # via -r requirements-dev.in
-identify==2.5.24
+identify==2.5.30
     # via pre-commit
+importlib-metadata==6.8.0
+    # via build
 isort==5.10.1
     # via -r requirements-dev.in
 mccabe==0.7.0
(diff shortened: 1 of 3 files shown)


goneri (changes requested):
You can actually just set `urllib3==1.26.17` in the `requirements.in`. It should be enough to address the issue.

```
(venv) [root@f8122628be05 www]# pip-audit
Found 1 known vulnerability in 1 package
Name    Version ID                  Fix Versions
------- ------- ------------------- -------------
urllib3 1.26.16 GHSA-v845-jxx5-vc9f 1.26.17,2.0.6
(venv) [root@f8122628be05 www]# pip install urllib3>=1.26.17
WARNING: You are using pip version 21.2.3; however, version 23.2.1 is available.
You should consider upgrading via the '/var/www/venv/bin/python3 -m pip install --upgrade pip' command.
(venv) [root@f8122628be05 www]# pip install --upgrade urllib3>=1.26.17
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
opensearch-py 2.1.1 requires urllib3<2,>=1.21.1, but you have urllib3 2.0.6 which is incompatible.
botocore 1.29.143 requires urllib3<1.27,>=1.25.4, but you have urllib3 2.0.6 which is incompatible.
WARNING: You are using pip version 21.2.3; however, version 23.2.1 is available.
You should consider upgrading via the '/var/www/venv/bin/python3 -m pip install --upgrade pip' command.
(venv) [root@f8122628be05 www]# pip install --upgrade urllib3==1.26.17
Collecting urllib3==1.26.17
  Downloading urllib3-1.26.17-py2.py3-none-any.whl (143 kB)
     |████████████████████████████████| 143 kB 3.0 MB/s 
Installing collected packages: urllib3
  Attempting uninstall: urllib3
    Found existing installation: urllib3 2.0.6
    Uninstalling urllib3-2.0.6:
      Successfully uninstalled urllib3-2.0.6
Successfully installed urllib3-1.26.17
WARNING: You are using pip version 21.2.3; however, version 23.2.1 is available.
You should consider upgrading via the '/var/www/venv/bin/python3 -m pip install --upgrade pip' command.
(venv) [root@f8122628be05 www]# pip-audit
No known vulnerabilities found
```

Note: Post-GA, it would be nice to bump most of this dependencies to the last version available.

PR #652, opened 2023-10-16
sonar: ignore the ansible_wisdom/users/migrations/* files

goneri:
The files are only used to upgrade the DB and don't need any test coverage.


manstis (approved):
Looks remarkably like https://github.com/ansible/ansible-wisdom-service/pull/650

:smile_cat:

PR #711, opened 2023-11-16
ams: add the mocker driver

goneri:
The mocker driver uses a static configuration that it gets from the settings.

e.g:

```bash
export AUTHZ_BACKEND_TYPE="mocker"
export AUTHZ_MOCKER_USERS_WITH_SEAT=gleboude1@redhat.com
export AUTHZ_MOCKER_RH_ORG_ADMINS=gleboude1@redhat.com
export AUTHZ_MOCKER_ORGS_WITH_SUBSCRIPTION=11009103
```  

The goal is to avoid any interaction with a third-party backend. The commit changes two parts:
 
- add a new the `mocker` backend type
- and modify the social_core pipeline

# Mocker backend type

It replaces a real backend and generates the answers from the content of the `AUTHZ_MOCKER_*` setting keys.

# Pipeline

This change slightly modify the way the `User.rh_user_is_org_admin` is set.
When the `AUTHZ_BACKEND_TYPE` environment variable is set, we check if the
user is listed in `AUTHZ_MOCKER_RH_ORG_ADMINS`.

Finally, since we need the `external_username` field early in the process, we
now read it from the JWT payload. This allow us to simplify the pipeline.


manstis (commented):
LGTM :+1: 

Just one question around the invocation of `load_extra_data` I don't understand :smile:

manstis (approved):
LGTM :+1:

PR #726, opened 2023-11-29
AAP-18625: Completions fails with 503 when user specifies model 'fred'

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-18625

## Description
Additional response handlers for the newly described scenario.

### Steps to test
1. Pull down the PR
2. Run `wisdom-service`
3. Make a Completions request with "fred" as the `model_id`.
4. Make a _contentmatch_ request with "fred" as the `model`.

### Scenarios tested
I have confirmed the fix using Swagger.

## Production deployment


robinbobbitt (changes requested):

robinbobbitt (approved):
Awesome, thanks @manstis !

PR #728, opened 2023-11-29
flake8: ignore two tests Superfluous because of Black

goneri:
The two errors are already avoid thanks tothe Black formatter.

- E231 missing whitespace after ':'
- E221 multiple spaces before operator

The remaining errors are only false positives.


manstis (approved):
LGTM :+1:

PR #732, opened 2023-11-30
mesh_client: adjust the name of the fake backends

goneri:
https://github.com/ansible/ansible-wisdom-service/pull/720 harmonized the way we name our fake backends.

- Dummy is for the fake backend that can be instanced during the run time. We enable them
  setting the value `dummy` for the associated `*BACKEND_TYPE` in the configuration.
- Anything called Mock or Mocked should only be used in the unit-tests

Also, to address a SonarCloud error, the PR adjusts the code of the
DummyClient to avoid the use of random.random().
Note: The original unit-test was wrong, the value returned by `random.random()`
is a float between `0` and `1`, not `10`.

### DIFF
--- a/ansible_wisdom/ai/api/model_client/dummy_client.py
+++ b/ansible_wisdom/ai/api/model_client/dummy_client.py
@@ -1,6 +1,6 @@
 import json
 import logging
-import random
+import secrets
 import time
 
 import requests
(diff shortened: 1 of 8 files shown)


manstis (dismissed):
LGTM :+1:

manstis (dismissed):
Minor comment, otherwise LGTM 👍

TamiTakamiya (commented):
LGTM, but I would like @kdelee to review this because I think the mock client has been used in performance tests.

kdelee (commented):
LGTM, though wonder if you would also want to update the settings to have consistent name. i'd say now is better than later when people may have forgotten that DummyClient used to be named different

manstis (approved):
(Still) LGTM :+1:

PR #733, opened 2023-11-30
gha/coverage: stop posting the coverage report after every PR update

goneri:
SonarCloud already provides the same service and these reports can quickly
take a bit too much space in the PR conversation interface.


TamiTakamiya (approved):
LGTM and makes a lot sense. Thank you!!

PR #734, opened 2023-12-01
dummy backend: accept an envvar with an empty string

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-17822>

## Description

Right now, if I set `AUTHZ_DUMMY_ORGS_WITH_SUBSCRIPTION`, I will
end-up with `11009103`. There is no way to get an empty string instead.

This is what the PR address.

## Testing

Use an empty `AUTHZ_DUMMY_ORGS_WITH_SUBSCRIPTION` environment variable.

## Production deployment


manstis (dismissed):
LGTM :+1:

TamiTakamiya (changes requested):
I think the line 75 is not needed. Pls take a look at my comment there.

TamiTakamiya (approved):
LGTM

PR #756, opened 2023-12-11
[AAP-18621] minimize use of lightspeed service by deactivated users

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-18621>

## Description

Check if the user is still active before granting the seat.

### Steps to test
1. Pull down the PR
2. ... Run the test-suite or disable a user
3. ...

### Scenarios tested

An active user has a token VSCode and can use the service, it then get removed from the RH Org.

## Production deployment

- [x] his code is ready for product

### DIFF
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -4,13 +4,9 @@ on:
   push:
     branches:
       - main
-    paths:
-      - ansible_wisdom/ai/api/aws/wca_secret_manager.py
   pull_request:
     branches:
       - main
-    paths:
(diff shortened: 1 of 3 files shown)


manstis (dismissed):
Minor comment for your consideration, otherwise LGTM :+1:

robinbobbitt (commented):
I've added a few questions to the code, and also am wondering about the users federated from their company login. I assume we need a different solution for them? Is there an option to check if they're federated and if so set a lower expiry on them (e.g. 24 hours)?

PR #760, opened 2023-12-12
rename settings keys associated with dummy backends

goneri:
Remove `MOCK_WCA_SECRETS_MANAGER`, you should now use the following configuration instead:
`WCA_SECRET_BACKEND_TYPE="dummy"`

Extend the `WCA_SECRET_DUMMY_SECRETS` to be able to set both the key and the model_id at the same time using the following syntax:
`WCA_SECRET_DUMMY_SECRETS='123:my-key<|sepofid|>sec'`

Also rename `DUMMY_MODEL_RESPONSE_BODY` to be consistent with the new name of the associated backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,12 +76,17 @@ export SECRET_KEY=somesecretvalue
 ```
 
 For most development usages, you can skip the call to AWS Secrets Manager
-and always use the `ANSIBLE_WCA_FREE_API_KEY`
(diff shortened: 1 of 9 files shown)


manstis (commented):
A couple of comments; but looking good.

manstis (dismissed):
LGTM :+1: Thanks.

manstis (approved):
LGTM :+1:

PR #783, opened 2024-01-15
Ensure organizations_organization is populated for existing Users.

manstis:

## Description
Following https://issues.redhat.com/browse/AAP-15464

The `organizations_organization` record is only created when the User logs in.

However, with newer container images, the table is used to determine the Users' `organization_id`.

Therefore, for Users who remain logged in after https://github.com/ansible/ansible-wisdom-service/pull/718 is rolled out, the `organizations_organization` table does not contain a record for their Organization and `wisdom-service` returns `None`. This leads to erroneous behaviour of the system for logged in Users.

This migration ensures the database is updated for existing Users.

## Testing
1. Before pulling down this PR.
2. `make stop-backends`
3. `make start-backends`
4. `make create-application`
5. `docker exec -it <postgress-docker-container-id> bash`
6. In the docker container: `psql -U wisdom`
7. `select * from users_user where fk_organization_id is null;`. There _may be_ records.
8. Pull down this PR
9. `make create-application`
10. `select * from users_user where fk_organization_id is null;`. There should be none.
11. `select * from users_user where fk_organization_id not in (select id from organizations_organization)`. There should be none.

### Scenarios tested
The above!

## Production deployment

### DIFF
--- a/ansible_wisdom/organizations/migrations/0002_user_organization_fix_id.py
+++ b/ansible_wisdom/organizations/migrations/0002_user_organization_fix_id.py
@@ -1,0 +1,27 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [('organizations', '0001_initial'), ('users', '0009_user_organization')]
+
+    operations = [
+        migrations.RunSQL(
+            """
+            insert into
+            organizations_organization
+            select
+            organization_id, false
+            from
+            users_user
+            where
+            organization_id is not null
+            on conflict (id) do nothing;
+            update
+            users_user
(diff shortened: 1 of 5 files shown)


robinbobbitt (dismissed):
I tested the sql against my local db and am comfortable it's proper. One note for your test step 10... It will only be none if you don't have any non-RHSSO users. I approve but as discussed in Slack will leave to Michael to decide between merging as-is and getting the unit test in place depending on how long it will take, as this PR loses value as time passes.

jameswnl (approved):
awesome!

PR #785, opened 2024-01-15
Upgrade jinja2 to 3.1.2

manstis:

```
--- | --- | --- | --- | ---
jinja2 | 3.1.2 | GHSA-h5c8-rqwp-cp95 | 3.1.3 | The `xmlattr` filter in affected versions of Jinja accepts keys containing spaces. XML/HTML attributes cannot contain spaces, as each would then be interpreted as a separate attribute. If an application accepts keys (as opposed to only values) as user input, and renders these in pages that other users see as well, an attacker could use this to inject other attributes and perform XSS. Note that accepting keys as user input is not common or a particularly intended use case of the `xmlattr` filter, and an application doing so should already be verifying what keys are provided regardless of this fix.
```

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -33,6 +33,10 @@ ipython==8.10.0
 # remove this once django-oauth-toolkit is updated to properly
 # pull a version of jwcrypto >= 1.5.1.
 jwcrypto==1.5.1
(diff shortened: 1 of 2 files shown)


hasys (approved):
Looks good, thank you @manstis

PR #787, opened 2024-01-18
repo_url no longer included in codematch response

robinbobbitt:
Jira Issue: N/A

## Description
Since 6 AM EST, contentmatch has been broken because WCA codematch response no longer includes a repo_url. TBD if this is by design or a bug, but this code change prevents the service from failing on a missing repo_url. Note it's missing from codematch response, but once it hits our serializer, it's an empty string due to https://github.com/ansible/ansible-wisdom-service/blob/8bca96cbedd6f9df4f534eab776bf6873537fb08/ansible_wisdom/ai/api/data/data_model.py#L44

### Steps to test
1. Pull down the PR
2. Run the wisdom-service
3. Test a contentmatch against prod WCA. Confirm it works.

### Scenarios tested
Unit test plus manually validated with VSCode extension. Content match works and URL is displayed as empty.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/serializers.py
+++ b/ansible_wisdom/ai/api/serializers.py
@@ -440,7 +440,7 @@ class AttributionResponseSerializer(serializers.Serializer):
 
 class ContentMatchSerializer(serializers.Serializer):
     repo_name = serializers.CharField()
-    repo_url = serializers.URLField()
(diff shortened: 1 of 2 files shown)


mabulgu (approved):
LGTM

PR #788, opened 2024-01-22
sessions expire after 14h

goneri:
## Description
Match the duration of the session with Red Hat SSO.

See: https://issues.redhat.com/browse/AAP-18621

## Testing

Log-in, wait 14h

### Steps to test

Start the service, log-in with VSCode and check the token in the DB:

```
$ podman exec --user=root -it docker-compose_django_1 wisdom-manage shell
In [1]: from oauth2_provider.models import AccessToken

In [2]: for t in AccessToken.objects.all():
   ...:     print(f"user={t.user} expires={t.expires}")
   ...: 
user=gleboude1@redhat.com expires=2024-01-23 03:26:39.381533+00:00
```

The token should not last more than 14h

## Production deployment

### DIFF
--- a/ansible_wisdom/main/settings/base.py
+++ b/ansible_wisdom/main/settings/base.py
@@ -176,8 +176,8 @@ OAUTH2_PROVIDER = {
         'code-oss',
         'checode',
     ],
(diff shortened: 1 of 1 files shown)


TamiTakamiya (approved):
LGTM. Thanks!

PR #789, opened 2024-01-22
disable tech preview in wisdom-service

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-19427>

## Description

Introduce the new `ANSIBLE_AI_ENABLE_TECH_PREVIEW` configuration key
that can be use to enable or disable the Tech Preview mode.
By default, the key is enable since the Tech Preview is still supported.

The commit also do the following changes when the Tech Preview is disabled:

- Block access to non seated users to the Completion and Attribution end points
- Return specific error to the VSCode extension when a non-seated user get blocked
- Display clear messages regarding the status of the user

## Testing

The PR is fully unit-tested, you can also set `ANSIBLE_AI_ENABLE_TECH_PREVIEW` to `False` to reproduce a situation where the Tech Preview is disabled.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/permissions.py
+++ b/ansible_wisdom/ai/api/permissions.py
@@ -1,5 +1,6 @@
 from ai.api.aws.wca_secret_manager import Suffixes
 from django.apps import apps
+from django.conf import settings
 from rest_framework import permissions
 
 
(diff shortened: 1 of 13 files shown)


TamiTakamiya (commented):
I left two comments. Would you take a look?

robinbobbitt (changes requested):
Looking good! I've left a few comments. An additional gap to call out is that new unseated users are still getting the tech preview terms of service presented when tech preview is off. Please take a look.

robinbobbitt (changes requested):
@goneri please note my previous comment about tech preview terms still being presented. Thanks!

manstis (commented):
A couple of minor comments.. they don't _break_ the PR; but may make nicer reading.

manstis (dismissed):
LGTM :+1:

manstis (dismissed):
LGTM 👍 

Is there a JIRA to remove the `user.community_terms_accepted` field (and all the `ANSIBLE_AI_ENABLE_TECH_PREVIEW` flag usage etc) when Tech Preview ends?

IDK if it'll be easier to track what needs refactoring now, as theses changes are added, than hunting for them later. Your choice 👍

robinbobbitt (changes requested):
We are so close! Just one concern noted. Thanks @goneri !

robinbobbitt (changes requested):

robinbobbitt (approved):

PR #794, opened 2024-01-24
Exclude GHSA-wj6h-64fc-37mp

manstis:

Fix two `pip-audit` issues:

```
--- | --- | --- | --- | ---
ecdsa | 0.18.0 | GHSA-wj6h-64fc-37mp |  | python-ecdsa has been found to be subject to a Minerva timing attack on the P-256 curve. Using the `ecdsa.SigningKey.sign_digest()` API function and timing signatures an attacker can leak the internal nonce which may allow for private key discovery. Both ECDSA signatures, key generation, and ECDH operations are affected. ECDSA signature verification is unaffected. The python-ecdsa project considers side channel attacks out of scope for the project and there is no planned fix.

pillow | 10.0.1 | GHSA-3f63-hfp8-52jq | 10.2.0 | Pillow through 10.1.0 allows PIL.ImageMath.eval Arbitrary Code Execution via the environment parameter, a different vulnerability than CVE-2022-22817 (which was about the expression parameter).
```
**Notes**
- `GHSA-wj6h-64fc-37mp` is ignored
 > ....The python-ecdsa project considers side channel attacks out of scope for the project and there is no planned fix.
 
 - `pillow` is upgrade to `10.2.0`

### DIFF
--- a/.github/workflows/pip_audit.yml
+++ b/.github/workflows/pip_audit.yml
@@ -43,3 +43,4 @@ jobs:
             PYSEC-2023-228
             GHSA-mq26-g339-26xf
             PYSEC-2022-43059
+            GHSA-wj6h-64fc-37mp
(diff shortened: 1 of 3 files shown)


TamiTakamiya (approved):
LGTM. Thanks!

PR #805, opened 2024-02-02
podman-compose: fix the name of the wisdom-service image

goneri:
When we only have the `build` key set, the image that is set to the
newly created image doesn't match the expectation.
By fixing the name of the local image, we avoid a problem.

From https://docs.docker.com/compose/compose-file/compose-file-v3/

> If you specify image as well as build, then Compose names the
> built image with the webapp and optional tag specified in image:


tomershinhar (approved):
LGTM
confirmed this is working locally using docker

PR #806, opened 2024-02-02
disable seat assignment check from wisdom service

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-19864>

## Description

This change introduce a Feature flag `has_seat_support` that can be used to
disable the seat check and always return `True`.
It also declares a new configuration key called `ENABLE_SEAT_SUPPORT_DEFAULT_VALUE`
that define what should be the default value for the new feature flag above.

The goal is to do the transition following these steps:

1. Define the `has_seat_support` Feature flag has `True`
2. Set it to `False` for a specific user
3. Validate the behaviour for the user
4. Switch it to `False` by default
5. Change the `ENABLE_SEAT_SUPPORT_DEFAULT_VALUE` envar to `False` to lock the behaviour

Once this is done, another commit will follow to remove the code associated with the Seat management.

### Scenarios tested

Switch `ENABLE_SEAT_SUPPORT_DEFAULT_VALUE` to `False` or create the `has_seat_support` FF.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/permissions.py
+++ b/ansible_wisdom/ai/api/permissions.py
@@ -53,7 +53,7 @@ class BlockUserWithoutSeatAndWCAReadyOrg(permissions.BasePermission):
 
     def has_permission(self, request, view):
         user = request.user
-        if user.organization is None:
+        if settings.ANSIBLE_AI_ENABLE_TECH_PREVIEW and user.organization is None:
             # We accept the Community users, the won't have access to WCA
             return True
         if user.rh_user_has_seat is True:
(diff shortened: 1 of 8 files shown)


robinbobbitt (changes requested):

tomershinhar (dismissed):
LGTM
we need to set ENABLE_SEAT_SUPPORT_DEFAULT_VALUE to false for  the test users though before i can merge the test updates

robinbobbitt (changes requested):
@goneri please see my comment re: updating `rh_user_has_seat` to be true for commercial users and false for tech preview users. I think that is the quickest and safest path forward.

PR #819, opened 2024-02-09
compose: ensure we use PostgreSQL 15, like in production

goneri:
`docker.io/library/postgres:alpine` is actually pointing on PostgreSQL 16 which
is not the version we run in production.

By pinning on `15-alpine` we reduce the delta with our target environment and
reduce the risk of introducing some breaking changes in the future.


manstis (approved):
LGTM :+1: 

Verified tag [here](https://github.com/docker-library/docs/blob/master/postgres/README.md).

PR #820, opened 2024-02-09
tools: add scripts to dump and restore the database

goneri:
Add two scripts that can be used to save and restore a dev environment.

The goal was to be able to quickly restart `podman-compose` and restart from
a previous backup.


manstis (commented):
Sorry @goneri I'm happy to approve; I understand _what_ you're doing.

However I don't quite understand the _how_ you're doing it.

manstis (approved):
Thanks for the explanations @goneri

PR #822, opened 2024-02-09
requirements: social-auth-app-django==5.4.0

goneri:
This partially revert change f17282e464463a6973a76983094cc2e6973b786a.

Because we were depending on an unreleased fix, we had to pin on a Git version
of socal-app-django. Since a release has been done recently with the fix, we
can now move on and update the `requirements.txt`.


manstis (approved):
LGTM 👍

PR #833, opened 2024-02-14
tests: add missing setUp()/tearDown() calls

goneri:
Class Constructors are supposed to call the upper `setUp()` methods. It's the same for the `tearDown()`.


manstis (approved):
LGTM :+1: 

Nice.

PR #847, opened 2024-02-28
Containerfile: use two targets to isolate the dev specific steps

goneri:
Jira Issue: https://issues.redhat.com/browse/AAP-21218

Use two different targets (https://docs.docker.com/compose/compose-file/build/#target) to isolate the steps that are specific to the development environment.

See: https://github.com/ansible/ansible-wisdom-service/pull/845


jameswnl (changes requested):
```sh-5.1$ ls -l /etc/supervisor/supervisord.d/*.conf
-rw-r--r--. 1 root root 369 Feb 29 14:49 /etc/supervisor/supervisord.d/auto-reload.conf
```

This file is there in the pr deployment (using the pr build image which should be excluding it, right?)

jameswnl (approved):

PR #871, opened 2024-03-11
users: reduce the size of the err message for Org with no subscription

goneri:
This commit removes a duplicated sentence in the error message.


manstis (approved):
LGTM :+1:

PR #874, opened 2024-03-13
AAP-21173: Integrate custom models for telemetry

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-21173>

## Description
Adding a new `model` field for the feedback request, which comes from the [VSCode request](https://issues.redhat.com/browse/AAP-21329), see [PR here](https://github.com/ansible/vscode-ansible/pull/1131), and propagating it into telemetry events.
It applies for both:
- `schema1` => All feedback events feedback events
- `schema2` =>  Just adding the model field into the `Product Feedback` (sentiment), as it is already present for `Recommendation Generated`, which correlates with the `Recommendation Action` one.

## Testing
1. Pull down the PR
2. ...
3. ...

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/model_client/base.py
+++ b/ansible_wisdom/ai/api/model_client/base.py
@@ -17,13 +17,12 @@ class ModelMeshClient:
     def set_inference_url(self, inference_url):
         self._inference_url = inference_url
 
-    @abstractmethod
     def get_model_id(
         self,
         organization_id: Optional[int] = None,
         requested_model_id: str = '',
     ) -> str:
-        raise NotImplementedError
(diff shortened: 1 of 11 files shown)


manstis (commented):
Some thoughts.

manstis (dismissed):
LGTM :+1: 

You could have made the _default_ implementation of `get_model_id` on `ModelMeshClient` concrete, as follows, to avoid the repeated code; but that's a detail..
```
def get_model_id(
        self,
        organization_id: Optional[int] = None,
        requested_model_id: str = '',
    ) -> str:
        return requested_model_id or settings.ANSIBLE_AI_MODEL_NAME
```

manstis (approved):
LGTM :+1:

PR #877, opened 2024-03-15
upgrade to Pydantic 2.6

goneri:
We were using a rather old version of Pydantic that is conflicting with
langchain.

Pydantic 1 will be EOL at the end of June: https://docs.pydantic.dev/latest/version-policy/


manstis (approved):
LGTM 👍

PR #878, opened 2024-03-15
API: add the explanation endpoint

goneri:
Use Ollama to process the answer.


manstis (commented):
More enquiry than review 😺

PR #920, opened 2024-04-04
multitasks: also convert the prompt to low-case

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-21078>

Follow up to af0b1b757e1463ef7958baa3fc29b9a46a1c4d75, our single-task prompts
are always lower-case. However it was not the case for the multi-tasks prompts.

- Fully covered by unit-tests


manstis (dismissed):
LGTM :+1:

robinbobbitt (approved):
Works but I do wonder if we should rename the function since it's pulling double duty now?

PR #926, opened 2024-04-09
AAP-22474: [PoC] Custom SKU checks, for AAP subscribed orgs

romartin:
Hey @robinbobbitt @jameswnl 

Just a DRAFT / PoC to brainstorm about the custom AMS / SKU checks. There are TODOs, refactoring, tests and maybe some logic still in progress.

See:
- https://issues.redhat.com/browse/AAP-21621
- https://issues.redhat.com/browse/AAP-22474

Use ENV: `AMS_ACTIVE_SKU_LIST=MCT3691,MCT3692,MCT3693,MCT3694,MCT3695,MCT3696,MCT4164`

@jameswnl I checked for "onprem" deployment mode, and this check is not being called.

Thanks!


robinbobbitt (commented):
General approach looks good to me. We should confirm every assumption we're making about resource_quota, as discussed in Slack. I think we should also cache the result for some amount of time once we determine the org does have a subscription.

PR #927, opened 2024-04-10
pyproject: depends on langchain 0.1.10

goneri:
Depend on an older version of langchain to avoid pulling `orjson` via `langsmith`. `orjson` requires Rust 1.72, which is not yet avaialble yet in RHEL8.


jameswnl (commented):
how about https://github.com/ansible/ansible-wisdom-service/blob/main/requirements.in#L40 ?

jameswnl (approved):

dsavineau (approved):

PR #939, opened 2024-04-15
reformatting to avoid an endless duel between Black and Flake8

goneri:
Minor adjustment to be sure the code, as reformated by Black is Flake8 compliant.

```
ansible_wisdom/users/models.py:100:89: E272 multiple spaces before keyword
ansible_wisdom/users/signals.py:60:54: E251 unexpected spaces around keyword / parameter equals
ansible_wisdom/users/signals.py:60:56: E251 unexpected spaces around keyword / parameter equals
ansible_wisdom/users/signals.py:60:59: E202 whitespace before '}'
1     E202 whitespace before '}'
2     E251 unexpected spaces around keyword / parameter equals
1     E272 multiple spaces before keyword
4
```

### DIFF
--- a/ansible_wisdom/users/models.py
+++ b/ansible_wisdom/users/models.py
@@ -96,10 +96,11 @@ class User(ExportModelOperationsMixin('user'), AbstractUser):
         """
 
         if self.organization and self.organization.is_subscription_check_should_be_bypassed:
-            logger.info(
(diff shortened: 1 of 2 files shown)


manstis (approved):
Don't ever expect the battle between competing `lint`'ers to end!

PR #943, opened 2024-04-16
[DRAFT] AAP-16888: I want to reduce time to resolve customer issues through better error messages

manstis:
**=== DRAFT ===**

Jira Issue: https://issues.redhat.com/browse/AAP-16888

## Description
This PR adds an `error_context_id` attribute to the API response payload for exceptions.

The full stack trace and `error_context_id` are logged on the server.

The additional `error_context_id` field is sent to VSCode.... where we can either log it or show it.

The purpose is to support correlation of exceptions observed in VSCode with what happened on the server.

### Steps to test
1. Pull down the PR
2. `make start-backend`
3. `make create-application`
4. Run `wisdom-service` however you would normally.

### Scenarios tested
Unit tests confirming the additional payload field.

## Production deployment
- [x] This code change requires the following considerations before going to production:-
This PR requires changes to VSCode.. to be determined; if the approach in this PR is acceptable.

### DIFF
--- a/ansible_wisdom/ai/api/exceptions.py
+++ b/ansible_wisdom/ai/api/exceptions.py
@@ -13,6 +13,9 @@
 #  limitations under the License.
 
 import json
+import logging
+import uuid
+from typing import Union
 
 from django.conf import settings
 from prometheus_client import Counter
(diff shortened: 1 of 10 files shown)


robinbobbitt (commented):
I believe including a stack trace in a response body is generally considered bad practice as we might dump sensitive information or clues that make us more vulnerable to attack.  For this issue, I think you've already done a ton to improve our error handling, and at this point all that remains might be a quick review of the API calls in the extension for any remaining low hanging fruit.

manstis:
@robinbobbitt I've added a new commit.

This PR now only propagates a `error_context_id` in the response to the client.

It logs this with the stack trace for the exception on the server and includes `error_context_id` in Segment events.

I've not adjusted the `ALLOW_LIST` of fields for Segment events; so it's only included for "non seated users".

IIRC the "seated user" concept is to be removed, so it makes little sense changing the `ALLOW_LIST`?

Personally, I'd include `error_context_id` in all Segment events for all users.. but I know there's more to it than that.

I'd show the `error_context_id` in the "Ansible" / "Ansible Support" panel in VSCode with a suitable message.

manstis:
Closing. JIRA moved to backlog.

The proposal is better provided by using W3C Trace Context.

PR #967, opened 2024-05-01
get_model_client: call get_wca_client() only when wca is actually enabled

goneri:
Ensure `get_model_client()` doesn't return a WCA client when the configuration
points on a different model.


manstis (dismissed):
LGTM :+1: 

I'll let your consciousness wrestle with `Code Coverage / Unit Test and SonarCloud Scan`.

manstis (commented):
A couple of inquisitive comments.

Do we also need a test for `wca-onprem` too?

manstis (approved):
LGTM :+1:

jameswnl (approved):

PR #971, opened 2024-05-01
Revert "verify signature of rhsso access token (#767)"

goneri:
This reverts commit ef407500155b530e634a720be0572274280417f6.

This commit depends on python3-jose which is a dependency we want to break.
In addition, once the authentication step is passed ( [See: find_valid_key()](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py#L192-L201) ), we can safely access
the JWT payload since it has already been checked.

Also a minor adjustment of the social-auth-core dependency to avoid a confict cause by 01045b2e53eaf42cf3546e86310b2f40ed3b31db.

### DIFF
--- a/ansible_wisdom/users/auth.py
+++ b/ansible_wisdom/users/auth.py
@@ -16,7 +16,6 @@ import logging
 
 import jwt
 from django.conf import settings
-from jose import jwk
 from rest_framework import authentication
 from social_core.backends.oauth import BaseOAuth2
 from social_django.models import UserSocialAuth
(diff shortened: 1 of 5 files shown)


manstis (commented):
Before approving; I'd like to better understand when `find_valid_key()` is used.

Can you explain?

manstis (dismissed):
LGTM :+1:

robinbobbitt (changes requested):

robinbobbitt (changes requested):

PR #996, opened 2024-05-09
AAP-23914: Lightspeed Telemetry :: Populate FQCN information, even if post-processing is disabled

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-23914>

## Description
Since the post-processing being disabled by [1], on staging clusters, some telemetry evens are lacking the following information:
- schema1: completion event => missing tasks
- schema2: Recommendation Generated event => collection and module fields are empty

[1] https://github.com/ansible/ansible-wisdom-ops/commit/ce89fb48544e300dc55c9db6969bd06a8b363ba4

### Steps to test
- Disable ARI post-processing => ENABLE_ARI_POSTPROCESS=False
- Run the wisdom-service
- Infer
- Look at either completion event, from schema1, and _Recommendation Generated event, from schema2, both are lacking information, as described above.

### Scenarios tested
Tested local against staging Segment account, when both post-processing enabled and disabled.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/formatter.py
+++ b/ansible_wisdom/ai/api/formatter.py
@@ -396,6 +396,8 @@ def parse_module_from_prediction(re, prediction):
 
 
 def get_fqcn_or_module_from_prediction(prediction):
+    if prediction is None:
+        return None
     fqcn = get_fqcn_from_prediction(prediction)
     if fqcn is None:
         fqcn = get_module_from_prediction(prediction)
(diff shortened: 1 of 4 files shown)


manstis (commented):
I see you've simply moved population of the `fqcn_module` outside of `ari_results` block. Fair enough. Makes sense. Shouldn't you however have a test so that the problematic scenario can be detected ongoing?

manstis (dismissed):
LGTM 👍

robinbobbitt (changes requested):
Thanks Roger! Just a couple small comments.

manstis (approved):
LGTM 👍

robinbobbitt (approved):
LGTM! Thanks.

PR #1004, opened 2024-05-13
Makefile: transition from ansible_wisdom to ansible_ai_wisdom

goneri:
Replace two `ansible_wisdom` imports with the new package name.


manstis (approved):
LGTM 👍

PR #1009, opened 2024-05-13
development: ignore empty WCA_SECRET_BACKEND_TYPE value

goneri:
If `WCA_SECRET_BACKEND_TYPE` is defined but empty, we should use the default
value which is `dummy`.


manstis (approved):
LGTM 👍

PR #1013, opened 2024-05-14
AAP-22640: original task names lost in multi-task plays in playbooks (fix)

mabulgu:
Jira Issue: <https://issues.redhat.com/browse/AAP-22640>

## Description
Please see the details here: https://issues.redhat.com/browse/AAP-22640

### Steps to test
1. Pull down the PR
2. Run the `ai.api.tests.test_formatter` tests to see if the formatter unit tests work.
3. On stage or your local: try to go thru the scenarios mentioned below both for playbooks and roles.

### Scenarios tested
Tests are seperated for roles and playbooks. 

1. Test for an ordinary role
2. Test for a role that has "loop" in it which contains "-name" under it. 
3. Test for a playbook 
4. Test for a playbook that has tasks contains "loop" in it which contains "-name" under it. 

Do all for single and multi task cases.

## Production deployment


mabulgu:
@goneri @robinbobbitt I am back on this one! Thanks for your comments. Let me re-check them one by one and try to adress/implement them. I will reach out to you in case of any questions.

ansible_wisdom/ai/api/formatter.py
```diff
@@ -353,15 +353,31 @@ def get_task_names_from_tasks(tasks):
...
 
-def restore_original_task_names(output_yaml, prompt):
+def restore_original_task_names(output_yaml, prompt, payload_context=''):
     if output_yaml and is_multi_task_prompt(prompt):
+        full_yaml = payload_context + output_yaml
+        try:
+            data = yaml.safe_load(full_yaml)
+        except Exception as exc:
+            logger.exception(f'Error while loading the result role/playbook YAML:{exc}')
+            return
```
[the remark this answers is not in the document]
mabulgu: @robinbobbitt what do you mean by returning `None`? for the `payload_context` or the `output_yaml`? By `the calling code` you mean the `data = yaml.safe_load(full_yaml)` line? 

This part of the code just tries to be safe with the YAML stuff. if there is a problem with the YAML content, then we return and the `restore_original_task_names` simply do nothing.

ansible_wisdom/ai/api/formatter.py
```diff
@@ -353,15 +353,31 @@ def get_task_names_from_tasks(tasks):
...
+                task_list = data
+
+        for i, task in enumerate(task_list):
             try:
-                task_line = match.group(0)
-                task = match.group(1)
-                restored_task_line = task_line.replace(task, prompt_tasks[i])
+                task_name = task.get("name", "")
+                if not task_name:
+                    continue
```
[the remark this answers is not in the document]
mabulgu: yeah right I am adding this into the tests. good catch. this is the scenario that we have a list, but we dont have a "name" key in that list. the list isnt always a task list. I will also add a comment explanation there with the specifics.

ansible_wisdom/ai/api/formatter.py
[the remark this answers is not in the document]
robinbobbitt: @mabulgu I'm talking about line 363 where we do a `return` with no value, which returns `None`. Re: calling code, I'm referring to the code that calls `restore_original_task_names`.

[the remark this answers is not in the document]
mabulgu: I think the current place of the `restore_original_task_names` is good because we pass the`anonymized_recommendation_yaml` and get the `recommendation_yaml` from the function. The yaml that I want to check in the `restore_original_task_names` is the full yaml so it is different in my case. I dont check the recommendation yaml only but the whole yaml with the potential payload context  if it is a playbook. 

Your idea of returning `output_yaml` makes alot sensein this case because if we have a problem with restoring the original task names because of a problem with the merged yaml (`payload_context + output_yaml`) then we should have the raw, unchanged one as the returned value to continue our validation like `yaml.safe_load(recommendation_yaml)` in the `post_process.py`. I've made the changes reg. this.

ansible_wisdom/ai/api/formatter.py
```diff
@@ -353,15 +353,31 @@ def get_task_names_from_tasks(tasks):
     return names
 
 
-def restore_original_task_names(output_yaml, prompt):
+def restore_original_task_names(output_yaml, prompt, payload_context=''):
     if output_yaml and is_multi_task_prompt(prompt):
+        full_yaml = payload_context + output_yaml
```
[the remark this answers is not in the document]
mabulgu: I've checked with your content by writing it by myself and faced no issue. probably because of the issue that Robin mentioned here: https://github.com/ansible/ansible-ai-connect-service/pull/1013#discussion_r1611491043

I fixed it by returning the same output yaml in case of any YAML load issue not to block the process.

ansible_wisdom/ai/api/formatter.py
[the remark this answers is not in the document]
mabulgu: Done

ansible_ai_connect/organizations/migrations/0003_remove_organization_telemetry_opt_out_and_more.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.11 on 2024-06-24 12:30
```
goneri, line 1: Thank you @mabulgu. Yes. This is indeed a left over change that was missing in the migration scripts. I agree we need to address this problem, but I would prefer if we handle it in a different PR.

mabulgu: oh I must missed this one. I would not include this in this PR already.I will remove it.

mabulgu: Done.

mabulgu:
@goneri @robinbobbitt would you mind checking this out again when you can?

goneri (dismissed):
Works fine. Thank you @mabulgu.

Note, this is a feature I would like to replace with a pre-filter at the extension level.

goneri:
@mabulgu could you Squash your branch before merging?

ansible_ai_connect/ai/api/formatter.py
```diff
@@ -363,15 +363,36 @@ def get_task_names_from_tasks(tasks):
...
+                f"Error while loading the result role/playbook YAML:{exc} "
+                f"for restoring the original task names"
+            )
+            return output_yaml
         prompt_tasks = get_task_names_from_prompt(prompt)
-        matches = re.finditer(r"^- name:\s+(.*)", output_yaml, re.M)
-        for i, match in enumerate(matches):
+        task_list = []
+        if isinstance(data, list) and len(data) > 0 and isinstance(data[0], dict):
+            # Tries to get the tasks from the dict if it's a playbook (payload_context!='')
```
robinbobbitt, line 380: We can't assume that empty payload_context implies it's a tasks file in a role and non-empty payload_context implies playbook. The multi-task prompt, in either type of file, may not be the first task in the file. And when it's not, the logic here currently fails because it assumes the first task in task_list is the first task of the multi-prompt. Please add tests (and fix the logic here) for playbooks and tasks in a role where the multi-task prompt is after an existing task.

ansible_ai_connect/ai/api/formatter.py
```diff
@@ -363,15 +363,36 @@ def get_task_names_from_tasks(tasks):
...
+                task_list = data
+
+        for i, task in enumerate(task_list):
             try:
-                task_line = match.group(0)
-                task = match.group(1)
-                restored_task_line = task_line.replace(task, prompt_tasks[i])
+                task_name = task.get("name", "")
+                # We have a list, but not all lists can be task lists
+                # This is for handling the non-task lists case
```
robinbobbitt, line 391: Can you explain when this would happen? Your comments above seem to imply that task_list is the tasks list. What else might it be?

robinbobbitt (changes requested):

manstis:
@robinbobbitt I assume @mabulgu is waiting for your approval/sign-off of your comments before merging?

@mabulgu if I am mistaken perhaps this can be merged soon?

(I'm just "House keeping", looking at old PRs)

mabulgu:
> @robinbobbitt I assume @mabulgu is waiting for your approval/sign-off of your comments before merging?
> 
> @mabulgu if I am mistaken perhaps this can be merged soon?
> 
> (I'm just "House keeping", looking at old PRs)

@manstis thanks man! @robinbobbitt had her comments lastly and I was expected to return this soon however I had another task and I could not get into it. Currently I am looking into it to see what's the buggy part about it as Robin stated. When I re-re-rework on this and solve the problems, then this can be merged. Thanks for the heads up!

jameswnl:
PR Deployment created in namespace wisdom-pr-1013

mabulgu:
@goneri can you pls take a look when available? Robin might not be available for this review anymore so I needed to ping you for this one again.

goneri (approved):
Thank you @mabulgu!

mabulgu:
@robinbobbitt could you give us a quick approval here, so that we can merge it? (I did not want to ping you first and I would dismiss the review (by applying all the requested changes) but thought its better to ping you for an approval)

ansible_ai_connect/ai/api/formatter.py
```diff
@@ -368,15 +368,33 @@ def get_task_names_from_tasks(tasks):
...
+        full_yaml = payload_context + output_yaml
+        try:
+            payload_context_data = yaml.safe_load(payload_context)
+            full_data = yaml.safe_load(full_yaml)
+        except Exception as exc:
+            logger.exception(
+                f"Error while loading the result role/playbook YAML:{exc} "
+                f"for restoring the original task names"
+            )
+            return output_yaml
```
robinbobbitt, line 382: Do we want a request to keep processing and potentially ultimately give the user a response when we weren't able to fix up the output? I'd think no. We know that currently it's going to fail in subsequent steps due to invalid yaml, but if the code around this function is restructured that wouldn't be the case. Should we raise this exception instead of just letting processing continue?

ansible_ai_connect/ai/api/formatter.py
```diff
@@ -368,15 +368,33 @@ def get_task_names_from_tasks(tasks):
...
+        prompt_task_names = get_task_names_from_prompt(prompt)
+
+        full_task_list = get_task_list_from_yaml_data_obj(full_data)
+        payload_context_task_list = get_task_list_from_yaml_data_obj(payload_context_data)
+        task_list = []
+
+        for task in full_task_list:
+            if task not in payload_context_task_list:
```
robinbobbitt, line 390: It's an edge case but what if there is a task name in the multitask prompt that ends up matching something in the context? Is there another way we could approach this, like maybe assuming the tasks we're updating will always be the last X in the list? Are there scenarios where that doesn't work?

ansible_ai_connect/ai/api/formatter.py
mabulgu: I think we should let the process continue but log it as an error to be traced just like we discussed before

ansible_ai_connect/ai/api/formatter.py
mabulgu: Instead of the comparison I can get the task list length from the context and start loping in the whole task list from that index.

The reason I implemented like this is I thought there can be a task added between two for example but yeah we are having the suggestions at the end of the file always, so the item count trick might work. I will update this and ping you.

mabulgu: @robinbobbitt applied the mentioned changes. Pls check when available. I want to merge this one in this sprint if no major issues exist.

goneri (approved):

PR #1098, opened 2024-06-05
TestCompletionWCAView: don't assume ANSIBLE_AI_MODEL_NAME is set

goneri:
Reduce the dependency between the configuration to run the tests.


manstis (approved):
LGTM 👍

PR #1099, opened 2024-06-06
Telemetry Schema 2 for Playbook Exp/Gen

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-21502>

## Description

Push events when:

- we preare an explanation for a playbook
- or we generate a new one

Also: 

- rename `PlaybookOutlineFeedback` as `PlaybookGenerationFeedback` and
  accept the `wizardId` parameter.
- rename `ContentMatches._write_to_segment()` to
  `ContentMatches.write_to_segment()` to match the other class of the module.

## Testing

- unit-tests

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/views.py
+++ b/ansible_wisdom/ai/api/views.py
@@ -617,7 +617,7 @@ class ContentMatches(GenericAPIView):
                 event['modelName'] = model_id
                 send_segment_event(event, event_name, user)
             else:
-                self._write_to_segment(
+                self.write_to_segment(
                     request_data,
                     duration,
                     exception,
@@ -664,7 +664,7 @@ class ContentMatches(GenericAPIView):
             )
         finally:
             duration = round((time.time() - start_time) * 1000, 2)
-            self._write_to_segment(
+            self.write_to_segment(
                 request_data,
                 duration,
                 exception,
@@ -677,7 +677,7 @@ class ContentMatches(GenericAPIView):
 
         return response_serializer
 
-    def _write_to_segment(
+    def write_to_segment(
         self,
         request_data,
         duration,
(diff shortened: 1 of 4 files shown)


ansible_wisdom/ai/api/views.py
```diff
@@ -753,11 +755,26 @@ def post(self, request) -> Response:
             "explanationId": explanation_id,
         }
 
+        self.write_to_segment(
```
TamiTakamiya, line 758: I think we can re-use the existing `_write_to_segment` method.  It contains the properties for capturing errors when API returns non-200 status codes and metadata for the version of metadata.

TamiTakamiya (commented):
I have left a one comment. Thanks.

ansible_wisdom/ai/api/views.py
goneri: The existing `_write_to_segment()` was associated with the `ContentMatches` class, the other classes of the module have the same method called `write_to_segment()`.

TamiTakamiya: I understand on the use of `wrote_to_segment`, but I still want to capture errors as we already see WCA API returns non-200 status codes like 500 or 400. For completion events, we have this Amplitude chart:

https://app.amplitude.com/analytics/redhat/chart/ejrbomr

I want to have similar analytics available for Explanation/Generation APIs.

goneri: :thinking:  I may be wrong, but I believe exceptions are already handled by Django here. https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_wisdom/main/exception_handler.py

goneri: @TamiTakamiya, oh now I understand, `SegmentMiddleware` only works for the completion requests.

TamiTakamiya: > SegmentMiddleware only works for the completion requests.

Unfortunetely, yes. That's the original code I wrote for telemetry support, but we added more later in the view classes. By implementing telemetry logic in a middleware, we can capture some errors like 401, which cannot be caputured in view classes, but the logic becomes more compicated (or messy...)

PR #1100, opened 2024-06-06
faux commercial users use faux commercial org

robinbobbitt:
Jira Issue: https://issues.redhat.com/browse/AAP-25100

## Description
This is a follow-up to the changes in https://github.com/ansible/ansible-ai-connect-service/commit/615dec7c38c5b2c4b3384ae0e27d0875b1c0cfe0
We have some test users that don't have an org. They are part of the "Commercial" django group and need to use this fake org to retrieve their WCA configuration. This PR leverages the org_id user property to account for faux commercial users.

### Steps to test
1. Pull down the PR
2. Run an API test with a faux commercial user
3. Inferences should work again as faux commercial user

### Scenarios tested

Verified with steps above plus unit test that fails without my code change.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/model_client/tests/test_wca_client.py
+++ b/ansible_wisdom/ai/api/model_client/tests/test_wca_client.py
@@ -21,6 +21,7 @@ from unittest.mock import ANY, Mock, patch
 
 import django.utils.timezone
 import requests
+from django.contrib.auth.models import Group
 from django.test import TestCase, override_settings
 from prometheus_client import Counter, Histogram
 from requests.auth import HTTPBasicAuth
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM :+1: Great debugging!

goneri (commented):
Should we consider removing the "commercial" group? It serves little purpose now, and it creates corner cases situation like this.

goneri (approved):

robinbobbitt:
> Should we consider removing the "commercial" group? It serves little purpose now, and it creates corner cases situation like this.

@goneri , yes, agree. We'd need to get any of our test users over to "real" users first. I opened https://issues.redhat.com/browse/AAP-25116. Thanks!

PR #1101, opened 2024-06-06
serializers: simplify the Pydantic configuration

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-21502>

## Description

- Avoid duplicated definition of the fields in both `Meta` and as a field.
- Inherite from the `Metadata` class to provide the `ansibleExtensionVersion` field.
 - Bonus point: the `ansibleExtensionVersion` fields are properly documented now.

## Testing

- unit-tests

## Production deployment


manstis (approved):
LGTM 👍

PR #1103, opened 2024-06-07
cleanup API type ... type

gebhardtr:


manstis (approved):
LGTM :+1: Good catch.

PR #1105, opened 2024-06-08
[REFACTORING DRAFT]Telemetry

goneri:
Push events when:

- we preare an explanation for a playbook
- or we generate a new one

Also:

- rename `PlaybookOutlineFeedback` as `PlaybookGenerationFeedback` and
  accept the `wizardId` parameter.
- rename `ContentMatches._write_to_segment()` to
  `ContentMatches.write_to_segment()` to match the other class of the module.


hasys:
Hi @goneri , could you please confirm if this PR is still active?

PR #1104, opened 2024-06-07
req: upgrade langchain

goneri:
Upgrade langchain to address GHSA-3hjh-jh2h-vrg6.

See: https://osv.dev/vulnerability/GHSA-3hjh-jh2h-vrg6

Also ignore SNYK-PYTHON-LANGCHAIN-7217837 since we don't use the module in question.
    
See:  https://security.snyk.io/vuln/SNYK-PYTHON-LANGCHAIN-7217837

PR #1107, opened 2024-06-10
update insecure dependencies

gebhardtr:


pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
   'PyYAML~=6.0',
   'requests~=2.32.0',
   'segment-analytics-python~=2.2.2',
-  'sentence-transformers==2.*',
+  'sentence-transformers~=2.6.1',
```
goneri, line 42: We don't really need to update the `pyproject.toml` as soon as one of our dep has a sec-vul. If we have too strict rules this can lead for dependency hell for the users, and it's the user responsibility to ensure the environment is up to date from a security perspective.

gebhardtr: thanks Gonéri; in this case I think it's what we want (newer than 2.6.1 but in the same family) and it's consistent with (most) of the other entries but your broader point about pyproject v requirements* is a good one.

Can we update this and the rest of our dependencies in a new PR when we've defined a standard for it?

gebhardtr:
> @gebhardtr I've actually a fix for that: #1104

Awesome, thank you. I think our PRs are complementary -- the vulnerabilities don't directly affect us (so we can merge this now) but it's still a good idea to update the libraries like you've done in your PR.

goneri:
@gebhardtr can you +1, the CI is :green_circle:  https://github.com/ansible/ansible-ai-connect-service/pull/1104

PR #1110, opened 2024-06-10
bump protobuf to 4.25.3

robinbobbitt:
Jira Issue: https://issues.redhat.com/browse/AAP-23758

## Description
This addresses CVE-2024-2410. 

### Scenarios tested
No testing has been completed on this change. We do not currently have an active use case for the gRPC client. No regressions are anticipated due to this minor version change.

## Production deployment


goneri (approved):
I assume protobuf is only used by the GRPC layer that we don't use anymore and is due for removal.

Issue #1113, opened 2024-06-11
unable to start ai connect service


gebhardtr:
Details:

| Item | Value |
|-------|------|
| Platform | Darwin  23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64


```
Traceback (most recent call last):                                                                                                                                                                                    
  File "/var/www/venv/bin/wisdom-manage", line 8, in <module>                                                                                                                                                         
    sys.exit(main())                                                                                                                                                                                                  
             ^^^^^^                                                                                                                                                                                                   
  File "/var/www/ansible-wisdom-service/ansible_wisdom/manage.py", line 34, in main                                                                                                                                   
    execute_from_command_line(sys.argv)                                                                                                                                                                               
  File "/var/www/venv/lib64/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line                                                                                      
    utility.execute()                                                                                                                                                                                                 
  File "/var/www/venv/lib64/python3.11/site-packages/django/core/management/__init__.py", line 416, in execute                                                                                                        
    django.setup()                                                                                                                                                                                                    
  File "/var/www/venv/lib64/python3.11/site-packages/django/__init__.py", line 24, in setup                                                                                                                           
    apps.populate(settings.INSTALLED_APPS)                                                                                                                                                                            
  File "/var/www/venv/lib64/python3.11/site-packages/django/apps/registry.py", line 124, in populate                                                                                                                  
    app_config.ready()                                                                                                                                                                                                
  File "/var/www/ansible-wisdom-service/ansible_wisdom/ai/apps.py", line 84, in ready                                                                                                                                 
    from .api.model_client.ollama_client import OllamaClient                                                                                                                                                          
  File "/var/www/ansible-wisdom-service/ansible_wisdom/ai/api/model_client/ollama_client.py", line 17, in <module>                                                                                                    
    from langchain.llms import Ollama                                                                                                                                                                                 
  File "/var/www/venv/lib64/python3.11/site-packages/langchain/llms/__init__.py", line 544, in __getattr__                                                                                                            
    from langchain_community import llms                                                                                                                                                                              
ModuleNotFoundError: No module named 'langchain_community'
```

goneri:
How did we end up with Ollama in prod?

manstis:
I guess because it's not a _lazy_ import?

[...]

PR #1114, opened 2024-06-11
make langchain_community a prod requirement

gebhardtr:
Fixes: https://github.com/ansible/ansible-ai-connect-service/issues/1113
Fixes: https://issues.redhat.com/browse/AAP-25218


goneri:
I believe we should make Langchain optional since we don't need it for the SAAS service https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

goneri (approved):

PR #1108, opened 2024-06-10
AAP-25087: Bump ARI to 0.2.6 to cut back on image size

romartin:
**WIP - Just testing how the generated images look....**

Jira Issue: <https://issues.redhat.com/browse/AAP-25087>

## Description
Bumping to ARI 0.2.6, in order to reduce the resulting service's image size.

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment

- [x] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1119, opened 2024-06-13
AAP-25320: Analytics Telemetry: Filter out playbook keywords from the prediction to guarantee parsing the right module/collection

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-25320>

## Description
Guarantee  we are filtering out Ansible `Playbook's Task` keywords from the module/collection parsing used for telemetry events payload.

## Considerations
- Tested this PR on top of https://github.com/ansible/ansible-ai-connect-service/pull/1108 , in order to ensure no dependency to `ansible` is declared (just the `ansible-core` one)

- Couldn't find a suitable (pre-computed) `List` object in `ansible-core` that provides all available `Task`'s class (and subclasses) fields available to use, so did a once (`global`) computation to get that list/set
-- Upsides: Dynamic Task fields available. No need for further maintaining once upgrading
-- Downsides: Memoized value (memory consumption). Computation costs once spinning up a pod, in order to memoize the value. Although honestly both values are not really considerable, I guess

- Another solution we can take is just to "hardcode" values somewhere in code, although will require for a few maintenance on doing `ansible-core` upgrades, but will not require for initial computation costs, just runtime predicate checks

- Another solution is do this same class introspection logic, but instead of at runtime, do it during some phase on the build time, so no initial computation costs, just runtime predicate checks again

### Scenarios tested
Tested by running actual unit tests, and increasing the coverage to support [those edge cases](https://github.com/ansible/ansible-ai-connect-service/pull/1119/files#diff-93b1364ecdbc7a29eb3a289b7f4d4c7673a3208c3bcbb401adae6ba583c01d07R655).

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/formatter.py
+++ b/ansible_wisdom/ai/api/formatter.py
@@ -17,6 +17,7 @@ import re
 from io import StringIO
 
 import yaml
+from ansible.playbook.task import Task
 from ruamel.yaml import YAML, scalarstring
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


TamiTakamiya (approved):
LGTM. Thank you!

romartin:
Rebased and signed commit

PR #1121, opened 2024-06-13
better logging for unrecognized 404 inference

robinbobbitt:
Jira Issue: https://issues.redhat.com/browse/AAP-25477

## Description
There are WCA requests failing in production with a 404 that don't match ResponseStatusCode404WCABadRequestModelId and fall into generic handling, which does not log enough detail to troubleshoot. This PR adds an additional 404 check that logs the response body. It also adds an extra check in ResponseStatusCode404WCABadRequestModelId to be sure it's json before calling json(), which raises an error if the response is not json.

### Steps to test
1. Pull down the PR
2. Adjust the prediction_url in BaseWCAClient.infer_from_parameters to something one character off
3. Make a completion request

Confirm the 404 response is nicely logged.

### Scenarios tested
Steps above plus unit tests.

## Production deployment

### DIFF
--- a/ansible_wisdom/ai/api/model_client/wca_utils.py
+++ b/ansible_wisdom/ai/api/model_client/wca_utils.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import logging
 from abc import abstractmethod
 from typing import Generic, TypeVar
 
@@ -19,12 +20,15 @@ from .exceptions import (
     WcaBadRequest,
     WcaCloudflareRejection,
     WcaEmptyResponse,
+    WcaInferenceFailure,
     WcaInvalidModelId,
     WcaTokenFailureApiKeyError,
     WcaUserTrialExpired,
 )
 
 T = TypeVar('T')
+
+logger = logging.getLogger(__name__)
 
 
 class Check(Generic[T]):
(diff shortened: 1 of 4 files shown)


ansible_wisdom/ai/api/model_client/wca_utils.py
```diff
@@ -161,6 +178,7 @@ def __init__(self):
                 ResponseStatusCode403UserTrialExpired(),
                 ResponseStatusCode403(),
                 ResponseStatusCode404WCABadRequestModelId(),
+                ResponseStatusCode404(),
```
manstis, line 181: Do we also need to check for the unknown 404 for `contentmatch`: https://github.com/ansible/ansible-ai-connect-service/blob/baf482d6490ddda9c09aa373dcf3d904d65544a1/ansible_wisdom/ai/api/model_client/wca_utils.py#L198

manstis (approved):
LGTM 👍 just one question, but approving as it's good for `completions`.

ansible_wisdom/ai/api/model_client/wca_utils.py
robinbobbitt: We should... I thought about it but didn't because it was bigger than just adding it to the contentmatch list since it raises a WcaInferenceFailure, which wouldn't be right. I was thinking perhaps we should have a generic check at the end that for any non-200 logs some details and raises a base WcaException. In a future iteration... :)

PR #1126, opened 2024-06-14
black: remove skip-string-normalization

goneri:
Remove the `skip-string-normalization = true` parameter that prevent Black
from fixing the quotes using in the code base.


manstis (approved):
"LGTM 👍 "

PR #1130, opened 2024-06-17
remove langchain from the base image

goneri:
Langchain is not an optional dependency that can be enabled with the `langchain` variant
of the package. e.g:`pip install --use-feature=langchain .`

As a result, the default image with the prod label won't embed langchain. However, the
default dev image will continue to include it.

See: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies
See: https://docs.podman.io/en/latest/markdown/podman-build.1.html#target-stagename

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ dependencies = [
   'ipython~=8.10.0',
   'jwcrypto~=1.5.6',
   'jinja2~=3.1.3',
-  'langchain~=0.2.3',
   'launchdarkly-server-sdk~=8.3.0',
   'opensearch-py~=2.1.1',
   'protobuf~=4.25.3',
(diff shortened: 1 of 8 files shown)


romartin:
Good to know, also good stuff, thanks @goneri

gebhardtr:
doesn't this increase the likelihood that there will be a failure in production because a code path is traversed that contains a reference to one of these dependencies (like Ollama and BAM clients)?

goneri:
> doesn't this increase the likelihood that there will be a failure in production because a code path is inadvertently traversed that contains a reference to one of these dependencies (like Ollama and BAM clients)?

We should catch those during the integration tests. We are also balancing this risk with the reduced security exposure due to the lowest number of dependencies.

PR #1136, opened 2024-06-18
settings: ensure we fallback on "saas" if DEPLOYMENT_MODE is unset

goneri:
## Description

As defined in the configuration, `podman-compose` will always define  `DEPLOYMENT_MODE`. However, the environment variable may be empty.

This change ensures we fallback on the value `saas` in such situation.

## Production deployment


manstis (approved):
LGTM :+1: 

I thought the old code would have used the default `cast(t_deployment_mode, "saas")` if the environment variable was undefined.. obviously(?) not.

goneri:
> LGTM 👍
> 
> I thought the old code would have used the default `cast(t_deployment_mode, "saas")` if the environment variable was undefined.. obviously(?) not.

Yes, the variable IS defined but an empty string.

PR #1142, opened 2024-06-18
continue the ansible_ai_connect transition

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-19331>

Continue the transition to `ansible_ai_connect`.

See: https://github.com/ansible/ansible-ai-connect-service/pull/990
See: https://github.com/ansible/ansible-ai-connect-service/pull/883

### DIFF
--- a/tools/configs/uwsgi.ini
+++ b/tools/configs/uwsgi.ini
@@ -14,7 +14,7 @@ lazy-apps = true                     ; if this is disabled, we must initialize t
                                      ; and feature_flags.py
 skip-atexit = true                   ; to avoid exception during the on_exit (e.g Pytorch)
 manage-script-name = true
-mount = /=ansible_wisdom.main.wsgi:application
+mount = /=ansible_ai_connect.main.wsgi:application
 
 need-app = true                      ; Exit if no python application loaded
 harakiri = 60                        ; Forcefully kill workers after 60 seconds
--- a/.flake8
+++ b/.flake8
@@ -58,8 +58,8 @@ per-file-ignores =
   # S101: Allow the use of assert within the tests directory, since tests require it.
   tests/**.py: S101
   # ignores for auto-generated files
-  ansible_wisdom/ai/api/model_client/grpc_pb/**.py: E402, E501, E712, F401, F821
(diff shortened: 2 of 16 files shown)


tools/configs/nginx-wisdom.conf
```diff
@@ -1,5 +1,5 @@
 upstream uwsgi {
-    server unix:///var/run/uwsgi/ansible_wisdom.sock;
+    server unix:///var/run/uwsgi/ansible_ai_connect.sock;
```
manstis, line 2: Does [this](https://github.com/ansible/ansible-ai-connect-service/blob/main/tools/configs/uwsgi.ini) need updating too? If so, there's an [equivalent](https://github.com/ansible/ansible-ai-connect-operator/blob/main/roles/model/templates/model.configmap.wisdom-service.yaml.j2) in the Operator code-base that would also need updating.

manstis (approved):
LGTM :+1: 

Just a small comment about another file not in this PR.

tools/configs/nginx-wisdom.conf
goneri: Yes you're right. I will remove this change from the PR.

manstis (dismissed):
LGTM :+1:

tools/configs/uwsgi.ini
```diff
@@ -14,7 +14,7 @@ lazy-apps = true                     ; if this is disabled, we must initialize t
                                      ; and feature_flags.py
 skip-atexit = true                   ; to avoid exception during the on_exit (e.g Pytorch)
 manage-script-name = true
-mount = /=ansible_wisdom.main.wsgi:application
+mount = /=ansible_ai_connect.main.wsgi:application
```
manstis, line 17: Can you please also submit a PR for the Operator.

https://github.com/ansible/ansible-ai-connect-operator/blob/main/roles/model/templates/model.configmap.wisdom-service.yaml.j2#L28

manstis (approved):
LGTM :+1:

tools/configs/uwsgi.ini
goneri: https://github.com/ansible/ansible-ai-connect-operator/pull/214

PR #1146, opened 2024-06-19
AAP-23235: Include variable values with enhanced context sent from VSCode

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-23235

## Description
Anonymize context variables before passing onto WCA.

## Testing
_Locally_ you need to put a break-point in `wca_client.py` on `BaseWCAClient.infer(..)`.

Submit a completion request from VSCode (compiled including the applicable enhancement, see below) with a playbook in VSCode that includes an external variables file (or other mechanism to provide _additional context_). Examine the value of `context` in `BaseWCAClient.infer(..)`. PII fields should be anonymized.

### Steps to test
1. Pull down the PR
2. `make start-backends`
3. `make create-application`
4. Set a break-point as above.
5. Run service in "debug mode".
6. Run VSCode and use the Ansible `.vsix` attached to JIRA.
7. Configure it to use your local `ansible-ai-connect-service` i.e. Lightspeed URL: http://localhost:8000
8. Create a playbook:
```
---
- name: Test the vscode extension
  hosts: all
  vars:
    favcolor: blue
  vars_files:
    - vars.yml

  tasks:
    - name: install apache
```
9. Create a variables file (called `vars.yml`):
```
---
external_var_1: value1
external_var_2: value2
password: magic
```
10. Submit a completion request (i.e. press <enter> after "install apache" in the playbook task).
11. Observe the `context` used to construct the `prompt` sent to WCA.

### Scenarios tested
As above.

## Production deployment

VSCode needs changing for the enhancement to be complete however it does not block this PR being merged.

See https://github.com/ansible/vscode-ansible/pull/1380

### DIFF
--- a/ansible_ai_connect/ai/api/tests/test_serializers.py
+++ b/ansible_ai_connect/ai/api/tests/test_serializers.py
@@ -133,6 +133,40 @@ class CompletionRequestSerializerTest(TestCase):
 
         self.assertTrue(serializer.is_valid())
 
+    def test_anonymized_additional_context(self):
+        user = Mock(rh_user_has_seat=True)
+        request = Mock(user=user)
+        data = {
+            "prompt": "---\n- hosts: all\n  become: yes\n\n  tasks:\n    - name: Install Apache\n",
+            "metadata": {
+                "additionalContext": {
+                    "playbookContext": {
+                        "varInfiles": {
+                            "vars.yml": "external_var_1: value1\n"
+                            "external_var_2: value2\n"
+                            "password: magic\n"
+                        },
+                        "roles": {},
+                        "includeVars": {},
+                    },
+                    "roleContext": {},
+                    "standaloneTaskContext": {},
+                },
+            },
+        }
+        serializer = CompletionRequestSerializer(context={"request": request}, data=data)
+        self.assertTrue(serializer.is_valid())
+        validated_data = serializer.validated_data["metadata"]
+        self.assertIsNotNone(validated_data)
+        self.assertTrue("additionalContext" in validated_data)
+        additional_context = validated_data["additionalContext"]
+        self.assertTrue("playbookContext" in additional_context)
+        playbook_context = additional_context["playbookContext"]
+        self.assertTrue("varInfiles" in playbook_context)
+        var_in_files = playbook_context["varInfiles"]
+        self.assertTrue("vars.yml" in var_in_files)
+        self.assertFalse("magic" in var_in_files["vars.yml"])
+
 
 class ContentMatchRequestSerializerTest(TestCase):
     def test_validate(self):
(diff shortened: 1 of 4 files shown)


ansible_wisdom/ai/api/tests/test_views.py
```diff
@@ -266,6 +266,31 @@ def test_wca_completion(self):
...
+        model_input["metadata"] = {
+            "additionalContext": {
+                "userid": "a-user-id",
+                "password": "a-password",
+            },
+        }
+        self.mock_model_client_with(model_client)
+        r = self.client.post(reverse("completions"), model_input, format="json")
+        self.assertEqual(r.status_code, HTTPStatus.OK)
+        self.assertEqual(
```
manstis, line 290: @robinbobbitt I need to assert that WCA is called with the additional context...

But this is where I don't understand _how_ WCA is meant to be passed the additional context.

I can trace passage of the `metadata` containing the additional context from the HTTP `request` into `model_mesh_client.infer(...)` but at this point `metadata` is dropped and only the `prompt` and `context` passed on (to WCA). I'm guessing something _extra_ needs adding.. but I'm not sure _how_ we should pass additional context to WCA. 

Can you advise?

manstis: I worked something out. See `test_views.py`.

manstis:
@robinbobbitt @goneri @TamiTakamiya the corresponding VSCode PR has been merged....

This means we risk sending *non-anonymized data* to WCA at the moment..

robinbobbitt:
> This means we risk sending _non-anonymized data_ to WCA at the moment..

Thanks for calling this out, Michael. Enhanced context isn't enabled yet in prod, so we are good on that front.

ansible_ai_connect/ai/api/tests/test_serializers.py
```diff
@@ -133,6 +133,40 @@ def test_validate_custom_model_no_seat_without_tech_preview(self):
...
+        user = Mock(rh_user_has_seat=True)
+        request = Mock(user=user)
+        data = {
+            "prompt": "---\n- hosts: all\n  become: yes\n\n  tasks:\n    - name: Install Apache\n",
+            "metadata": {
+                "additionalContext": {
+                    "playbookContext": {
+                        "varInfiles": {
+                            "vars.yml": "external_var_1: value1\n"
+                            "external_var_2: value2\n"
```
robinbobbitt, line 146: @manstis Are you sure about this example? Would we see individual key/value secrets in varInfiles, or will the field names always be file names?

manstis: @robinbobbitt yes, I captured it by running VSCode which expands variable files into the `additionalContext` provided.

Ideally I'd have liked to used other examples of `additionalContext` too but I don't know how to _create_ them. That said the anonymization [code](https://github.com/ansible/ansible-ai-connect-service/blob/19a2e9792a28a46a364c293632a98352f2c1f2cc/ansible_ai_connect/ai/api/fields.py#L36) works on the whole JSON dictionary and therefore, whatever is provided is anonymized.

robinbobbitt: Thanks for confirming, @manstis . When I work on my enhanced context issue, I'll try to test some additional scenarios and share any details.

robinbobbitt (approved):

PR #1147, opened 2024-06-19
api/view: improve the collect of the Schema1 events

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-25231>

## Description

Harmonize how we collect we prepare the Schema1 events and how we
do the payload validation and the exception handling.

Changes:
- Reduce the complexity of the middleware
- Revert https://github.com/ansible/ansible-ai-connect-service/pull/692, which will be replace by a better solution
- No more event when a payload Validation fails. The current design (one single end-point can have several different structures per type of payload) is too complicated and we can replace this we an generic `validationFailure` event.

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment


ansible_ai_connect/ai/api/views.py
```diff
@@ -162,11 +208,14 @@ def post(self, request) -> Response:
         return pipeline.execute()
 
 
-class Feedback(APIView):
+class Feedback(OurAPIView):
     """
     Feedback API for the AI service
     """
 
+    serializer_class = FeedbackRequestSerializer
```
manstis, line 216: This view is still handling de-serialization and validation [itself](https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9R242-R247) (in addition to that in `OurAPIView`).

ansible_ai_connect/ai/api/views.py
```diff
@@ -134,6 +131,55 @@
 }
 
 
+class OurAPIView(APIView):
```
manstis, line 134: Is it better to extend `GenericAPIView`?

`GenericAPIView` extends `APIView` and already has a `serializer_class` attribute

ansible_ai_connect/ai/api/views.py
```diff
@@ -337,40 +353,14 @@ def write_to_segment(
...
-                event_type = "unknown"
-
-            event = {
-                "data": ano_request_data,
-                "exception": str(exception),
-            }
-            send_segment_event(event, event_type, user)
-
 
 class ContentMatches(GenericAPIView):
```
manstis, line 357: Should this extend `OurAPIView` too?

You've [removed](https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9L550-L573) the code sending segment events; so presumably expect `OurAPIView` to be handling it?

You could also then be able to remove [the](https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9R396-R399) de-serialization?

manstis (commented):
A couple of comments...

It'll be an interesting exercise to update the `Completions` [view](https://github.com/ansible/ansible-ai-connect-service/blob/goneri/api-view-improve-the-collect-of-the-Schema1-events_2563/ansible_ai_connect/ai/api/views.py#L206-L208) too which [buries](https://github.com/ansible/ansible-ai-connect-service/blob/goneri/api-view-improve-the-collect-of-the-Schema1-events_2563/ansible_ai_connect/ai/api/views.py#L206-L208) the segment code in a class away from the `View`.

ansible_ai_connect/ai/api/views.py
goneri: Done, thanks.

ansible_ai_connect/ai/api/views.py
goneri: Yes, maybe. I will take a look once this PR is merged.

ansible_ai_connect/ai/api/views.py
goneri: Yes, indeed.

ansible_ai_connect/ai/api/views.py
goneri: So, answering a bit late on this. I'm not so sure `GenericAPIView` comes with various new attributes we don't need and if we want to use the `serializer_class` attribute, we need to also overload some existing classes.
My conclusion is both work and that we're better off with just `APIView` since it's the most basic one.

goneri:
unit-tests -> `FAILED (failures=54, errors=80, skipped=1)`

jameswnl:
PR Deployment created in namespace wisdom-pr-1147

ansible_ai_connect/ai/api/telemetry/schema1.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+@define
+class CompletionRequestPayload:
+    context: str = field(validator=validators.instance_of(str), converter=str, default="")
+    prompt: str = field(validator=validators.instance_of(str), converter=str, default="")
+
+
+@define
+class CompletionEvent(Schema1Event):
```
TamiTakamiya, line 152: Where is this event used?

goneri: It's associated with the Completion view here: https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9R221

ansible_ai_connect/ai/api/pipelines/completion_stages/post_process.py
```diff
@@ -78,32 +79,29 @@ def write_to_segment(
...
-        }
+        schema1_event = schema1.Postprocess()
+        schema1_event.details = postprocess_detail
+        schema1_event.truncated = truncated_yaml
+
+    elif event_type == "ansible-lint":
+
+        schema1_event = schema1.PostprocessLint()
+        schema1_event.postprocessed = postprocessed_yaml
+        schema1_event.problem = problem
```
manstis, line 92: Is this line needed? You do the same on L#98

ansible_ai_connect/ai/api/pipelines/completion_stages/post_process.py
```diff
@@ -78,32 +79,29 @@ def write_to_segment(
...
-            "suggestionId": str(suggestion_id) if suggestion_id else None,
-        }
+        schema1_event = schema1.Postprocess()
+        schema1_event.details = postprocess_detail
+        schema1_event.truncated = truncated_yaml
+
+    elif event_type == "ansible-lint":
+
+        schema1_event = schema1.PostprocessLint()
+        schema1_event.postprocessed = postprocessed_yaml
```
manstis, line 91: Is this line needed? You do the same on L#97

ansible_ai_connect/ai/api/pipelines/completion_stages/response.py
```diff
@@ -63,6 +63,7 @@ def process(self, context: CompletionContext) -> None:
         # Note: Currently we return an array of predictions, but there's only ever one.
         # The tasks array added to the completion event is representative of the first (only)
         # entry in the predictions array
+        # https://github.com/ansible/ansible-ai-connect-service/blob/0e083a83fab57e6567197697bad60d306c6e06eb/ansible_ai_connect/ai/api/pipelines/completion_stages/response.py#L64
```
manstis, line 66: Is this comment meant to be here?

ansible_ai_connect/ai/api/telemetry/schema1.py
```diff
@@ -0,0 +1,326 @@
...
+            "playbookGenerationAction": schema1.PlaybookGenerationActionEvent,
+        }
+        # TODO: handles the key that are at the root level of the structure
+        for key_name, schema1_class in mapping.items():
+            if key_name in validated_data:
+                schema1_event = schema1_class()
+                schema1_event.set_user(user)
+                schema1_event.set_validated_data(validated_data)
+                return schema1_event
+        print("Failed to init BaseFeedbackEvent")
```
manstis, line 267: Should we not at least use a logger?

I would have thought we should probably raise an exception too (although we hope this line is never reached)?

ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -41,9 +41,9 @@
 from rest_framework.test import APITransactionTestCase
 from segment import analytics
 
+import ansible_ai_connect.ai.api.utils.segment
 from ansible_ai_connect.ai.api.data.data_model import APIPayload
-from ansible_ai_connect.ai.api.exceptions import (
-    FeedbackValidationException,
+from ansible_ai_connect.ai.api.exceptions import (  # FeedbackValidationException,
```
manstis, line 46: I guess we don't need this comment.

ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -967,6 +965,8 @@ def test_completions_preprocessing_error(self):
                 )
                 self.assertSegmentTimestamp(log)
 
+    @skip("Collect invalid payload error")
```
manstis, line 968: This is a "TODO"?

ansible_ai_connect/ai/api/utils/segment.py
```diff
@@ -118,11 +120,32 @@ def base_send_segment_event(
...
+        if not analytics.write_key:
+            analytics.write_key = settings.SEGMENT_WRITE_KEY
+            analytics.debug = settings.DEBUG
+            analytics.gzip = True  # Enable gzip compression
+            # analytics.send = False # for code development only
+            analytics.on_error = on_segment_error
+
+
+def send_schema1_event(event_obj) -> None:
+    print(f"SENDING SCHEMA1 EVENT (name={event_obj.event_name})\n{event_obj.as_dict()} ({type(event_obj)})")
```
manstis, line 142: Use a logger vs `print`?

ansible_ai_connect/ai/api/views.py
```diff
@@ -134,11 +127,99 @@
...
+        return response
+
+    def get_ids(self):
+        allowed = ["explanationId", "generationId", "suggestionId"]
+        # Return the ids we want to include in the exception messages
+        ret = {}
+        for k, v in self.validated_data.items():
+            if k in allowed and v:
+                ret[k] = v
+            elif isinstance(v, dict):
```
manstis, line 181: Do we need to make handling of `dict` recursive?

ansible_ai_connect/ai/api/views.py
```diff
@@ -134,11 +127,99 @@
...
+            # during the processing of the request.
+            self.schema1_event.set_request(request)  # Read the note above
+            # before moving the line ^
+
+            # NOTE: We also want to include the final response in the event, we do that
+            # we need to jump back and do it from within a final middleware that wrap
+            # everything.
+            import ansible_ai_connect.main.middleware
+
+            ansible_ai_connect.main.middleware.global_schema1_event = self.schema1_event
```
manstis, line 210: Oeeeee... this smells bad, does it not!?!? Is this thread safe?

ansible_ai_connect/ai/api/views.py
```diff
@@ -469,142 +457,91 @@ def perform_content_matching(
                 logger.exception(f"error serializing final response for suggestion {suggestion_id}")
                 raise InternalServerError
 
-        except ModelTimeoutError as e:
-            exception = e
+        except ModelTimeoutError as exc:
```
manstis, line 460: Aren't all of these `except`'s handled by `OurAPIView` (or at least is that not the intention)?

ansible_ai_connect/main/middleware.py
```diff
@@ -31,129 +24,48 @@
...
-                        )
-                        request_data = anonymize_request_data(request_data)
-                    except Exception:  # when an invalid json or an invalid encoding is detected
-                        request_data = {}
-                else:
-                    request_data = anonymize_request_data(request.POST)
+                segment_analytics_telemetry.on_error = on_segment_schema2_error
 
         response = self.get_response(request)
+        if global_schema1_event:
```
manstis, line 59: See my comment earlier about use of this and thread-safety.

manstis (commented):
A few comments for consideration. There's only one _strong_ comment about passing the Schema1 event to `middleware` as a global.

ansible_ai_connect/ai/api/views.py
goneri: Yes, this is not thread-safe, but use one process per request.

ansible_ai_connect/ai/api/utils/segment.py
goneri: This is a debug trace will be removed later.

ansible_ai_connect/ai/api/views.py
goneri: Yes, we can also remove some of them in https://github.com/ansible/ansible-ai-connect-service/pull/1168, but I was just waiting for the unit-tests to pass before starting more changes.

ansible_ai_connect/ai/api/pipelines/completion_stages/post_process.py
goneri: indeed, thanks!

ansible_ai_connect/ai/api/telemetry/schema1.py
goneri: Totally agree, I converted that as a logger.error().

ansible_ai_connect/ai/api/views.py
goneri: I now use a Semaphore to protect the global variable.

ansible_ai_connect/ai/api/telemetry/schema1.py
```diff
@@ -0,0 +1,340 @@
...
+from ansible_ai_connect.healthcheck.version_info import VersionInfo
+from ansible_ai_connect.users.models import User
+
+logger = logging.getLogger(__name__)
+version_info = VersionInfo()
+
+
+@define
+class ResponsePayload:
+    exception: str = field(validator=validators.instance_of(str), converter=str, default="")
```
romartin, line 46: another comment that applies to all fields that have a `validator`, not just this one. It could be the case once the object instance  is created that some validator fails? If so, it will throw an exception, and I don't see any try/catch blocks around those instance creations around the code. 

For schema2 I just create every instance by using a lambda function, so the instance creation method does internally the try/catch, and in case any validation error in some segment event, it sends another segment event (schema1) that describes the error

romartin (commented):
Hey @goneri , 

Looks a great code improvement, thanks!! 

I just did a quick code review, added a few comments only. When this is ready for testing , please let me konw and will give it a try!!

ansible_ai_connect/ai/api/views.py
```diff
@@ -134,11 +128,100 @@
...
+            )
+            return model_name
+        except (WcaNoDefaultModelId, WcaModelIdNotFound, WcaSecretManagerError):
+            return ""
+
+    def handle_exception(self, exc):
+        self.exception = exc
+
+        # Mapping between the internal exceptions and the API exceptions (with a message and a code)
+        mapping = [
```
manstis, line 157: I think you need to also [map](https://github.com/ansible/ansible-ai-connect-service/blob/0f793e1e8bbec92177ff462ab0c7eb8f669e694a/ansible_ai_connect/ai/api/pipelines/completion_stages/inference.py#L141-L147) `WcaSuggestionIdCorrelationFailure`/`WcaSuggestionIdCorrelationFailureException` mapping? These exception names are changed in https://github.com/ansible/ansible-ai-connect-service/pull/1177

What about `ModelTimeoutError`/`ModelTimeoutException` [too](https://github.com/ansible/ansible-ai-connect-service/blob/0f793e1e8bbec92177ff462ab0c7eb8f669e694a/ansible_ai_connect/ai/api/pipelines/completion_stages/inference.py#L105-L111)? You don't have a re-try/timeout on Playbook gen/exp (yet) but I think it has been requested... so you'll need these exceptions in your refactoring too.

ansible_ai_connect/ai/api/views.py
```diff
@@ -134,11 +128,100 @@
...
+                        ret[subk] = subv
+        return ret
+
+    def dispatch(self, request, *args, **kwargs):
+        start_time = time.time()
+        self.exception = False
+        self.schema1_event = None
+        response = super().dispatch(request, *args, **kwargs)
+
+        if self.schema1_event:
```
robinbobbitt, line 195: If the request fails before `initial` is called, this won't be set. One example where this occurs is if authentication fails.

ansible_ai_connect/main/middleware.py
```diff
@@ -165,8 +81,8 @@ def __call__(self, request):
                                 for task in tasks
                             ],
                             rh_user_org_id=getattr(request.user, "org_id", None),
-                            suggestion_id=request_suggestion_id,
-                            model_name=model_name,
+                            suggestion_id=getattr(response, "suggestionId", ""),
```
robinbobbitt, line 84: There are scenarios where the suggestionId is in the request but doesn't make it into the response, such as if the request fails due to an authentication failure. This change would break that.

ansible_ai_connect/ai/api/views.py
robinbobbitt: Could/should we just stuff the schema1_event in the request object instead?

ansible_ai_connect/ai/api/views.py
goneri: Yes, this is an assumed decision. Those authentication errors should be handled differently IMO.

ansible_ai_connect/ai/api/views.py
goneri: I think both strategy are ugly and can be avoided. At this stage we use the middleware only to get the final HTTP code/message, and those can easily be retrieved from the final exception that we return https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9R169

Django will return a 500 unless the exception has a specific code. There is still potential 500 if the exception happens outside of the view, and we can use the `exception_handler()` to catch them.

PR #1138, opened 2024-06-18
finish the ansible_wisdom ansible_ai_connect transition

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-19331>

Rename the Python package from `ansible_wisdom` to `ansible_ai_connect`.

See: https://github.com/ansible/ansible-ai-connect-service/pull/990
See: https://github.com/ansible/ansible-ai-connect-service/pull/883

- Deploy the branch
- Validate the service runs like before

- [x] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1151, opened 2024-06-21
docker-compose: adjust the path of the Python directory

goneri:


manstis (approved):

goneri:
It will actually be easier to just land https://github.com/ansible/ansible-ai-connect-service/pull/1138 which set the final location.

PR #1165, opened 2024-07-01
Sync grpcio and grpcio-tools dependencies

hasys:

## Description

During fresh installation there was an error `ERROR: Could not build wheels for grpcio-tools, which is required to install pyproject.toml-based projects`

Make `grpcio` and `grpcio-tools` in sync with the same version fixed an issue.

### Steps to test
1. Pull down the PR
2. Run all unit-tests

### Scenarios tested

Described above and fresh local build.

## Production deployment


pb82 (approved):
I can confirm that the service starts successfully with this version of `grpcio` and Python 3.12.

PR #1167, opened 2024-07-01
tests: properly mock the model in AcceptedTermsPermissionTest

goneri:
Ensure we actually get a `WcaDummy` client as expected by mocking the
model with `WisdomAppsBackendMocking`.


goneri:
Without this change, the CI actually use the `HttpClient` model.

TamiTakamiya (approved):

PR #1169, opened 2024-07-01
completion_stages/response: only set model when True

goneri:
Only define the model key when the model is Truthy and avoid the
situation where the model key is set to `None` or `''`, which will
raise the following validation error:

```
'{"model":["This field may not be blank."]}'
```


manstis (approved):
LGTM 👍

PR #1170, opened 2024-07-01
settings: fix how default value are set

goneri:
`{docker,podman}-compose` set environment variables defined in the configuration.
Most of them will be empty. Because of the way `getenv()` works, we never actually
get the expected default values.

```
$ MY_VAR="" python3
>>> import os
>>> os.getenv("MY_VAR", "never-used")
''
```

This patch ensures we actually fail back on what was the intended defaults.


manstis (approved):
LGTM :+1:

PR #1172, opened 2024-07-02
telemetry_opt_out: use a different name for the property

goneri:
The `makemigrations` command was confused by the current situation were the column, the Django field and the virtual property were colliding.

As a result it was trying to:
- drop the column
- and recreate it.

This change clarify the model and uses a designed property called `has_telemetry_opt_out` to ready the value.


ansible_ai_connect/organizations/models.py
```diff
@@ -23,20 +23,19 @@
 
 class Organization(models.Model):
     id = models.IntegerField(primary_key=True)
-    _telemetry_opt_out = models.BooleanField(default=False, db_column="telemetry_opt_out")
+    telemetry_opt_out = models.BooleanField(default=False, db_column="telemetry_opt_out")
+    extra_kwargs = {
+        "telemetry_opt_out": {"write_only": True},
```
manstis, line 28: I don't understand why you're adding the `extra_kwargs` for `write_only: True`. Can you explain?

manstis (commented):
LGTM 👍 

Just a question before approving.

ansible_ai_connect/organizations/models.py
goneri: It was to turn the flag read-only as a safeguard and be sure we only use the has_ property instead.

goneri: https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

goneri: Actually, ignore what I just said. I was missing the serializer and the model...

ansible_ai_connect/organizations/migrations/0003_alter_organization_telemetry_opt_out.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.2.11 on 2024-07-02 17:23
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
manstis, line 6: Do we need to worry about data migration?

Will this rename a database column or create a new one and delete the old?

manstis (commented):
Sorry, your mention of the model and serialiser made me think.

ansible_ai_connect/organizations/migrations/0003_alter_organization_telemetry_opt_out.py
goneri: The migration script doesn't change anything. This is already the name in the DB. But the problem was that `makemigrations` has never been called after the previous model changes.

manstis (approved):
Thanks @goneri LGTM 👍

PR #1173, opened 2024-07-03
cleanup: remove seat management

pb82:
Jira Issue: https://issues.redhat.com/browse/AAP-25245

## Description

Removes the seat check logic and relevant unit tests.

## Testing

1. Run the wisdom service locally and try to log in with an AAP licensed user
2. It should succeed

### Scenarios tested

as described above

## Production deployment


ansible_ai_connect/users/authz_checker.py
```diff
@@ -1,459 +0,0 @@
...
-        try:
-            for item in result["items"]:
-                if item["role"]["id"] == "OrganizationAdmin":
-                    return True
-        except (KeyError, ValueError):
-            return False
-
-        return False
-
-    def rh_org_has_subscription(self, organization_id: int) -> bool:
```
goneri, line 366: We still need this method to know if the Org has an AAP subscription.

goneri (changes requested):

pb82:
@goneri thanks for the review. In that case, I guess we should remove the `check` method of the AMSCheck and CIAMCheck since they are no longer used?

goneri:
That sounds reasonable to me, even if we still use a tiny part the module.

pb82:
Closing this to split it up into multiple PRs

PR #1176, opened 2024-07-03
convert the T&C flow as a regular post-login page

goneri:
Move the existing T&C out of the login flow and prepare an associate page where
the user can accept them after the login.

See: https://issues.redhat.com/browse/AAP-26573


ansible_ai_connect/users/pipeline.py
```diff
@@ -124,69 +120,3 @@ def redhat_organization(backend, user, response, *args, **kwargs):
...
-
-def load_extra_data(backend, details, response, uid, user, *args, **kwargs):
-    social = kwargs.get("social") or backend.strategy.storage.user.get_social_auth(
-        backend.name, uid
-    )
-    if social:
-        extra_data = backend.extra_data(user, uid, response, details, *args, **kwargs)
-        user.external_username = extra_data.get("login")
-        user.save()
-        social.extra_data["aap_licensed"] = (
```
manstis, line 183: We need these properties for AAP user mapping. I don't see them elsewhere in this PR. 

Without them we loose certain [functionality](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/users/models.py#L132-L144) for "on prem" including protection of the `/metrics` [endpoint](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/main/views.py#L127-L145).

manstis (changes requested):
This change removes population of various `user` properties used for AAP/on-prem.

manstis (approved):
LGTM :+1:

PR #1177, opened 2024-07-04
AAP-25893: Set x-request-id header on playbook explanation/generation endpoints

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-25893

## Description
Include `X-Request-ID` header in the `POST`'s for Playbook generation/explanation.

Also include a check that the response contains the same identifier and raise an exception if not accordingly.
 
## Testing
It is not possible to test for real since it requires WCA to fail.

Unit tests emulate such failures.

### Scenarios tested
Unit tests.

## Production deployment

VSCode change to support new error code: https://github.com/ansible/vscode-ansible/pull/1414

### DIFF
--- a/ansible_ai_connect/ai/api/exceptions.py
+++ b/ansible_ai_connect/ai/api/exceptions.py
@@ -116,10 +116,13 @@ class WcaNoDefaultModelIdException(WisdomAccessDenied):
     default_detail = "No default WCA Model ID was found."
 
 
-class WcaSuggestionIdCorrelationFailureException(BaseWisdomAPIException):
+class WcaRequestIdCorrelationFailureException(BaseWisdomAPIException):
     status_code = 500
-    default_code = "error__wca_suggestion_correlation_failed"
-    default_detail = "WCA Request/Response Suggestion ID correlation failed."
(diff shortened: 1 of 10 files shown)


goneri (approved):
Side note for the future: We need to stop using different names for the request ID for each of our end-point.

manstis:
> Side note for the future: We need to stop using different names for the request ID for each of our end-point.

I did refactor some bits to make them generic for completions/gen/exp.

https://github.com/ansible/ansible-ai-connect-service/pull/1177/files#diff-052b0c3ff4c4ad8cbf30bbdaa0f88e9157b5e33a94c6ea85d38257f7010b576eR304

https://github.com/ansible/ansible-ai-connect-service/pull/1177/files#diff-f990bd5d355030dc1495e2872de831f3dc59732bf6d9c9a6528ce4064c97de8dR101

goneri:
> > Side note for the future: We need to stop using different names for the request ID for each of our end-point.
> 
> I did refactor some bits to make them generic for completions/gen/exp.

Yes, I saw that. I really like it :-).

robinbobbitt (approved):
Looks great! Thanks @manstis

PR #1179, opened 2024-07-08
increase AMS timeout to 2 seconds

pb82:
Increase the timeout for all AMS requests from 0.8 to 2 seconds. This hopefully resolves the recent AMS related test failures.

### DIFF
--- a/ansible_ai_connect/users/authz_checker.py
+++ b/ansible_ai_connect/users/authz_checker.py
@@ -195,6 +195,7 @@ class AMSCheck(BaseCheck):
         self._api_server = api_server
         self._ams_org_cache = {}
         self.retries = settings.AUTHZ_AMS_SERVICE_RETRY_COUNT
+        self.timeout = settings.AUTHZ_AMS_SERVICE_TIMEOUT
 
         if self._api_server.startswith("https://api.stage.openshift.com"):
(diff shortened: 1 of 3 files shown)


ansible_ai_connect/users/authz_checker.py
```diff
@@ -290,7 +290,7 @@ def check(self, user_id: str, username: str, organization_id: int) -> bool:
             # Organization has not yet been created in AMS, either the organization
             # is too recent (sync is done every 1h), or the organization has no AMS related
             # services (e.g Ansible, OpenShift, cloud.r.c) and so is not synchronized.
-            logger.warning(f"Organization not found in AMS, organization_id={organization_id}")
+            logger.warning(f"Organization can't be found in AMS, organization_id={organization_id}")
```
pb82, line 293: flake8 complains about the `not .. in` otherwise.

manstis (dismissed):
If the `2.0s` is _experimental_ it could be moved to a setting and looked up from the environment variables. We could then set this in the deployed environment secrets.

pb82:
@manstis it is experimental in the sense that I don't know if this will fix the AMS issues we're having. I can move this to an env var with a default value too :+1:

manstis (approved):
Super, smashing, great :+1:

goneri (approved):

PR #1185, opened 2024-07-09
drop the AcceptedTermsPermission permission

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-26773>

## Description

We don't use the permission anymore since the end of the TechPreview.

## Testing

SAAS already ignores this permission.

## Production deployment


manstis (approved):
LGTM 👍

PR #1188, opened 2024-07-10
home view: Don't use DOCUMENTATION_URL anymore

goneri:
DOCUMENTATION_URL was associated with the Tech Preview which is now done.


manstis (approved):
LGTM :+1:

PR #1190, opened 2024-07-10
pip-compile: set the SELinux context

goneri:
Ensure we set the SELinux label `container_file_t` on the shared volume.

See: https://docs.podman.io/en/latest/markdown/podman-run.1.html
Similar to what was done with 35e4fbc749a20c330d370239fce0927845c8499e.


manstis (approved):
LGTM :+1:

PR #1192, opened 2024-07-10
do not default ANSIBLE_AI_MODEL_MESH_MODEL_ID

robinbobbitt:
Jira Issue: https://issues.redhat.com/browse/AAP-24704

## Description
Reverts default value for ANSIBLE_AI_MODEL_MESH_MODEL_ID. When this env var is set, we no longer retrieve the model id from secrets manager, which breaks us when deploying in SaaS mode. This PR also removes model_name from the status endpoint since it is not useful in its current form.

### Steps to test
1. Pull down the PR
2. Run the wisdom service in SaaS mode
3. Run a polaybook generation
4. Confirm the model id is pulled from secrets manager, not hardcoded to "wisdom"

### Scenarios tested
above

## Production deployment

- [x] This code change requires the following considerations before going to production: Changes will be needed in ansible-wisdom-testing since model_name is being removed from the status endpoint.

### DIFF
--- a/ansible_ai_connect/ai/api/tests/test_views.py
+++ b/ansible_ai_connect/ai/api/tests/test_views.py
@@ -1364,7 +1364,7 @@ class TestCompletionView(WisdomServiceAPITestCaseBase):
             "suggestionId": str(uuid.uuid4()),
         }
         response_data = {
-            "model_id": settings.ANSIBLE_AI_MODEL_MESH_MODEL_ID,
+            "model_id": "wisdom",
             "predictions": ["      ansible.builtin.apt:\n        name: apache2"],
         }
         self.client.force_authenticate(user=self.user)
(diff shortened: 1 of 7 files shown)


manstis (approved):
LGTM :+1: Thank-you and sorry..

If I was being pedantic I'd suggest a test to avoid the scenario being repeated in the future.

goneri:
> LGTM 👍 Thank-you and sorry..
> 
> If I was being pedantic I'd suggest a test to avoid the scenario being repeated in the future.

well, the problem was caught by a test :-).

manstis:
> well, the problem was caught by a test :-).

Interesting. My PR was green before merge, so presumably an integration test somewhere.

robinbobbitt:
> Interesting. My PR was green before merge, so presumably an integration test somewhere.
Yep! And no worries of course, I didn't catch in the review either the side effect of reducing from two env vars to one. Fixing the unit tests now.

manstis (approved):
LGTM :+1: Thank-you.

manstis (approved):
LGTM :+1:

PR #1193, opened 2024-07-10
Add histograms and more logs to Authz APIs

TamiTakamiya:
Jira Issue: [AAP-26593](https://issues.redhat.com/browse/AAP-26593)

## Description

Add Prometheus histograms and more logs to Authz APIs for better traceability on those APIs.

## Testing

Unit tests are included.
### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

See above.

## Production deployment

### DIFF
--- a/ansible_ai_connect/users/authz_checker.py
+++ b/ansible_ai_connect/users/authz_checker.py
@@ -15,6 +15,7 @@
 #  limitations under the License.
 
 import logging
+import sys
 from abc import abstractmethod
 from datetime import datetime, timedelta
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM 👍

ansible_ai_connect/users/authz_checker.py
```diff
@@ -77,6 +120,15 @@ def fatal_exception(exc) -> bool:
         return False
 
 
+def log_backoff_exception(details):
+    _, exc, _ = sys.exc_info()
+    logger.info(str(exc))
+    logger.info(
```
jameswnl, line 126: should this be using `.error()`?

TamiTakamiya: @jameswnl  thanks for your comment.  Since these are for the exceptions that trigger retries, I think we can use `.info()` here.

However, it is possible that we might be retrying after receiving more critical exceptions. Until now we did not log them and were not sure what kind exceptions were thrown... If those un-retryable exceptions are identified, we should not retry API call and handle those exceptions separately. What do you think?

jameswnl (approved):

PR #1195, opened 2024-07-10
test_middleware: remove ANSIBLE_AI_ENABLE_TECH_PREVIEW=True

goneri:
The change breaks the `ANSIBLE_AI_ENABLE_TECH_PREVIEW=True` dependency in `test_middleware.py` and also introduces `WisdomServiceAPITestCaseBaseOIDC` which is `WisdomServiceAPITestCaseBase` with an OIDC user. This is a more realistic scenario and the new class should supersede the old one.

### DIFF
--- a/ansible_ai_connect/ai/api/tests/test_views.py
+++ b/ansible_ai_connect/ai/api/tests/test_views.py
@@ -94,7 +94,10 @@ from ansible_ai_connect.test_utils import (
     WisdomLogAwareMixin,
     WisdomServiceLogAwareTestCase,
 )
-from ansible_ai_connect.users.constants import USER_SOCIAL_AUTH_PROVIDER_AAP
+from ansible_ai_connect.users.constants import (
+    USER_SOCIAL_AUTH_PROVIDER_AAP,
+    USER_SOCIAL_AUTH_PROVIDER_OIDC,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -191,16 +194,22 @@ class WisdomServiceAPITestCaseBase(APITransactionTestCase, WisdomServiceLogAware
         analytics.send = False  # do not send data to segment from unit tests
         segment_analytics_telemetry.send = False  # do not send data to segment from unit tests
 
+    def create_user(self):
+        logger.warn("Please move this test to WisdomServiceAPITestCaseBaseOIDC")
+
+        self.user = get_user_model().objects.create_user(
+            username=self.username,
+            email=self.email,
+            password=self.password,
+        )
+
     def setUp(self):
         super().setUp()
         self.username = "u" + "".join(random.choices(string.digits, k=5))
         self.password = "secret"
-        email = "user@example.com"
-        self.user = get_user_model().objects.create_user(
-            username=self.username,
-            email=email,
-            password=self.password,
-        )
+        self.email = "user@example.com"
+        self.create_user()
+
         self.user.user_id = str(uuid.uuid4())
         self.user.community_terms_accepted = timezone.now()
         self.user.save()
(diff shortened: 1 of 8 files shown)


ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -220,6 +228,32 @@ def login(self):
         self.client.login(username=self.username, password=self.password)
 
 
+class WisdomServiceAPITestCaseBaseOIDC(WisdomServiceAPITestCaseBase):
+    """This should ultimately replace WisdomServiceAPITestCaseBase"""
+
+    def setUp(self):
+        self._inibate_create_user = True
```
manstis, line 235: Is this a typo?

`inibate` vs `inhibit`?

ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -220,6 +228,32 @@ def login(self):
...
+        super().setUp()
+        self.user = create_user_with_provider(
+            username=self.username,
+            email=self.email,
+            password=self.password,
+            provider=USER_SOCIAL_AUTH_PROVIDER_OIDC,
+            rh_org_id=1981,
```
manstis, line 242: Your year of birth? 😺

ansible_ai_connect/main/tests/test_middleware.py
```diff
@@ -26,17 +26,30 @@
...
     MockedMeshClient,
-    WisdomServiceAPITestCaseBase,
+    WisdomAppsBackendMocking,
+    WisdomServiceAPITestCaseBaseOIDC,
 )
 
 
-class TestMiddleware(WisdomServiceAPITestCaseBase):
-    @override_settings(ANSIBLE_AI_ENABLE_TECH_PREVIEW=True)
+def dummy_react_seated_users_data(event, allow_list):
```
manstis, line 34: Typo?

`redact` vs `react`?

manstis (dismissed):
Possibly a couple of typos, otherwise LGTM 👍

ansible_ai_connect/ai/api/tests/test_views.py
goneri: ahah

ansible_ai_connect/main/tests/test_middleware.py
goneri: Damn.

ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -195,12 +198,17 @@ def setUp(self):
         super().setUp()
         self.username = "u" + "".join(random.choices(string.digits, k=5))
         self.password = "secret"
-        email = "user@example.com"
+        self.email = "user@example.com"
+        if hasattr(self, "_inhibit_create_user"):
+            return
+        logger.warn("Please move this test to WisdomServiceAPITestCaseBaseOIDC")
+
         self.user = get_user_model().objects.create_user(
```
manstis, line 206: I was thinking.. and hence my delay in re-reviewing.

Would it not be better to have something like:
```
def setUp(self):
    super().setUp()
    self.username = "u" + "".join(random.choices(string.digits, k=5))
    self.password = "secret"
    self.email = "user@example.com"
    self.user = create_user() # <--- Defer to a method that can be overridden

    self.user.user_id = str(uuid.uuid4())
    self.user.community_terms_accepted = timezone.now()
    self.user.save()

    group_1, _ = Group.objects.get_or_create(name="Group 1")
    group_2, _ = Group.objects.get_or_create(name="Group 2")
    group_1.user_set.add(self.user)
    group_2.user_set.add(self.user)
    cache.clear()
```
where the implementation in `WisdomServiceAPITestCaseBase` is:
```
def create_user(self):
    logger.warn("Please move this test to WisdomServiceAPITestCaseBaseOIDC")
    return get_user_model().objects.create_user(
        username=self.username,
        email=self.email,
        password=self.password,
    )
```
Then override the method in `WisdomServiceAPITestCaseBaseOIDC` with:
```
def create_user(self):
    return create_user_with_provider(
        username=self.username,
        email=self.email,
        password=self.password,
        provider=USER_SOCIAL_AUTH_PROVIDER_OIDC,
        rh_org_id=1981,
        social_auth_extra_data={},
    )
```

manstis (commented):
It seems your use of `_inhibit_create_user` is to allow overriding creation of the `user`?

Would it not be better to move creation to a function that can be overridden?

ansible_ai_connect/ai/api/tests/test_views.py
goneri: Yes, this is indeed way more elegant.

manstis (approved):
Thanks. LGTM :+1:

PR #1196, opened 2024-07-11
limit hosts matched by regex

gebhardtr:
Closes: https://github.com/ansible/ansible-ai-connect-service/issues/1197

The current regexes will more hosts than intended; example:

image


manstis (approved):
Good spot. LGTM :+1:

PR #1200, opened 2024-07-12
unredact inlineSuggestionFeedback for opted in orgs

robinbobbitt:
Jira Issue: https://issues.redhat.com/browse/AAP-20367

## Description
Allows collection of inlineSuggestionFeedback for licensed users who have not opted out of analytics telemetry.

### Steps to test
1. Pull down the PR
2. Test an inlineSuggestionFeedback as a user in a not opted out org
3. inlineSuggestionFeedback flows to Segment
4. Test an inlineSuggestionFeedback as a user in an opted out org
5. Request fails with a 400

### Scenarios tested
As described above. Also, there were already existing unit tests to confirm the serializer enforces inlineSuggestionFeedback not being available to users in opted out orgs, so nothing additional to add.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/tests/test_serializers.py
+++ b/ansible_ai_connect/ai/api/tests/test_serializers.py
@@ -236,10 +236,6 @@ class FeedbackRequestSerializerTest(TestCase):
             },
         )
 
-        # inlineSuggestion feedback raises exception when seat
-        with self.assertRaises(serializers.ValidationError):
-            serializer.is_valid(raise_exception=True)
-
         # inlineSuggestion feedback raises exception when seat and not telemetry enabled
         with self.assertRaises(serializers.ValidationError):
             serializer.is_valid(raise_exception=True)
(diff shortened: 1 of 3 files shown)


manstis (approved):
LGTM :+1: 

Whilst you're testing `someUnallowedFeedback` is not sent to Segment I did not see any tests verifying `inlineSuggestionFeedback` events are sent to Segment for commercial users. Do you think it prudent to add a test for the positive case?

robinbobbitt:
@manstis good question. I had thought about that as well but thought `test_send_segment_event_seated_user` was already covering the positive case and anything additional would essentially just be covering the content of ALLOW_LIST. wdyt? 
Force pushed same commit content, just signed.

manstis:
@robinbobbitt 

> I had thought about that as well but thought `test_send_segment_event_seated_user` was already covering the positive case and anything additional would essentially just be covering the content of `ALLOW_LIST`.

Well.... that test uses the `postprocess` event.. whereas this JIRA/PR is explicitly for `inlineSuggestionFeedback`...

IMO we should probably therefore have a positive test for `inlineSuggestionFeedback`.

PR remains approved. I'll leave it to your judgement.

manstis:
Woot, woot. Merged 🥳

PR #1204, opened 2024-07-12
wca: get api_key and model from plan

goneri:
Get the LLM api_key and default model from the user's plan.


ansible_ai_connect/ai/api/model_client/wca_client.py
```diff
@@ -592,15 +612,18 @@ def post_request():
         playbook = response["playbook"]
         outline = response["outline"]
 
-        if ansible_lint_caller := apps.get_app_config("ai").get_ansible_lint_caller():
+        from ansible_ai_connect.ai.apps import AiConfig
```
manstis, line 615: I'm more intrigued by this change than anything... why the need?

manstis (approved):
LGTM :+1:

ansible_ai_connect/ai/api/model_client/wca_client.py
goneri: It's because the get_app_config returns different type of config depending on the App, and Pyright was not happy because of an undefined method. I'm not totally sure Why the code change triggered this behavior.

PR #1207, opened 2024-07-16
Add backoff exception logs to wca_client

TamiTakamiya:
Jira Issue: n/a. This is for getting more info for [wca-feedback issue#62](https://github.com/rh-ibm-synergy/wca-feedback/issues/62)

## Description

Add exception log when exceptions are thrown while WCA API is processed. The backoff tool provided the retry mechanism, but we have not logged those exceptions and did not have detailed information on them.

## Testing

Unit test cases are provided with the PR.  In addition to test cases and log level setting for WCA API, changes for AMS API (log level setting and test cases) are also provided.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

See above.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_client/wca_client.py
+++ b/ansible_ai_connect/ai/api/model_client/wca_client.py
@@ -15,6 +15,7 @@
 import base64
 import json
 import logging
+import sys
 from abc import abstractmethod
 from typing import Any, Dict, Optional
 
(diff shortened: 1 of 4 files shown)


ansible_ai_connect/ai/api/model_client/wca_client.py
```diff
@@ -200,24 +201,38 @@ def fatal_exception(exc) -> bool:
             # retry on all other errors (e.g. network)
             return False
 
+    @staticmethod
+    def log_backoff_exception(details):
```
manstis, line 205: I wonder whether it would also be helpful to log the cause of the backoff? For example:
```
def log_backoff_exception(details, cause):
  logger.info(f"Caught retryable error for '{cause}' after {details['tries']} tries."
  ...etc
```

manstis (approved):
LGTM :+1: 

Minor comment (to possibly aid usefulness) but nothing wrong with your PR.

ansible_ai_connect/ai/api/model_client/wca_client.py
TamiTakamiya: Thanks for your comment. This method was copied from codes in other repo ([example](https://github.com/airbytehq/airbyte/blob/91b290d372425b27dd6866258cb011e638f75a53/airbyte-integrations/connectors/source-instagram/source_instagram/common.py#L26-L29)) and probably there are some rooms for improvements.

For example, on line 206:
```
_, exc, _ = sys.exc_info()
```
only the second value from `sys.exc_info()` is used while he first value should contain the type of the exception... I think the line 207:
```
logger.info(str(exc))
```
would be sufficient for now, but would like to modify these code (another idea is adding a label to the retry counters) later when we get more information on these exceptions.

PR #1214, opened 2024-07-17
trial: send schema1 event after new subscription

goneri:
Jira Issue: <https://issues.redhat.com/browse/AAP-27313>

## Description

Send a `newTrial` event when the user accepts the Trial period.

## Testing

- Add a `print()` at the top of `ansible_ai_connect.ai.api.utils.segment.base_send_segment_event` to trace the payload we will send, or do something more elegant.
- Enable the Trial with `ANSIBLE_AI_ENABLE_ONE_CLICK_TRIAL=True`
- Enable the Schema1 event transmission with `SEGMENT_WRITE_KEY=whatever`
- Use an OIDC/RHSSO account with an AAP subscript but without any IBM api key
- Connect to the dashboard
- Accept the Trial period
- Check the logs

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/telemetry/test_schema1.py
+++ b/ansible_ai_connect/ai/api/telemetry/test_schema1.py
@@ -1,0 +1,72 @@
+#  Copyright Red Hat
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from datetime import datetime, timedelta
+from unittest import TestCase, mock
+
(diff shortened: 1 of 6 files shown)


ansible_ai_connect/ai/api/telemetry/test_schema1.py
```diff
@@ -0,0 +1,53 @@
...
+        as_dict = event1.as_dict()
+
+        self.assertEqual(as_dict.get("event_name"), None)
+        self.assertFalse(as_dict.get("exception"), False)
+
+
+class TestNewTrial(TestCase):
+    def test_new_trial(self):
+        event1 = NewTrialEvent()
+        self.assertEqual(event1.event_name, "newTrial")
```
robinbobbitt, line 53: We already have an event called "trialExpired" for when a WCA trial's tokens are used up. It might be good to name this something more distinctive, like oneClickTrialStarted, or lightspeedTrialStarted, unless you're planning to add a field including some "plan" information. We should check in with Marty re: what else he needs, like maybe the start date (which granted is the timestamp on the event, but might be nice to be explicit) and the future expiry date.

goneri: Updated. I've renamed the event and I also return the user plans.

romartin (dismissed):
Thanks @goneri , looks a nice change!! LGTM!

romartin (dismissed):

romartin (approved):

PR #1216, opened 2024-07-18
Update openAPI Schema for /me/summary endpoint

SiddharthRajaraman:
Updating openAPI Schema for:
Jira Issue: <https://issues.redhat.com/browse/AAP-26574>

## Description

As a new endpoint is being added, a new serializer must be created as well. This PR creates a new serializer which parses the html content from the /me/markdown endpoint
### Steps to test
1. Pull down the PR
2. Run the service
3. Access localhost:8000/api/v0/me/markdown and ensure content appears
4. Access s[waggerUI](http://localhost:8000/api/schema/swagger-ui/#/me/me_summary_retrieve) and validate example schema

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/tests/test_views.py
+++ b/ansible_ai_connect/main/tests/test_views.py
@@ -206,7 +206,6 @@ class TestMarkdownMe(TestCase):
         content = r.json()["content"]
         expectation = """
         Logged in as: test_user_name
-
         - User Type: Trial
         - Plan: trial of 90 days
         - Expiration: 2024-10-14
--- a/ansible_ai_connect/users/serializers.py
+++ b/ansible_ai_connect/users/serializers.py
@@ -11,7 +11,9 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from textwrap import dedent
 
+from django.conf import settings
 from rest_framework import serializers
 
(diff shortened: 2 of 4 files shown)


goneri:
I suspect you need to run `make update-openapi-schema` after a change like this.

TamiTakamiya (dismissed):
LGTM

SiddharthRajaraman:
> I suspect you need to run `make update-openapi-schema` after a change like this.

@goneri I believe I did

TamiTakamiya (dismissed):
LGTM. Thanks!

TamiTakamiya:
Hmmm.... Code looks good, but unit tests are failing... Why?

SiddharthRajaraman:
@TamiTakamiya I'm not sure.. not only are the tests failing but when testing extension using the service running from this branch, it also fails

TamiTakamiya:
@SiddharthRajaraman It was much more complicated than I thought. Sorry for having approved this without paying much attension to UT failures.

It looks like we need to move the logic to convert a user object to markdown from view.py to serializer.py, which makes sense because a serializer should serialize an object to a serialized form.  Pls take a look at #1220, which I prototyped for this. Thanks!

goneri:
Honestly, I don't understand what this is for. But it sounds strange. The `serializer_class` attribute, is to set a class to load the endpoint parameters, here we don't have any parameter. So it should just be find.

TamiTakamiya (approved):

PR #1219, opened 2024-07-18
Bump ansible-risk-insight to 0.2.7

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-24378

## Description
Bumping ARI does not fix the issue; as the CVE is being disputed by the affected library's maintainers.

That said, bumping to the latest version ensures we as close to being unaffected as possible.
 
## Testing
Not required.

## Production deployment


pyproject.toml
```diff
@@ -9,7 +9,7 @@ version = "0.1.0"
 dependencies = [
   'ansible-core~=2.15.9',
   'ansible-anonymizer~=1.5.0',
-  'ansible-risk-insight~=0.2.6',
+  'ansible-risk-insight~=0.2.7',
```
goneri, line 12: side note: You don't really need to change this one `~= 0.2.6` already includes 0.2.7 and having to tight dependency can create dependency hell for the users.

goneri (approved):

pyproject.toml
manstis: Ah, IDK that. Thanks.

Since `0.2.7` broke a test (from `0.2.6`) it's possibly a good thing to explicitly bump it. WDYT?

I can close this PR if you think it is better to do so.

goneri: No worries, it was a minor observation :-).

PR #1221, opened 2024-07-19
[DRAFT] AAP-TBE-model-id-validation: Consider WCA 204's as a valid model_id

manstis:
**=== DRAFT ===**

Jira Issue: https://issues.redhat.com/browse/AAP-27482

## Description
Currently HTTP204's from WCA are mapped to HTTP503 in validation of the `model_id`. However in these scenarios the `model_id` was found (otherwise we get other exceptions) it just returned an empty response for the `prompt` we [use](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/wca/model_id_views.py#L213-L215) to validate `model_id`'s.

## Testing
This is difficult to test for real; as it needs a model on WCA that, whilst valid, returns HTTP204 for the validation prompt.

### Steps to test
1. Pull down the PR
2. Unit tests only.

## Production deployment

@jameswnl would like to discuss this at the scrum meeting.

### DIFF
--- a/ansible_ai_connect/ai/api/wca/model_id_views.py
+++ b/ansible_ai_connect/ai/api/wca/model_id_views.py
@@ -37,6 +37,7 @@ from ansible_ai_connect.ai.api.exceptions import (
     WcaUserTrialExpiredException,
 )
 from ansible_ai_connect.ai.api.model_client.exceptions import (
+    WcaEmptyResponse,
     WcaInvalidModelId,
     WcaKeyNotFound,
     WcaModelIdNotFound,
@@ -220,10 +221,12 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
     event = None
     start_time = time.time()
     model_id = UNKNOWN_MODEL_ID
+
+    # An OrgId must be present
+    # See https://issues.redhat.com/browse/AAP-16009
+    organization = request._request.user.organization
+
     try:
-        # An OrgId must be present
-        # See https://issues.redhat.com/browse/AAP-16009
-        organization = request._request.user.organization
         if not organization:
             return Response(status=HTTP_400_BAD_REQUEST)
 
@@ -232,6 +235,11 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
 
         validate(api_key, model_id)
 
+        return on_success(organization.id, model_id)
+
+    except WcaEmptyResponse:
+        # An empty response does not mean the model_id is invalid.
+        # More specific exceptions represent an invalid or missing model_id.
         return on_success(organization.id, model_id)
 
     except (WcaInvalidModelId, WcaModelIdNotFound, WcaTokenFailure) as e:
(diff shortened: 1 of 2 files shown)


manstis:
@jameswnl @goneri @robinbobbitt For contemplation :yin_yang:

ansible_ai_connect/ai/api/wca/model_id_views.py
```diff
@@ -220,10 +221,12 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
...
     model_id = UNKNOWN_MODEL_ID
+
+    # An OrgId must be present
+    # See https://issues.redhat.com/browse/AAP-16009
+    organization = request._request.user.organization
+
     try:
-        # An OrgId must be present
-        # See https://issues.redhat.com/browse/AAP-16009
-        organization = request._request.user.organization
```
goneri, line 226: No need to keep the two lines below in the try/except block.

ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
```diff
@@ -276,6 +277,47 @@ def test_set_model_id_throws_validation_exception(self, *args):
...
+        self.user.organization = Organization.objects.get_or_create(id=123)[0]
+        mock_secret_manager = apps.get_app_config("ai").get_wca_secret_manager()
+        mock_wca_client = apps.get_app_config("ai").model_mesh_client
+        self.client.force_authenticate(user=self.user)
+
+        # Set ModelId
+        mock_secret_manager.get_secret.return_value = {"SecretString": "someAPIKey"}
+        mock_wca_client.infer_from_parameters = Mock(side_effect=WcaEmptyResponse)
+        with self.assertLogs(logger="ansible_ai_connect.users.signals", level="DEBUG") as signals:
+            with self.assertLogs(logger="root", level="DEBUG") as log:
```
goneri, line 291: I don't really see the point of checking the log, it's already done multiple times in other tests and it reduces the readability of the test.

ansible_ai_connect/ai/api/wca/model_id_views.py
manstis: Ah, but there is. I originally moved these too but the finally block did not run when the User has no Organisation and tests that check Sement events had been created failed.

ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
```diff
@@ -276,6 +277,47 @@ def test_set_model_id_throws_validation_exception(self, *args):
...
+                    content_type="application/json",
+                )
+
+                self.assertEqual(r.status_code, HTTPStatus.NO_CONTENT)
+                mock_secret_manager.save_secret.assert_called_with(
+                    123, Suffixes.MODEL_ID, "secret_model_id"
+                )
+                self.assert_segment_log(log, "modelIdSet", None)
+
+            # Check audit entry
```
goneri, line 304: It's a bit the same for this part.

ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
manstis: Fair enough. Hasty copy-and-paste.

I can change if we agree this approach is reasonable..  @jameswnl didn't think it desirable so this PR is to gather opinion.

robinbobbitt (commented):
@manstis +1 for this approach of treating a 204 as a positive indication that the model id is valid.

manstis:
Replaced by https://github.com/ansible/ansible-ai-connect-service/pull/1223 that uses correct branch name.

PR #1223, opened 2024-07-22
AAP-27482: Lightspeed admin portal saving custom model-ID return 503

manstis:
**=== REPLACES #1221 WITH CORRECT BRANCH NAME ===**

Jira Issue: https://issues.redhat.com/browse/AAP-27482

## Description
Currently HTTP204's from WCA are mapped to HTTP503 in validation of the `model_id`. However in these scenarios the `model_id` was found (otherwise we get other exceptions) it just returned an empty response for the `prompt` we [use](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/wca/model_id_views.py#L213-L215) to validate `model_id`'s.

## Testing
This is difficult to test for real; as it needs a model on WCA that, whilst valid, returns HTTP204 for the validation prompt.

### Steps to test
1. Pull down the PR
2. Unit tests only.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/wca/model_id_views.py
+++ b/ansible_ai_connect/ai/api/wca/model_id_views.py
@@ -37,6 +37,7 @@ from ansible_ai_connect.ai.api.exceptions import (
     WcaUserTrialExpiredException,
 )
 from ansible_ai_connect.ai.api.model_client.exceptions import (
+    WcaEmptyResponse,
     WcaInvalidModelId,
     WcaKeyNotFound,
     WcaModelIdNotFound,
@@ -220,6 +221,7 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
     event = None
     start_time = time.time()
     model_id = UNKNOWN_MODEL_ID
+    organization = None
     try:
         # An OrgId must be present
         # See https://issues.redhat.com/browse/AAP-16009
@@ -232,6 +234,16 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
 
         validate(api_key, model_id)
 
+        return on_success(organization.id, model_id)
+
+    except WcaEmptyResponse:
+        # An empty response does not mean the model_id is invalid.
+        # More specific exceptions represent an invalid or missing model_id.
+        # organization is guaranteed to be non-None here.
+        logger.info(
+            "WCA returned an empty response validating "
+            f"model_id '{model_id}' for organisation '{organization.id}'."
+        )
         return on_success(organization.id, model_id)
 
     except (WcaInvalidModelId, WcaModelIdNotFound, WcaTokenFailure) as e:
--- a/ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
+++ b/ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
@@ -28,6 +28,7 @@ import ansible_ai_connect.ai.apps
 from ansible_ai_connect.ai.api.aws.exceptions import WcaSecretManagerError
 from ansible_ai_connect.ai.api.aws.wca_secret_manager import AWSSecretManager, Suffixes
 from ansible_ai_connect.ai.api.model_client.exceptions import (
+    WcaEmptyResponse,
     WcaInvalidModelId,
     WcaUserTrialExpired,
 )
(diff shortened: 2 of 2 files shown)


ansible_ai_connect/ai/api/wca/model_id_views.py
```diff
@@ -220,10 +221,12 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
...
+
+    # An OrgId must be present
+    # See https://issues.redhat.com/browse/AAP-16009
+    organization = request._request.user.organization
+
     try:
-        # An OrgId must be present
-        # See https://issues.redhat.com/browse/AAP-16009
-        organization = request._request.user.organization
         if not organization:
```
manstis, line 230: Retrieval of `organization` is kept within the `try` block to ensure the Segment Event is [sent](https://github.com/ansible/ansible-ai-connect-service/pull/1223/files#diff-f2f7efb68b77c2674b8507874c3eb91a12d3d992e643daeaa9e5d4ba70b8610fR277-R288). 

See tests:
- `test_model_id_views.test_validate_error_no_org_id`
- `test_model_id_views.test_set_key_without_org_id`

jameswnl: not sure if I understand this: why are you moving this org-id retrieval outside of the `try`?

ansible_ai_connect/ai/api/wca/model_id_views.py
```diff
@@ -234,6 +237,11 @@ def do_validated_operation(request, api_key_provider, model_id_provider, on_succ
 
         return on_success(organization.id, model_id)
 
+    except WcaEmptyResponse:
+        # An empty response does not mean the model_id is invalid.
```
jameswnl, line 241: can we log that we have got an empty response here?

ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
```diff
@@ -475,3 +497,13 @@ def test_validate_error_model_id_exception(self, *args):
             self.assertEqual(r.status_code, HTTPStatus.SERVICE_UNAVAILABLE)
             self.assertInLog("Exception", log)
             self.assert_segment_log(log, "modelIdValidate", "Exception")
+
+    @override_settings(SEGMENT_WRITE_KEY="DUMMY_KEY_VALUE")
+    def test_validate_error_model_id_empty_response(self, *args):
```
jameswnl, line 502: does `test_set_model_id_empty_response()` already test this same case? Are you doing "one-test-for-one-symptom"? if so, I think we are duplicating the effort for IMO a super-minor benefit - I think we want the test to be as lightweight...

maybe I got it wrong?

jameswnl (commented):
it's there. just a few comments/questions

ansible_ai_connect/ai/api/wca/model_id_views.py
manstis: We log all exceptions [here](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/model_client/wca_utils.py#L48-L57).

ansible_ai_connect/ai/api/wca/tests/test_model_id_views.py
manstis: Validation (and the test for `WcaEmptyResponse`) is performed from two different URLs; one _saving_ the `model_id` the other _validating_ the (already saved) `model_id`. The two tests are for the two code paths: one in `TestWCAModelIdView` (saving) and the other in `TestWCAModelIdValidatorView` (validating).

ansible_ai_connect/ai/api/wca/model_id_views.py
manstis: Moved back. The new `expect` block needed `organization` however I took a different approach.

ansible_ai_connect/ai/api/wca/model_id_views.py
jameswnl: but here you are catching the empty response exception and return `on_success` -- you are not raising it again, how will the exception be handled again?

manstis: For the sole purpose of validating a `model_id` a `WcaEmptyResponse` exception is considered a valid response and hence the exception is swallowed and the `on_success` invoked. Elsewhere, for example [`explanations`](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/views.py#L704-L709), the exception is caught and raised again as an HTTP error.

manstis: @jameswnl Logging added.

jameswnl (approved):

PR #1235, opened 2024-07-25
don't pause trial if we disable ANSIBLE_AI_ENABLE_ONE_CLICK_TRIAL

goneri:
Once the Trials are started, we don't pause them if the `ANSIBLE_AI_ENABLE_ONE_CLICK_TRIAL`
value change.


manstis (approved):
LGTM :+1: 

A unit test would have been nice.

manstis (approved):
LGTM :+1: 

Although I see you don't include similar tests for `BlockUserWithoutSeatAndWCAReadyOrg` or `BlockUserWithSeatButWCANotReady`.

PR #1237, opened 2024-07-26
AAP-27922: /trial: show the days left instead of the end of the trial period

goneri:


ansible_ai_connect/users/views.py
```diff
@@ -173,6 +174,7 @@ def get_context_data(self, **kwargs):
         context["project_name"] = settings.ANSIBLE_AI_PROJECT_NAME
         context["deployment_mode"] = settings.DEPLOYMENT_MODE
         context["has_active_plan"] = has_active_plan
+        context["days_left"] = int((has_active_plan.expired_at - timezone.now()).days)
```
manstis, line 177: This will throw an exception if the User has no active plans, as `has_active_plan` will be `None`.

manstis: If more time has elapsed than the plan duration this will show a negative figure. 

Would you rather limit to a minimum of `0`?

manstis (commented):
It'd be nice to have some tests.

ansible_ai_connect/users/views.py
manstis: Won't the call `has_active_plan.expired_at` fail due to the `None`.. way before you try to cast to an `int`?

ansible_ai_connect/users/views.py
```diff
@@ -170,9 +171,14 @@ def get_context_data(self, **kwargs):
                 else:
                     has_expired_plan = up
 
+        days_left: int = int(
+            ((has_active_plan.expired_at - timezone.now()).days) if has_active_plan else 0
```
manstis, line 175: This an still be negative:
```
>>> from datetime import date
>>> expired_at = date.fromisoformat('2024-01-01')
>>> now = date.fromisoformat('2024-12-31')
>>> int((expired_at - now).days)
-365
```

manstis (commented):
My parting comments before the weekend... Sorry.

ansible_ai_connect/users/views.py
goneri: Yeah, I didn't really pay attention since in this case we don't display the information. But let's fix that.

manstis: Sorry, my bad.. I just looked at [`UserPlan.is_active`](https://github.com/ansible/ansible-ai-connect-service/blob/dda8106987cffba1cd809d703400c2609f0ed111/ansible_ai_connect/users/models.py#L179). I really should get some work to do rather than bother you. Sorry.

manstis (approved):
LGTM :+1: 

:fireworks:

PR #1243, opened 2024-07-30
AAP-28143: Trial users should use Org key is configurated

goneri:
For the case the Org admin set-up the WCA key after the Trial begin, we
use the new WCA key.


manstis (approved):
LGTM :+1:

PR #1244, opened 2024-07-30
trial: properly wrap in a <div/> the expiration message

goneri:
Use a dedicated `<div />` to hold the "Your trial period has expired" message.


manstis (approved):
LGTM 👍

PR #1251, opened 2024-08-08
AAP-26787: API: add a endpoint to list the users

goneri:
Expose a new end-points to list (`/api/v0/users`) and retrieve user information (`/api/v0/users/<PK>`) (e.g `/api/v0/users/1`).

To be able to access the end-point, the user needs to be part of a group with the following permissions:
- `view_user`
-  `view_userplan`
- `view_organization`

This explains how to prepare your user:

```python
from django.contrib.auth import get_user_model
from django.contrib.auth.models import Group, Permission
User = get_user_model()
me = User.objects.get(username="gleboude1@redhat.com")

from django.contrib.auth.models import Permission
wisdom_view_users, _ = Group.objects.get_or_create(name="wisdom_view_users")
wisdom_view_users.permissions.add(Permission.objects.get(codename='view_user'))
wisdom_view_users.permissions.add(Permission.objects.get(codename='view_userplan'))
wisdom_view_users.permissions.add(Permission.objects.get(codename='view_organization'))
me.groups.add(wisdom_view_users)
```

image

### DIFF
--- a/ansible_ai_connect/users/tests/test_views.py
+++ b/ansible_ai_connect/users/tests/test_views.py
@@ -19,13 +19,18 @@ from unittest.mock import ANY, Mock, patch
 
 import boto3
 from django.conf import settings
+from django.contrib.auth.models import Group, Permission
 from django.test import Client, TestCase, override_settings
 from django.urls import reverse
 from rest_framework.test import APITransactionTestCase
 
 import ansible_ai_connect.users.models
 from ansible_ai_connect.main.tests.test_views import create_user_with_provider
-from ansible_ai_connect.test_utils import WisdomAppsBackendMocking, create_user
+from ansible_ai_connect.test_utils import (
+    WisdomAppsBackendMocking,
+    WisdomServiceAPITestCaseBaseOIDC,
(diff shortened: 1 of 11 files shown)


ansible_ai_connect/users/tests/test_views.py
```diff
@@ -487,3 +492,26 @@ def test_no_trial_for_you(self):
...
+        self.client = Client()
+        self.client.force_login(self.user)
+        self.group, _ = Group.objects.get_or_create(name="wisdom_view_users")
+        for i in ["view_user", "view_userplan", "view_organization"]:
+            self.group.permissions.add(Permission.objects.get(codename=i))
+
+    def test_get_list_with_no_permission(self):
+        r = self.client.get(reverse("user_list"))
+        self.assertEqual(r.status_code, 403)
+        print(r.content)
```
manstis, line 510: You probably don't need to print the content; but it does no harm.

manstis (approved):
LGTM :+1: 

Lots of nice new Django things for me to investigate during my review :smile_cat:

PR #1268, opened 2024-08-23
AAP-29861: WCA: Handle 'HTTP404: Failed to get remaining capacity' response

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-29861

## Description
Handle WCA responses for when the WCA instance on which a model is associated has been deleted.

## Testing
Unit tests. There's nothing we can test interactively without mocking WCA's responses which is what the unit tests do.

## Production deployment
  - [PR for VSCode](https://github.com/ansible/vscode-ansible/pull/1519)

### DIFF
--- a/ansible_ai_connect/ai/api/exceptions.py
+++ b/ansible_ai_connect/ai/api/exceptions.py
@@ -140,6 +140,15 @@ class WcaHAPFilterRejectionException(WisdomBadRequest):
     default_detail = "WCA Hate, Abuse, and Profanity filter rejected the request."
 
 
+class WcaInstanceDeletedException(BaseWisdomAPIException):
+    status_code = 418
+    default_code = "error__wca_instance_deleted"
+    default_detail = (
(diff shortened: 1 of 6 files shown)


manstis:
Reviewers sort of chosen at random (or at least because you've had some involvement with code in the affected areas).

ansible_ai_connect/ai/api/exceptions.py
```diff
@@ -74,6 +74,10 @@ class WisdomAccessDenied(BaseWisdomAPIException):
     status_code = 403
 
 
+class WisdomNotFound(BaseWisdomAPIException):
+    status_code = 404
```
goneri, line 78: To be this is more a 5xx error, since the request is valid, the end-point is correct however the backend cannot process the request.

manstis: Careful what you wish for.

goneri (approved):

PR #1270, opened 2024-08-26
container: refresh browserslist

goneri:
Call `npx update-browserslist-db@latest` to avoid a failure later during the `npm run build` if the browser DB is outdated.

```
[1/2] STEP 33/47: RUN npm --prefix /tmp/ansible_ai_connect_admin_portal run build

> admin-portal@0.1.0 build
> node scripts/build.js

Creating an optimized production build...
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```


manstis (approved):
LGTM 👍 

Do you think it should be added to the `makefile` (for building the Admin Portal) too?

goneri:
> LGTM 👍
> 
> Do you think it should be added to the `makefile` (for building the Admin Portal) too?

It's a bit different since the user may not want to run the command again and again.

PR #1275, opened 2024-08-28
Aap 30054

manstis:
**=== DRAFT ===**

Jira Issue: https://issues.redhat.com/browse/AAP-30054

## Description
Ensure we use a version of `micromatch` that has the vulnerability fixed.

## Testing
n/a

## Production deployment


omaciel (approved):
/lgtm

It would be nice to have a set of automated `smoke` tests to give us a higher level of confidence but since we don't seem to have one, I won't block this from merging now :)

PR #1284, opened 2024-08-30
AAP-30180: don't read account info from AMS anymore

goneri:
Remove the method used to read the account information from AMS. We don't need it anymore.


manstis (approved):
I assume the command will have been ran in production before this is merged. 

Bye bye.. it was nice knowing you.

PR #1286, opened 2024-09-02
AAP-27481: Lightspeed Service to prevent users from submitting invalid multitask requests

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-27481

## Description
Stop sending duff multi-task completion requests.

## Testing
See below.

### Steps to test
1. Pull down the PR
2. Run `ansible-ai-connect-service` with this PR applied.
3. Run VSCode with [this](https://github.com/ansible/vscode-ansible/pull/1528) PR applied.
4. Enter some multi-task completion requests (see JIRA for examples).
5. Observe the failure messages presented to Users.

### Scenarios tested
As above.

## Production deployment

See https://github.com/ansible/vscode-ansible/pull/1528 (however not essential for merge of this PR)

### DIFF
--- a/ansible_ai_connect/ai/api/pipelines/completion_stages/pre_process.py
+++ b/ansible_ai_connect/ai/api/pipelines/completion_stages/pre_process.py
@@ -15,9 +15,12 @@
 import logging
 import time
 
+import yaml
 from django.conf import settings
 from django_prometheus.conf import NAMESPACE
 from prometheus_client import Histogram
+from yaml.error import Mark
+from yaml.scanner import ScannerError
 
 from ansible_ai_connect.ai.api import formatter as fmtr
 from ansible_ai_connect.ai.api.exceptions import (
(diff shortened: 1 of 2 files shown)


jameswnl (approved):
So we can't deploy to prod until vs code change is released?

manstis:
@jameswnl 

> So we can't deploy to prod until vs code change is released?

We can, but the _new_ YAML related messages will not be visible to VSCode users.

"This code change requires the following **considerations** before going to production".

"consideration" being the operative word.. we can consider to deploy the service changes without the client.

jameswnl:
> @jameswnl
> 
> > So we can't deploy to prod until vs code change is released?
> 
> We can, but the _new_ YAML related messages will not be visible to VSCode users.
> 
> "This code change requires the following **considerations** before going to production".
> 
> "consideration" being the operative word.. we can consider to deploy the service changes without the client.

In this case, please uncheck that box because checking it implies hard dependency

PR #1288, opened 2024-09-09
requirements: upgrade to cryptography==43.0.1

goneri:
This to address GHSA-h4gh-qq45-vh27.


manstis (commented):
I think you'll need the [requirements-aarch64.txt](https://github.com/ansible/ansible-ai-connect-service/blob/main/requirements-aarch64.txt) requirements files too.

Does [pyproject.toml](https://github.com/ansible/ansible-ai-connect-service/blob/main/pyproject.toml) need looking at?

goneri:
> I think you'll need the [requirements-aarch64.txt](https://github.com/ansible/ansible-ai-connect-service/blob/main/requirements-aarch64.txt) requirements files too.

Oh, thanks.

> Does [pyproject.toml](https://github.com/ansible/ansible-ai-connect-service/blob/main/pyproject.toml) need looking at?

No, the current versions should still work as expected.

goneri:
> > Does [pyproject.toml](https://github.com/ansible/ansible-ai-connect-service/blob/main/pyproject.toml) need looking at?
> 
> No, the current versions should still work as expected.

Actually, I was wrong and we should not even have a dependency on `cryptography` since we don't use it directly.

manstis (approved):
LGTM :+1: 

But Snyk would like you to upgrade Django too!

goneri:
> LGTM 👍
> 
> But Snyk would like you to upgrade Django too!

Thanks for the +1. Yes, I saw that. It's coming in a different PR.

goneri:
Follow up here: https://github.com/ansible/ansible-ai-connect-service/pull/1289

PR #1290, opened 2024-09-09
pip-compile: do a sequential run

goneri:
Avoid triggering the two builds at the same time. This leads to a race condition
during the build on my local set-up.


manstis (approved):
LGTM :+1:

PR #1293, opened 2024-09-10
AAP-27466: CVE-2024-37891 ansible-lightspeed-container: urllib3: proxy-authorization request header is not stripped during cross-origin redirects [ansible_automation_platform-2.4]

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-27466

## Description
Update `urllib3` to (minimum of) `1.26.19`

`pyproject.toml` is used to build downstream AAP.

## Testing
N/A. Minor revision bump.

## Production deployment

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
   'slack-sdk~=3.31.0',
   'social-auth-app-django~=5.4.1',
   'social-auth-core[openidconnect]>=4.4.2',
-  'urllib3~=1.26.18',
+  'urllib3~=1.26.19',
   'uwsgi~=2.0.22',
   'uwsgi-readiness-check~=0.2.0',
   'django-allow-cidr',


goneri:
I don't think we need to do that since the user may have valid reasons to use the older lib version that is compatible with us. Our `pyproject.toml` should only list the dependencies that we need to get our project to work as expected.

manstis:
Hi @goneri 

> I don't think we need to do that since the user may have valid reasons to use the older lib version that is compatible with us. 

It's a single point increment in the revision version.. but something RH are insisting to address the CVE.

We've bumped versions for lessor reasons. Why the sudden concern?

> Our `pyproject.toml` should only list the dependencies that we need to get our project to work as expected.

`pyproject.toml` already lists `urllib3`.

goneri:
> Hi @goneri
> 
> > I don't think we need to do that since the user may have valid reasons to use the older lib version that is compatible with us.
> 
> It's a single point increment in the revision version.. but something RH are insisting to address the CVE.
> 
> We've bumped versions for lessor reasons. Why the sudden concern?
> 
> > Our `pyproject.toml` should only list the dependencies that we need to get our project to work as expected.
> 
> `pyproject.toml` already lists `urllib3`.

It's not a concern, the purpose of the requirements field is to trace the minimal requirements that we have to get our package to work properly. It's not to trace all the CVE of our dependencies. Having too tight version constraints will create different problems.
This is the standard procedure for most of the Python distributions and this is the reason why tools like `pip-audit` may be handy.

goneri (approved):

PR #1295, opened 2024-09-11
Github: enable the authbackend only when needed

goneri:
Only enable `social_core.backends.github.GithubTeamOAuth2` or `social_core.backends.github.GithubOAuth2` when
the associated configuration key are set.


ansible_ai_connect/main/settings/base.py
```diff
@@ -167,7 +167,7 @@
     SOCIAL_AUTH_GITHUB_TEAM_ID = os.environ.get("SOCIAL_AUTH_GITHUB_TEAM_ID") or 7188893
     SOCIAL_AUTH_GITHUB_TEAM_SCOPE = ["read:org"]
     SOCIAL_AUTH_GITHUB_TEAM_EXTRA_DATA = ["login"]
-else:
+elif os.environ.get("SOCIAL_AUTH_GITHUB_KEY"):
     USE_GITHUB_TEAM = False
```
manstis, line 171: `USE_GITHUB_TEAM` is now obsolete and can be removed.

manstis (commented):
You can delete `USE_GITHUB_TEAM` now too.

ansible_ai_connect/main/settings/base.py
goneri: Note: I've another PR pending to do a bigger clean up. The change is a bit too large to fit in this one.

jameswnl (approved):

PR #1297, opened 2024-09-11
reports: move position of "Org has_api_key" column

goneri:
Move the `Org has_api_key` column to the last position.


manstis (approved):
LGTM 👍

PR #1298, opened 2024-09-11
AAP-30596: improve the WcaInstanceDeleted check

goneri:
Use the new and more specific error message from WCA to identify a delete WCA instance.

See: https://github.com/rh-ibm-synergy/wca-feedback/issues/63

### DIFF
--- a/ansible_ai_connect/ai/api/model_client/wca_utils.py
+++ b/ansible_ai_connect/ai/api/model_client/wca_utils.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 import logging
+import re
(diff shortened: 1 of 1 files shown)


manstis (approved):
LGTM 👍 

I suspect (hope) there will be a failing test with this change.... and there is 🎉

manstis:
@goneri I don't understand how you got the tests to pass... from what I recall they should still be failing (typed on my mobile sat on the sofa). What dark magic have you done?

goneri:
Ahah, It's because your tests were already coming with the correct message. I just had to adjust the filter.

PR #1299, opened 2024-09-12
AAP-25344: AAP 2.5 Gateway SSO for Lightspeed

hasys:
Jira Issue: <https://issues.redhat.com/browse/AAP-25344>

## Description
Fixed links for AAP 2.5 Gateway API. Now it supports both AAP Controller and AAP Gateway.

### Steps to test
1. Pull down the PR
2. Ran in on-prem mode
3. Configure AAP 2.5 Gateway
4. Log in into Lightspeed using AAP Gateway
5. Configure AAP 2.4/2.5 Controller
6. Log in into Lightspeed using AAP Controller

### Scenarios tested
Described above (used AAP 2.5 Gateway and Controller)

## Production deployment

### DIFF
--- a/ansible_ai_connect/users/auth.py
+++ b/ansible_ai_connect/users/auth.py
@@ -52,7 +52,7 @@ class AAPOAuth2(BaseOAuth2):
 
     def user_data(self, access_token, *args, **kwargs):
         """Loads user data from service"""
-        url = f"{settings.AAP_API_URL}/v2/me/"
+        url = self.get_me_endpoint(settings.AAP_API_URL)
         resp_data = self.get_json(url, headers={"Authorization": f"bearer {access_token}"})
         return resp_data.get("results")[0]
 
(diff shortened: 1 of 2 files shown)


BrennanPaciorek:
These changes work for me. Should any documentation detailing what to set `AAP_API_URL` to be written?

ansible_ai_connect/users/auth.py
```diff
@@ -67,10 +67,24 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
+        url = self.get_config_endpoint(settings.AAP_API_URL)
         data = self.get_json(url, headers={"Authorization": f"bearer {access_token}"})
         return not data["license_info"]["date_expired"] if "license_info" in data else False
 
+    def get_me_endpoint(self, api_url):
+        """Creates me link to the AAP API depending on the Auth platform"""
+
+        # AAP Controller has /api at the end for API link, AAP Gateway doesn't
+        url = api_url.rstrip("/")
+        return f"{url}/v2/me/" if url.endswith("/api") else f"{url}/api/gateway/v1/me/"
```
omaciel, line 79: Maybe extract `v1` and `v2` and turn them into constants?

ansible_ai_connect/users/tests/test_auth.py
```diff
@@ -112,6 +115,50 @@ def test_missing_values(self):
...
+        authentication = AAPOAuth2()
+        api_url = "http://controller.test"
+        self.assertEqual(
+            "http://controller.test/api/controller/v2/config/",
+            authentication.get_config_endpoint(api_url),
+        )
+
+    def test_get_config_endpoint_gateway_ended(self):
+        authentication = AAPOAuth2()
+        api_url = "http://controller.test"
```
jameswnl, line 156: you meant to have trailing `/` here?

hasys: By default `/` shouldn't be there at all (it is the same now for 2.4, but no `/` after `/api`), but I added rstrip in the `get_config_endpoint` and `get_me_endpoint` just in case.

ansible_ai_connect/users/auth.py
hasys: v1 and v2 are from different APIs one if for v2 is version of API of controller and v1 is version of Gateway (new API).

So next in line is Gateway v2, but no clarity if other part of the link will change as well.

ansible_ai_connect/users/tests/test_auth.py
jameswnl: I am trying to understand the differences of this test case vs that in L146

hasys: Since I check `endswith("/api")` then ending API url with `/` can affect the flow and provide Gateway API instead of expected Controller API.

So if user by mistake put `/` at the end of the API url, it shouldn't break this code. Maybe it will break something in another place but at least not here. 

And there are test cases for it to check the logic in production.

hasys: Sorry @jameswnl, great catch, fixed!

jameswnl (approved):

hasys:
Re-pushed to sign the commit, nothing changed. Waiting until all checks are green and merging. Thank you @jameswnl @omaciel!

PR #1300, opened 2024-09-12
using quay.io/jameswong/git instead of alpine/git

jameswnl:
https://issues.redhat.com/browse/AAP-29987

## Description

`alpine/git` is being used in our Konflux pipeline but docker rate limit once a while will block access and fail our pipeline. Now switching to a clone in quay.io

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment


manstis (dismissed):
LGTM 👍 

Shame you have to use your `jameswong` organisation (should you leave Red Hat, revoke access, or whatever).

manstis (approved):
Thanks.  LGTM 👍

PR #1301, opened 2024-09-12
postman/slack: print the actual error message

goneri:
The message itself is actually in the `body` attribute.

See: https://slack.dev/python-slack-sdk/api-docs/slack_sdk/webhook/webhook_response.html#slack_sdk.webhook.webhook_response.WebhookResponse


manstis (approved):
LGTM :+1:

PR #1302, opened 2024-09-13
users: {community,commercial}_terms_accepted

goneri:
We don't use the fields anymore.


goneri:
I need to be rebased. I will create a new one later.

PR #1303, opened 2024-09-16
AAP-30888: CVE-2024-45590 ansible-lightspeed-container: Denial of Service Vulnerability in body-parser [ansible_automation_platform-2.4]

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-30888

## Description
This PR bumps `body-parser` to `1.20.3` (the fixed version).

It was achieved by running `npm audit fix` that also resolved other issues.

## Testing
Re-compiled the webapp.

Ran locally and verified operation.

### Steps to test
1. Pull down the PR`
2. `cd ./ansible_ai_connect_admin_portal`
3. `npm install`
4. `cd ..`
5. `make admin-portal-bundle`
6. `make start-backends`
7. `make create-application`
8. Run server
9. Login with a User that has Admin permissions
10. Navigate to "Admin Portal".

### Scenarios tested
As above.

## Production deployment
- [x] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1304, opened 2024-09-16
AAP-31326: Add OrgName to the Marketing report

goneri:


ansible_ai_connect/users/reports/generators.py
```diff
@@ -125,6 +125,7 @@ def generate(
                 "First name",
                 "Last name",
                 "Email",
+                "Organization name",
```
manstis, line 128: I imagine having the "Organisation name" column next to "Org ID" may have made Marty's life easier.

manstis (approved):
LGTM :+1: 

Personally, I'd have placed the new column after "Org ID"; but there's nothing in the JIRA.

manstis:
Vulnerabilities scan fix can be found here: https://github.com/ansible/ansible-ai-connect-service/pull/1303

ansible_ai_connect/users/reports/generators.py
goneri: I was following the same order we've got for the other report.

manstis (approved):
LGTM 👍

PR #1311, opened 2024-09-17
reports: use django.utils.timezone.now() to get the current day

goneri:
We want to use the `django.utils.timezone.now()` to get a date with UTC TZ.


manstis (approved):
LGTM 👍 Great catch.

goneri:
I converted this to draft until we get the full story from Marty and some proper unit-tests.

manstis (approved):
The test doesn't assert use of `timezone` and `20999999_my_report.csv` would pass. I suspect the test was written to keep SonarCloud happy.... but it's better than nothing.

goneri:
> The test doesn't assert use of `timezone` and `20999999_my_report.csv` would pass. I suspect the test was written to keep SonarCloud happy.... but it's better than nothing.

Agreed. I'm now doing a parameter validation inside the function.

manstis (approved):
Thanks. LGTM 👍

PR #1315, opened 2024-09-19
generate_users_trials_reports: Clarify that all the dates are UTC

goneri:


manstis (approved):
As if we read the help text... :rofl: 

LGTM :+1:

goneri:
> As if we read the help text... 🤣
> 
> LGTM 👍

Tsss :-)

PR #1323, opened 2024-09-23
AAP-31447: build the filename from the day before created_before

goneri:
Use the data BEFORE the `created_after` parameter to build the report file name.

### DIFF
--- a/ansible_ai_connect/users/reports/postman.py
+++ b/ansible_ai_connect/users/reports/postman.py
@@ -18,6 +18,7 @@ from datetime import datetime
 from typing import List, Optional
 from urllib import parse
 
+from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.utils import timezone
 from oauth2client.service_account import ServiceAccountCredentials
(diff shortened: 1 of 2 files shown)


ansible_ai_connect/users/reports/postman.py
```diff
@@ -218,7 +219,8 @@ def get_drive(self) -> GoogleDrive:
     def create_filename(title: str, report_date: datetime):
         if report_date.tzinfo is None:
             raise ValueError
-        return f"{report_date.strftime('%Y%m%d')}_{title}.csv"
+        day_before = report_date - relativedelta(days=1)
```
manstis, line 222: `report_date` is *created_before*

See https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/users/reports/postman.py#L226

Therefore if you really want "day before `created_after`.." then L#226 should also be changed accordingly.

manstis (commented):
The JIRA title is "... day before *created_after*..." however the code change supports "... day before *created_before*...". One of the other probably needs changing to make JIRA/PR consistent.

manstis (approved):
LGTM :+1:

PR #1324, opened 2024-09-23
Do not send model ID to WCA Playbook Gen/Exp APIs

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-30690>

## Description

While WCA is using different models for (1) Task generation, (2) Playbook generation and (3) Playbook explanation, AI Connect Service/VS Code extension are designed to handle a single model per user. 

With the recent change on WCA, the explanation API stopped working properly since it stopped picking up the default model for certain set of accounts when an incorrect model was specified. 

Even though it was a behavior change on WCA API, AI Connect Service also had better stop sending a model ID, which is for Task generation, when Playbook generation/explanation APIs are called because now we know the specified model ID is not used for those APIs.

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment


manstis (commented):
From what I understand the `MODEL_ID` can be removed and the default used but we should supply the `SPACE_ID` where the full model identifier (we use) is `SPACE_ID|<spegoid>|MODEL_ID` (exact format is probably a little different, I'm not at a computer).

IBM have also said certain exceptions can be thrown if the customer (I assume `SPACE_ID`) does not have Playbook Generation/Explanation enabled.

TamiTakamiya:
@manstis Yes, you're right. I need to re-consider what we can (or cannot) do for this issue...

TamiTakamiya:
@manstis I have updated this PR, but am not 100% clear yet on what the final code should be.  While the discusstion on Slack is underway, I will keep this as a draft PR.

ansible_ai_connect/ai/api/model_client/tests/test_wca_client.py
```diff
@@ -452,6 +455,29 @@ def test_playbook_exp_request_id_correlation_failure(self):
                 request, content="Some playbook", explanation_id=str(DEFAULT_REQUEST_ID)
             )
 
+    def test_get_default_model_id(self):
+        self.wca_client.get_model_id = Mock(
+            return_value="12345678-0000-0000-0000-000000000000<|sepofid|>granite-3b"
+        )
+        default_model_id = self.wca_client.get_default_model_id(self.user, 123)
+        self.assertEqual(default_model_id, "12345678-0000-0000-0000-000000000000<|sepofid|>")
```
manstis, line 463: I don't think the default `MODEL_ID` includes the delimiter.. it should just be the `SPACE_ID`.

ansible_ai_connect/ai/api/model_client/wca_client.py
```diff
@@ -514,6 +517,32 @@ def get_model_id(
...
+        # send it to WCA playbook explanation/generation API may cause an issue. Especially it
+        # can occur the model_is is "granite-3b", which is used only for task generation.
+        #
+        # Following if block is for getting rid of such model_id before calling WCA playbook
+        # explanation/generation API.
+        separator = "<|sepofid|>"
+        index = model_id.find(separator)
+
+        if index > -1:
+            wca_model_id = model_id[(index + len(separator)) :]
```
manstis, line 538: I think this should probably be `model_id[index:]`.

manstis (commented):
I'm in two minds about this.

I know it's a hack to fix use of different `model_id`'s for different features so on that basis it seems ok. However in the IBM sync call today they were undertaking to: (1) consolidate all features into a single model so we would not need to worry, (2) implement the _gateway_ in WCA so the correct model is invoked no matter what `model_id` we send them.

We also spoke about support for configuring the use of different models for different features within Lightspeed to better support testing of different models for different features (e.g. if the playbook generation model is being updated we could configure Lightspeed to use that for playbook generation, but use the existing model for completions etc).

I'll add @jameswnl as a reviewer... might be this is  OK for now but we create a new ticket for future refinement.

ansible_ai_connect/ai/api/model_client/tests/test_wca_client.py
TamiTakamiya: Thank you for your comment. I have updated the code and the test case.

ansible_ai_connect/ai/api/model_client/wca_client.py
TamiTakamiya: Updated the code. Thanks!

TamiTakamiya:
@jameswnl As @manstis pointed out above, this might be a short-lived solution for a particular issue, I think it is ready for review. Please take a look and let me know your thoughts. Thanks!

jameswnl (commented):
I feel we are jumping through hoops on our end.  We probably should request WCA to cleanup their API.

PR #1325, opened 2024-09-23
AAP-31593: One click Trial - schema2 telemetry enablement.

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-31593>

## Description
- Adding `OneClickTrialStarted` for `schema2`telemetry
- Refactoring some `schema2` stuff into `telemetry` package
- `OneClickTrialStarted` event definition => Given the actual use cases (see [AAP-31593](https://issues.redhat.com/browse/AAP-31593)), notice I don't send the `accept_marketing` field for `schema2`, as we already do for `schema1`, and I understand that's not for the use case of telesense team. Other fields' are preserved. Let me know if you think it should be preserved otherwise.

Thanks!

## Scenarios tested
Tested locally against our wisdom staging Segment's  account. Events are flowing.

### Steps to test
1. Pull down the PR
2. Run the service
3. Start a trial
4. Look at Segment, if the `OneClickTrialStarted` has been sent, as expected.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/telemetry/schema2.py
+++ b/ansible_ai_connect/ai/api/telemetry/schema2.py
@@ -24,6 +24,7 @@ class AnalyticsTelemetryEvents(Enum):
     RECOMMENDATION_ACTION = "Recommendation Action"
     PRODUCT_FEEDBACK = "Product Feedback"
     PLAYBOOK_GENERATION_ACTION = "Playbook Generation Action"
+    ONECLICK_TRIAL_STARTED = "OneClickTrial Started"
 
 
 @frozen
@@ -79,6 +80,24 @@ class AnalyticsProductFeedback:
     )
     rh_user_org_id: int = field(validator=validators.instance_of(int), converter=int, default=0)
     model_name: str = field(default="")
+    timestamp: str = field(
+        default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
+    )
+
+
+@frozen
+class OneClickTrialPlan:
+    created_at: str = field(validator=validators.instance_of(str), converter=str)
+    expired_at: str = field(validator=validators.instance_of(str), converter=str)
+    is_active: bool
+    name: str = field(validator=validators.instance_of(str), converter=str)
+    id: int = field(validator=validators.instance_of(int))
+
+
+@frozen
+class OneClickTrialStarted:
+    plans: list[OneClickTrialPlan]
+    rh_user_org_id: int = field(validator=validators.instance_of(int), converter=int)
     timestamp: str = field(
         default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
     )
--- a/ansible_ai_connect/ai/api/telemetry/schema2_utils.py
+++ b/ansible_ai_connect/ai/api/telemetry/schema2_utils.py
@@ -1,0 +1,53 @@
+#  Copyright Red Hat
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+
+from ansible_ai_connect.ai.api.telemetry.schema2 import (
+    AnalyticsTelemetryEvents,
+    OneClickTrialPlan,
+    OneClickTrialStarted,
+)
+from ansible_ai_connect.ai.api.utils.segment_analytics_telemetry import (
+    send_segment_analytics_event,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def oneclick_trial_event_send(user):
+    send_segment_analytics_event(
+        AnalyticsTelemetryEvents.ONECLICK_TRIAL_STARTED,
+        lambda: _oneclick_trial_event_build(user),
+        user,
+    )
+
+
+def _oneclick_trial_event_build(user):
+    if not user.organization:
+        logger.error("Only users within an org are allowed to send trial events.")
+        raise ValueError("No user organization specified. Cannot send OneClickTrialStarted.")
+    return OneClickTrialStarted(
+        plans=[
+            OneClickTrialPlan(
+                created_at=plan.created_at,
+                expired_at=plan.expired_at,
+                is_active=plan.is_active,
+                name=plan.plan.name,
+                id=plan.plan.id,
+            )
+            for plan in user.userplan_set.all()
+        ],
+        rh_user_org_id=user.organization.id,
+    )
--- a/ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
+++ b/ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
@@ -43,8 +43,12 @@ segment_analytics_client = None
 
 
 def meets_min_ansible_extension_version(version) -> bool:
+    """
+    Checks if the extension version, if exists, checks if it satisfies a defined min. version value.
+    Requests outside the extension context, such as one-click trial, assume no version parameter.
+    """
     if not version:
-        return False
+        return True
     minVersion = Version(settings.ANALYTICS_MIN_ANSIBLE_EXTENSION_VERSION)
     try:
         userVersion = Version(version)
--- a/ansible_ai_connect/users/views.py
+++ b/ansible_ai_connect/users/views.py
@@ -31,6 +31,7 @@ from ansible_ai_connect.ai.api.aws.exceptions import (
     WcaSecretManagerMissingCredentialsError,
 )
 from ansible_ai_connect.ai.api.telemetry import schema1
+from ansible_ai_connect.ai.api.telemetry import schema2_utils as schema2
 from ansible_ai_connect.ai.api.utils.segment import send_schema1_event
 from ansible_ai_connect.main.cache.cache_per_user import cache_per_user
 from ansible_ai_connect.users.constants import TRIAL_PLAN_NAME
@@ -227,6 +228,8 @@ class TrialView(TemplateView):
             event = schema1.OneClickTrialStartedEvent()
             event.set_request(request)
             send_schema1_event(event)
+
+            schema2.oneclick_trial_event_send(request.user)
 
         context = self.get_context_data(**kwargs)
         return render(request, self.template_name, context=context)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,18 +1,18 @@
 repos:
 - repo: https://github.com/pycqa/isort
-  rev: 5.12.0
+  rev: 5.13.2
   hooks:
   - id: isort
     name: isort (python)
 - repo: https://github.com/ambv/black
-  rev: 24.3.0
+  rev: 24.8.0
   hooks:
   - id: black
     args: [--safe, --quiet, --line-length, "100"]
     language_version: python3
     require_serial: true
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.4.0
+  rev: v4.6.0
   hooks:
   - id: trailing-whitespace
     language_version: python3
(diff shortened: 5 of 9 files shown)


ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
```diff
@@ -43,8 +43,12 @@
 
 
 def meets_min_ansible_extension_version(version) -> bool:
+    """
+    Checks if the extension version, if exists, checks if it satisfies a defined min. version value.
+    Requests outside the extension context, such as one-click trial, assume no version parameter.
```
romartin, line 48: @goneri Assumming this assumption makes sense to you as well. Let me know otherwise! Thanks!

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
     timestamp: str = field(
         default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
     )
+
+
+@frozen
+class OneClickTrialPlan:
+    created_at: str = field(validator=validators.instance_of(str), converter=str, default="")
+    expired_at: str = field(validator=validators.instance_of(str), converter=str, default="")
+    is_active: bool = False
```
romartin, line 92: @goneri I am not sure how this boolean field works. When does its value become `False`, in which situation/s? Do we really need to export it to analytics / telesense?  Thanks!

goneri: When the plan becomes inactive, we don't remove the association but the `is_active` flag changes from `True` to `False`.

We don't remove the user / plan relationship to be sure the user doesn't start yet another trial.

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
...
+    name: str = field(validator=validators.instance_of(str), converter=str, default="")
```
goneri, line 93: You can remove the `default=""`.

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
     timestamp: str = field(
         default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
     )
+
+
+@frozen
+class OneClickTrialPlan:
+    created_at: str = field(validator=validators.instance_of(str), converter=str, default="")
```
goneri, line 90: You can remove the `default=""`.

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
...
+    expired_at: str = field(validator=validators.instance_of(str), converter=str, default="")
+    is_active: bool = False
+    name: str = field(validator=validators.instance_of(str), converter=str, default="")
+    id: int = 0
```
goneri, line 94: The default to `0` is wrong. `field(validator=validators.instance_of(int))` should be enough

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
...
+
+
+@frozen
+class OneClickTrialStarted:
+    plans: list[OneClickTrialPlan] = field(factory=list, validator=validators.min_len(1))
+    rh_user_org_id: int = field(validator=validators.instance_of(int), converter=int, default=0)
```
goneri, line 100: Here again the `default=0` is wrong IMO, we would certainly prefer to get an exception if the field is not defined..

ansible_ai_connect/ai/api/telemetry/schema2.py
```diff
@@ -82,3 +83,21 @@ class AnalyticsProductFeedback:
...
+    timestamp: str = field(
+        default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
```
goneri, line 102: For the record, you have an UTC datetime that you convert to a naive date string. This may be problematical if it's not properly handled as such later.

You can add a trailing `Z` to the string to clarify that this is indeed an UTC time, e.g:

```
>>> datetime.now(timezone.utc).strftime('%Y-%m-%dT%H:%M:%SZ')
'2024-09-30T15:10:06Z'
```

ansible_ai_connect/ai/api/telemetry/schema2_utils.py
```diff
@@ -0,0 +1,46 @@
...
+            OneClickTrialPlan(
+                created_at=plan.created_at,
+                expired_at=plan.expired_at,
+                is_active=plan.is_active,
+                name=plan.plan.name,
+                id=plan.plan.id,
+            )
+            for plan in user.userplan_set.all()
+        ],
+        rh_user_org_id=None if user.organization is None else user.organization.id,
```
goneri, line 45: `rh_user_org_id` cannot be `None`. We should raise a `ValueError` exception in this case.

ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
goneri: The part that confuse me the most is that this is about the vscode-ansible extension. I was originally assuming wrongly that it was the local Ansible version of the user.  We should be more clear about that.

ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
```diff
@@ -43,8 +43,12 @@
...
+    """
     if not version:
-        return False
+        return True
```
goneri, line 51: This sounds wrong. What do we change this in this PR in particular?

If we don't want to check the version anymore, we should remove the method and the `if meets_min_ansible_extension_version(ansibleExtensionVersion) is False:` from `send_segment_analytics_event()`.

ansible_ai_connect/users/views.py
```diff
@@ -228,5 +229,7 @@ def post(self, request, *args, **kwargs):
             event.set_request(request)
             send_schema1_event(event)
 
+            schema2.oneclick_trial_event_send(request.user)
```
goneri, line 232: As mentioned above, I would prefer if this was done in two different operations, one to get the correct payload, and a second one for the transport. This to avoid a dependency hell where our payload validation part ends up loading the Segment client that it doesn't need.

goneri (changes requested):
c.f comments

ansible_ai_connect/ai/api/telemetry/schema2.py
goneri: I see two options:
- either you can skip the Trial plan when is_active is false, but you will have a hard time later telling how many users have gone through a Trial. This would also create a different between the event structure and our DB structure.
- or we can be consistent and returns the flatten information and just add a filter on the analytics side.

ansible_ai_connect/ai/api/telemetry/schema2.py
romartin: Hey @goneri , as far as I understand calling the formatter, when no parameteres, already does that, as I see this results on Segment:
`
analytics.track('4e5221f7-beb2-41d3-a731-f2a89a188e39', 'OneClickTrial Started', {
    'plans': [
        {
            'created_at': '2024-09-27T21:38:47.165Z',
            'expired_at': '2024-12-26T21:38:47.165Z',
            'is_active': true,
            'name': 'trial of 90 days',
            'id': 1
        }
    ],
    'rh_user_org_id': 17329057,
    'timestamp': '2024-09-27T23:38:47.172Z'
})

So they already contain the trailing 'Z'. 

Please let me know if I am missing something else, or that's already looking fine.

Thx!
`

ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
romartin: hmmm that's a good point....

ansible_ai_connect/ai/api/telemetry/schema2.py
goneri: How awesome. Thanks!

ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
romartin: the intention for this, time ago, was about to ensure some min. version of hte extension was used to send telemtry, otherwise, given issues in that older extension version, it was not working as epxected

At this point, maybe we can assume extension users are already up to date. But maybe there should be cases on some companies where is not so easy to update vscode, and they may be running under older versions. Due to this reason I was preserving this min. version check, if version is available form the request.

So given above consideration, what do you think? can we consider dropping this?

ansible_ai_connect/ai/api/telemetry/schema2_utils.py
```diff
@@ -0,0 +1,46 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from ansible_ai_connect.ai.api.telemetry.schema2 import (
+    AnalyticsTelemetryEvents,
+    OneClickTrialPlan,
+    OneClickTrialStarted,
+)
+from ansible_ai_connect.ai.api.utils.segment_analytics_telemetry import (
+    send_segment_analytics_event,
```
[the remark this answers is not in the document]
romartin: not completely sure what you mean yet :(

I was explicitly trying to decouple the model definition (classes) from the builder and helper objects. This is why I created the `schema2.py`, as for the model classes, no logic at all, easier to ready and manage. Then the `schema2_utils.py `for other stuff, decoupling model from logic.

ansible_ai_connect/users/views.py
romartin: oh I see what you mean now! about my comment above!

romartin: yes I was thinkkng on just adding here 2 operations, get the payload and send the event, but I was at the end tyring to encapsulate mehtods for reusing. Although it is true that it provides less flexibility, it is not maybe necessary

ansible_ai_connect/ai/api/telemetry/schema2_utils.py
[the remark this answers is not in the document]
romartin: oh look my comment below now, it makes sense what you mention now, I can just drop this file and do both operations in the view (consumer) code.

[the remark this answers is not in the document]
goneri: Oh, I realize that `schema2.py` is the file used to actually do the payload validation. So this is ok. This file can remain.

ansible_ai_connect/users/views.py
goneri: Ignore my comment, I didn't realize we had two different files (schema2.py and schema2_utils.py).

ansible_ai_connect/ai/api/utils/segment_analytics_telemetry.py
romartin: I am fine with finding a suitable solution wiht Marty, but we need a change here in order to make the trial events flow right now. So aren't you agree on actual change and reporting a new ticket? WDYT?

ansible_ai_connect/users/views.py
romartin: thanks , anyway it is great brainstorming about this! It is always good to care about code <3

ansible_ai_connect/ai/api/telemetry/schema2.py
romartin: done

ansible_ai_connect/ai/api/telemetry/schema2.py
romartin: done

ansible_ai_connect/ai/api/telemetry/schema2.py
romartin: done

ansible_ai_connect/ai/api/telemetry/schema2_utils.py
romartin: done

ansible_ai_connect/ai/api/telemetry/schema2.py
romartin: done

goneri (approved):

romartin:
Tested new "schema2" event within the analytics pipeline, merging. Thanks!

romartin:
just signed the commit, waiting for builds again.

PR #1330, opened 2024-09-30
AAP-31593: schema2: drop the vscode-ansible extension check

goneri:
We don't need the check anymore.


ansible_ai_connect/main/settings/base.py
```diff
@@ -233,9 +233,6 @@
 # Write key for sending analytics data to Segment. Note that each of Prod/Dev have a different key.
 SEGMENT_WRITE_KEY = os.environ.get("SEGMENT_WRITE_KEY")
 SEGMENT_ANALYTICS_WRITE_KEY = os.environ.get("SEGMENT_ANALYTICS_WRITE_KEY")
-ANALYTICS_MIN_ANSIBLE_EXTENSION_VERSION = os.environ.get(
```
romartin, line 236: If removing this variable, then we can also remove it from ops https://github.com/ansible/ansible-wisdom-ops/pull/1039

romartin (commented):
Thx @goneri LGTM

As I already did some tests here , I can anticipate that you'll have to drop this setting for the tests work

https://github.com/ansible/ansible-ai-connect-service/blob/8ce3b757475a5fed5654f87028a8bddf63914b2d/ansible_ai_connect/main/tests/test_middleware.py#L48

Also added a comment about another PR for the ops (ConfigMap)

PR #1338, opened 2024-10-07
users: drop ams attribute from the model

goneri:
We don't use it anymore.

See: f3a4cdc0d471cdaf83f722aee0bb6445630c1757


manstis (approved):
LGTM :+1: 

Bye, bye old code :wave:

PR #1340, opened 2024-10-10
AAP-30618: refactoring: migrate PlaybookExplanation to the new Schema1 object

goneri:


manstis (approved):
LGTM :+1: 

Your `duration` is now the time between the event being created and converted to a `dict` whereas it was previously the time it took solely to run the explanation. I suspect the difference is not significant and your approach a lot cleaner.

goneri:
> LGTM 👍
> 
> Your `duration` is now the time between the event being created and converted to a `dict` whereas it was previously the time it took solely to run the explanation. I suspect the difference is not significant and your approach a lot cleaner.

Yes, duration is currently a loose concept, once the refactoring will be finish, each view will be associated with an Schema1 event that is initialized as soon as possible. The duration will always be the time spent between the beginning of the view and the event emission.

PR #1341, opened 2024-10-10
AAP-25116: Remove the Commercial group

goneri:
We don't rely on the group anymore. The last log-in of one of its members was more
than 6 months ago.


ansible_ai_connect/healthcheck/backends.py
```diff
@@ -103,6 +102,7 @@ def check_status(self):
         if not self.enabled:
             return
 
+        FAUX_COMMERCIAL_USER_ORG_ID = "9999999999"
```
manstis, line 105: It may be nicer to remove `COMMERCIAL` from the test/temp Organisation parameter name?

manstis (approved):
LGTM 👍

manstis (approved):
LGTM 👍

PR #1333, opened 2024-10-03
AAP-30986: Support deletion of WCA API_KEY from Admin Portal (Frontend)

mabulgu:
Jira Issue: <https://issues.redhat.com/browse/AAP-30986>

## Description
Please see the Jira issue for details. Note that this issue splitted into two: frontend & backend. This PR includes the FE changes work.

## Testing
Follow the following steps. Do not commit anything by mistake as you will need to de-comment some places in the code.
### Steps to test
1. Go to ansible_ai_connect_admin_portal/src/ModelSettingsOverview.tsx:288
2.  Comment the related part (the delete button)
3. Run `npm start` Open the link in browser tab. This is the mocked mode.
4. Test out the deletion. You might need to play with the mock API here (ansible_ai_connect_admin_portal/config/webpack.mock.api.keys.ts:66) to simulate an error.
5. Revert your changes and dont commit/push them by mistake.
6. Run the `npm run test` command to make sure all tests pass.

### Scenarios tested
* Deletion with success
* Deletion with failure

## Production deployment

- [ ] This code change is ready for production on its own
- [x] This code change requires the following considerations before going to production: The backend part of this code should be completed and done. See AAP-32688.

PR #1342, opened 2024-10-11
AAP-32688: Support deletion of WCA API_KEY from Admin Portal (Backend) 

mabulgu:
Jira Issue: <https://issues.redhat.com/browse/AAP-32688>

## Description
See issue: https://issues.redhat.com/browse/AAP-32688

## Testing
* Follow the basic test steps that are in https://github.com/ansible/ansible-ai-connect-service/pull/1333 (but not with mocked backend but with a running backend on your local.)
* Run the `ai.api.wca.tests`. 

### Scenarios tested
* Delete api key and model id
* Delete api key without model id
* Try to delete api key that doesnt exist
* Fail deleting model id and see api key is not deleted afterward.

## Production deployment

- [] This code change is ready for production on its own
- [x] This code change requires the following considerations before going to production: End to end tests should be added with a seperate PR on tests repo (Thanks @jameswnl for the reminder!)

### DIFF
--- a/ansible_ai_connect/ai/api/wca/api_key_views.py
+++ b/ansible_ai_connect/ai/api/wca/api_key_views.py
@@ -246,7 +246,7 @@ class WCAApiKeyView(RetrieveAPIView, CreateAPIView, DestroyAPIView):
 
                     logger.info(f"Deleted API key secret for org_id '{organization.id}'")
                 except Exception as e:
-                    logger.exception(
+                    raise Exception(
                         "Deleted the WCA Model Id, but failed while deleting the WCA Key: " + e
                     )
 
--- a/ansible_ai_connect/ai/api/wca/api_key_views.py
+++ b/ansible_ai_connect/ai/api/wca/api_key_views.py
@@ -180,6 +180,70 @@ class WCAApiKeyView(RetrieveAPIView, CreateAPIView):
 
         return Response(status=HTTP_204_NO_CONTENT)
 
+    @extend_schema(
+        responses={
+            200: OpenApiResponse(description="OK"),
+            401: OpenApiResponse(description="Unauthorized"),
+            403: OpenApiResponse(description="Forbidden"),
+            404: OpenApiResponse(description="Not found"),
+            429: OpenApiResponse(description="Request was throttled"),
+            500: OpenApiResponse(description="Internal service error"),
+        },
+        summary="DELETE WCA key for an Organization",
+        operation_id="wca_api_key_delete",
+    )
+    def delete(self, request, *args, **kwargs):
+        logger.debug("WCA API Key:: DELETE handler")
+
+        organization = None
+        exception = None
+        start_time = time.time()
+        try:
+            organization = request._request.user.organization
+            if not organization:
+                return Response(status=HTTP_400_BAD_REQUEST)
+
+            secret_manager = apps.get_app_config("ai").get_wca_secret_manager()
+            wca_key = secret_manager.get_secret(organization.id, Suffixes.API_KEY)
+            if wca_key is None:
+                return Response(status=HTTP_400_BAD_REQUEST)
+
+            secret_manager.delete_secret(organization.id, Suffixes.API_KEY)
+
+            # Audit trail/logging
+            user_set_wca_api_key.send(
+                WCAApiKeyView.__class__,
+                user=request._request.user,
+                org_id=organization.id,
+                api_key=wca_key,
+            )
+            logger.info(f"Deleted secret for org_id '{organization.id}'")
+
+        except WcaTokenFailureApiKeyError as e:
+            exception = e
+            logger.info(
+                f"An error occurred trying to retrieve a WCA Token for "
+                f"Organization '{organization.id}'.",
+                exc_info=True,
+            )
+            return Response(status=HTTP_400_BAD_REQUEST)
+
+        except Exception as e:
+            exception = e
+            logger.exception(e)
+            raise ServiceUnavailable(cause=e)
+
+        finally:
+            duration = round((time.time() - start_time) * 1000, 2)
+            event = {
+                "duration": duration,
+                "exception": exception is not None,
+                "problem": None if exception is None else exception.__class__.__name__,
+            }
+            send_segment_event(event, "modelApiKeyDelete", request.user)
+
+        return Response(status=HTTP_204_NO_CONTENT)
+
 
 class WCAApiKeyValidatorView(RetrieveAPIView):
     required_scopes = ["read"]
--- a/ansible_ai_connect/ai/api/wca/api_key_views.py
+++ b/ansible_ai_connect/ai/api/wca/api_key_views.py
@@ -39,7 +39,11 @@ from ansible_ai_connect.ai.api.permissions import (
 from ansible_ai_connect.ai.api.serializers import WcaKeyRequestSerializer
 from ansible_ai_connect.ai.api.utils.segment import send_segment_event
 from ansible_ai_connect.ai.api.views import ServiceUnavailable
-from ansible_ai_connect.users.signals import user_set_wca_api_key
+from ansible_ai_connect.users.signals import (
+    user_delete_wca_api_key,
+    user_delete_wca_model_id,
+    user_set_wca_api_key,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


ansible_ai_connect/ai/api/wca/api_key_views.py
```diff
@@ -180,6 +180,70 @@ def post(self, request, *args, **kwargs):
...
+            organization = request._request.user.organization
+            if not organization:
+                return Response(status=HTTP_400_BAD_REQUEST)
+
+            secret_manager = apps.get_app_config("ai").get_wca_secret_manager()
+            wca_key = secret_manager.get_secret(organization.id, Suffixes.API_KEY)
+            if wca_key is None:
+                return Response(status=HTTP_400_BAD_REQUEST)
+
+            secret_manager.delete_secret(organization.id, Suffixes.API_KEY)
```
manstis, line 211: Don't forget to delete the Model ID too.

ansible_ai_connect/ai/api/wca/api_key_views.py
```diff
@@ -180,6 +180,70 @@ def post(self, request, *args, **kwargs):
...
+
+            # Audit trail/logging
+            user_set_wca_api_key.send(
```
manstis, line 214: They've deleted the API Key, so you probably don't want to use `user_set_wca_api_key`.

I'd be tempted to add a new `Signal` so Marty can track deletions if he wishes.

ansible_ai_connect/ai/api/wca/api_key_views.py
```diff
@@ -180,6 +180,70 @@ def post(self, request, *args, **kwargs):
...
+        finally:
+            duration = round((time.time() - start_time) * 1000, 2)
+            event = {
+                "duration": duration,
+                "exception": exception is not None,
+                "problem": None if exception is None else exception.__class__.__name__,
+            }
+            send_segment_event(event, "modelApiKeyDelete", request.user)
+
+        return Response(status=HTTP_204_NO_CONTENT)
```
manstis, line 245: :+1:

manstis (commented):
I know this is a draft; but thought I'd take a quick "Friday afternoon" look.

ansible_ai_connect/ai/api/wca/api_key_views.py
```diff
@@ -180,6 +180,70 @@ def post(self, request, *args, **kwargs):
...
+            # Audit trail/logging
+            user_set_wca_api_key.send(
+                WCAApiKeyView.__class__,
+                user=request._request.user,
+                org_id=organization.id,
+                api_key=wca_key,
+            )
+            logger.info(f"Deleted secret for org_id '{organization.id}'")
+
+        except WcaTokenFailureApiKeyError as e:
```
manstis, line 222: I suspect you copy/pasted the `POST` method implementation.

This exception can be raised when trying to get a WCA Access Token.

You won't need that here.

mabulgu: yep, I did so. I am aware, thanks!

ansible_ai_connect/ai/api/wca/api_key_views.py
mabulgu: I will contact Marty as soon as he comes back. noted this.

manstis (approved):
LGTM :+1: 

I know @jameswnl is concerned about the lack of an atomic transaction around deletion of the two secrets in AWS Secret Manager. I had a read and it does not seem possible to ensure a single transaction around their deletion.

I'm not sure what we can do about this.. other than moving the onus onto Users and having separate support for deleting the API Key and deleting the Model ID. Deleting the API Key but not* the Model ID leaves the system in an inconsistent state.

@jameswnl Thoughts?

mabulgu:
> it does not seem possible to ensure a single transaction around their deletion.

@manstis yep. it's a concern. ˆ actually thought about this but somehow missed to call again among all other things. What I would suggest is: we can try deleting model id first, then if it is successfull we can try deleting the key. If the model deletion fails, then we dont delete the key, but if we can delete the model but not the key, at least we will not end up with sth inconsistent but sth possible (an org can have a key but not a model id set). wdyt?

manstis:
@mabulgu 

> @manstis yep. it's a concern. I actually thought about this but somehow missed to call again among all other things. What I would suggest is: we can try deleting model id first, then if it is successfull we can try deleting the key. If the model deletion fails, then we don't delete the key, but if we can delete the model but not the key, at least we will not end up with sth inconsistent but sth possible (an org can have a key but not a model id set). wdyt?

That seems a reasonable approach.

You might need to consider how we can propagate a scenario when deletion of the Model ID was successful but deletion of the API Key resulted in an error. We should be able to send some form of response/error back that we can use to show an information message to Users... it's an edge case; but might be worth considering.

jameswnl:
@manstis my suggestion to decouple the 2 deletions is meant to simply our logic (handling partial success situations etc). since you guys already did the handwork to handle them. I am fine :)

mabulgu:
@manstis can you recheck when you have time pls?

mabulgu:
@manstis @jameswnl once this goes to stage I will add a test here to cover deletion as well: https://github.com/ansible/ansible-wisdom-testing/blob/main/tests/ui/test_admin_ui.py

Thanks

goneri:
My understanding was that if the api_key is missing, we can safely ignore the presence model_id. This because without the api_key, it won't be used and if the user set a new api_key, the model_id will also be overwritten.

I don't think we need to do an atomic change. Did I miss something?

ansible_ai_connect/ai/api/wca/api_key_views.py
```diff
@@ -233,17 +233,22 @@ def delete(self, request, *args, **kwargs):
...
+                        user=request._request.user,
+                        org_id=organization.id,
+                        api_key=wca_key,
+                    )
+
+                    logger.info(f"Deleted API key secret for org_id '{organization.id}'")
+                except Exception as e:
+                    logger.exception(
+                        "Deleted the WCA Model Id, but failed while deleting the WCA Key: " + e
+                    )
```
goneri, line 251: Since the situation should never happen, I would prefer we reraise the exception.

I know there is yet another `except Exception` just below, but this will be one less to clean up later.

goneri (changes requested):

ansible_ai_connect/ai/api/wca/api_key_views.py
mabulgu: this is a situation that can happen. the call for deleting the api key can fail right after a call for deleting the model id.

> if the api_key is missing, we can safely ignore the presence model_id. 

yes but this requires a change on the UI end and I dont think not showing sth actually existing as not existing on the UI is not a good practice; actually a UI anti-pattern. 

My intention with that logger was different but I need to change it to an exception message as we show the details on the UI modal. My intention is to let the users about this. using a logger is my mistake here.

jameswnl:
> I don't think we need to do an atomic change. Did I miss something?

There's a misunderstanding about the need for atomic change (I probably unintentionally planted this idea into @manstis ' head). 

My suggestion is really about [this](https://github.com/ansible/ansible-ai-connect-service/pull/1342#issuecomment-2414004542)

ansible_ai_connect/ai/api/wca/api_key_views.py
goneri: The exception should never happen, and we hit it in production it has to get has much priority/visiblity as possible. With a except Exception like this, we are taking the risk of hiding a real problem. The user will get a 200 and we will be in the dark until s/he contact the support and we actually take a look at the logs.

Having a err 500 is the right thing to do in case of a problem with AWS. Unfortunately, we have another `try Except` just below, and we should also get right of it later.

mabulgu: we are actually returning a 503 there. by raising an exception within, we catch it at the higher level one and raise a ServiceUnavailable. I did this regarding @manstis following comment:

> You might need to consider how we can propagate a scenario when deletion of the Model ID was successful but deletion of the API Key resulted in an error. We should be able to send some form of response/error back that we can use to show an information message to Users... it's an edge case; but might be worth considering.

@manstis if this is not the way we want to do it pls let me know so we can revert this part and merge this. pls re-check when available.

manstis: @mabulgu @goneri I'd get rid of the inner `try.. except` and let the outer one handle the failure. Users will get a 5xx type notification. I'd also be tempted to remove the clever logic where the `model_id` is deleted first and then the `api_key`. They've requested deletion of the `api_key` and hence IMO we should at least try that first.

It's not ideal, but the solution is becoming increasingly sophisticated for a scenario that *should* never happen.. and if it does, as @goneri states, we'd like the whole thing to blow up loud and clear.

mabulgu: Ok upon request I am reverting the logic I implemented after your review approval @manstis. the inner exception was meant to have an extra message to be shown in the UI. but if we revert the model id deletion we dont need that as well. I am reverting back the changes so that we can merge this today hopefully.

manstis (approved):
LGTM :+1: @mabulgu 

Nice tests.

PR #1344, opened 2024-10-15
AAP-29291: Trial: redirect Anonymous used to the login page

goneri:
If an Anonymous hits /trial, we now redirect it to the Login page. This will
be enough to avoid the err 500.

### DIFF
--- a/ansible_ai_connect/users/views.py
+++ b/ansible_ai_connect/users/views.py
@@ -16,6 +16,7 @@ import logging
 
 from django.apps import apps
 from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
 from django.forms import Form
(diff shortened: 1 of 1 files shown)


ansible_ai_connect/users/views.py
```diff
@@ -158,6 +159,9 @@ def get_trial_plan(self):
         return trial_plan
 
     def dispatch(self, request, *args, **kwargs):
+        if isinstance(request.user, AnonymousUser):
```
manstis, line 162: Me thinks SonarCloud will complain about coverage.

manstis (approved):
LGTM 👍

ansible_ai_connect/users/views.py
manstis: Whaaaaaat. How can it not complain!!

PR #1352, opened 2024-10-17
AAP-32417: Define API end-point for the role generation

hasys:
Jira Issue: <https://issues.redhat.com/browse/AAP-32417>

## Description
Creates new API end point for role generation so it will be possible to start work on VS Code extension

### Steps to test
1. Run unit tests, all should be green
2. Web Service tests in should be green as well https://github.com/ansible/ansible-wisdom-testing

### Scenarios tested
All unit tests are green.

## Production deployment


ansible_ai_connect/ai/api/serializers.py
```diff
@@ -472,12 +472,87 @@ def validate(self, data):
...
+    wizardId = serializers.UUIDField(
+        format="hex_verbose",
+        required=False,
+        label="wizard ID",
+        help_text=("A UUID to track the succession of interaction from the user."),
+    )
+
+    metadata = Metadata(required=False)
+
+    def validate(self, data):
```
goneri, line 540: You can remove the `validate()` method since it does nothing.

ansible_ai_connect/ai/api/serializers.py
```diff
@@ -472,12 +472,87 @@ def validate(self, data):
...
 class GenerationWarningResponseSerializer(serializers.Serializer):
     id = serializers.CharField()
     message = serializers.CharField()
     details = serializers.CharField(required=False)
 
 
+class GenerationRoleFileEntrySerializer(serializers.Serializer):
+    path = serializers.CharField()
+    content = serializers.CharField()
+    file_type = serializers.CharField()
```
goneri, line 553: It should be called `fileType` if we want to continue to respect the camelCase convention.

goneri (changes requested):
Great job @hasys, Just two minor comments.

goneri (approved):

PR #1356, opened 2024-10-17
schema1: set duration and timestamp by default

goneri:
All the events based on `Schema1Event` have now the `duration` and the `timestamp` fields set.


TamiTakamiya (approved):
LGTM

PR #1357, opened 2024-10-18
AAP-33711: Enable SSL connection if parameter is not recognized

hasys:
Jira Issue: <https://issues.redhat.com/browse/AAP-33711>

## Description

Incorrect value in environments variable shouldn't disable the SSL protection.

### Steps to test
1. All unit tests are green

### Scenarios tested

Tests are green

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/settings/base.py
+++ b/ansible_ai_connect/main/settings/base.py
@@ -171,12 +171,19 @@ elif os.environ.get("SOCIAL_AUTH_GITHUB_KEY"):
 
 SOCIAL_AUTH_LOGIN_ERROR_URL = "/unauthorized/"
 
+
+def is_ssl_enabled(value: str) -> bool:
+    """SSL should be enabled if value is not recognized"""
+    disabled = value.lower() in ("false", "f", "0", "-1")
+    return not disabled
+
+
 AAP_API_URL = os.environ.get("AAP_API_URL")
 AAP_API_PROVIDER_NAME = os.environ.get("AAP_API_PROVIDER_NAME", "Ansible Automation Platform")
-SOCIAL_AUTH_VERIFY_SSL = os.getenv("SOCIAL_AUTH_VERIFY_SSL", "True").lower() in ("true", "1", "t")
-ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL = os.getenv(
-    "ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL", "True"
-).lower() in ("true", "1", "t")
+SOCIAL_AUTH_VERIFY_SSL = is_ssl_enabled(os.getenv("SOCIAL_AUTH_VERIFY_SSL", "True"))
+ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL = is_ssl_enabled(
+    os.getenv("ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL", "True")
+)
 SOCIAL_AUTH_AAP_KEY = os.environ.get("SOCIAL_AUTH_AAP_KEY")
 SOCIAL_AUTH_AAP_SECRET = os.environ.get("SOCIAL_AUTH_AAP_SECRET")
 SOCIAL_AUTH_AAP_SCOPE = ["read"]
--- a/ansible_ai_connect/main/settings/tests/test_base.py
+++ b/ansible_ai_connect/main/settings/tests/test_base.py
@@ -1,0 +1,25 @@
+from django.test import TestCase
+
+from ansible_ai_connect.main.settings.base import is_ssl_enabled
+
+
+class TestBaseSettings(TestCase):
+
+    def test_is_ssl_enabled_true(self):
+        self.assertTrue(is_ssl_enabled("True"))
+        self.assertTrue(is_ssl_enabled("TruE"))
+        self.assertTrue(is_ssl_enabled("true"))
+        self.assertTrue(is_ssl_enabled("t"))
+        self.assertTrue(is_ssl_enabled("1"))
+
+    def test_is_ssl_enabled_incorrect_value(self):
+        self.assertTrue(is_ssl_enabled("Yes"))
+        self.assertTrue(is_ssl_enabled("Ano"))
+        self.assertTrue(is_ssl_enabled("No"))
+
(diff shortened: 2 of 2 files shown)


ansible_ai_connect/main/settings/base.py
```diff
@@ -171,12 +171,19 @@
...
 AAP_API_URL = os.environ.get("AAP_API_URL")
 AAP_API_PROVIDER_NAME = os.environ.get("AAP_API_PROVIDER_NAME", "Ansible Automation Platform")
-SOCIAL_AUTH_VERIFY_SSL = os.getenv("SOCIAL_AUTH_VERIFY_SSL", "True").lower() in ("true", "1", "t")
-ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL = os.getenv(
-    "ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL", "True"
-).lower() in ("true", "1", "t")
+SOCIAL_AUTH_VERIFY_SSL = is_ssl_enabled(os.getenv("SOCIAL_AUTH_VERIFY_SSL", "True"))
+ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL = is_ssl_enabled(
+    os.getenv("ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL", "True")
+)
```
goneri, line 186: What about just:
```python
ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL = not os.getenv(
    "ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL"
).lower() in ("true", "1", "t")
```

goneri: The idea of having a function to check if a envvar is Tru-ish makes a lot of sense. But it would be more elegant if it was a generical one, this way we can reuse the function in the rest of the file:

```python
not is_false("ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL")
```

hasys: It's harder to test, and a harder to read. Actually if you set up `ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL` to `True`, you will disable SSL protection with it.

goneri: And what about the `is_false()` `is_true()` functions, this would allow us to simplify 90% of the configuration file.

hasys: Sorry @goneri didn't see the second message, had to refresh the page before answering. Yes, `is_false`, `is_true` sounds tempting but adding to the picture default value and now we have `is_false("SOME_VAR", "True")` which already not so elegant. So, yes sounds interesting, but it wasn't the goal for this PR. Let's discuss it offline, how can we unify it. Thank you!

goneri (changes requested):
My suggestion is to define helper functions like this:

def envvar_is_true(envvar_name: str):
    return  os.getenv(envvar_name, "False").lower() in ["true", "1", "t"]


def envvar_is_false(envvar_name: str):
    return  os.getenv(envvar_name, "True").lower() in ["false", "0", "f"]

```python
SOCIAL_AUTH_VERIFY_SSL = not envvar_is_false("SOCAL_AUTH_VERIFY_SSL")
WCA_ENABLE_ARI_POSTPROCESS = envvar_is_true("WCA_ENABLE_ARI_POSTPROCESS")
```

This way, we don't a gazillion of different functions to unit-tests and we can reuse them everywhere in the configuration file.

hasys:
Hi @goneri, I think it makes code harder to read. For an instance what are the default values in this sample?

```python
SOCIAL_AUTH_VERIFY_SSL = not envvar_is_false("SOCAL_AUTH_VERIFY_SSL")
WCA_ENABLE_ARI_POSTPROCESS = envvar_is_true("WCA_ENABLE_ARI_POSTPROCESS")
```

Which function should I use if I want to add NEW_VAR with default `True` value? What if I want default `False` value?

Unify things sounds good to me, but unification should simplify things, for now I don't see how can we simplify future code reading/writing with those unified functions.

goneri:
> Hi @goneri, I think it makes code harder to read. For an instance what are the default values in this sample?
> 
> ```python
> SOCIAL_AUTH_VERIFY_SSL = not envvar_is_false("SOCAL_AUTH_VERIFY_SSL")
> WCA_ENABLE_ARI_POSTPROCESS = envvar_is_true("WCA_ENABLE_ARI_POSTPROCESS")
> ```
> 
> Which function should I use if I want to add NEW_VAR with default `True` value? What if I want default `False` value?
> 
> Unify things sounds good to me, but unification should simplify things, for now I don't see how can we simplify future code reading/writing with those unified functions.

We are returning boolean, so we don't really need default value for those. It may actually be nice use type hints to clarify that:

```python
def envvar_is_false(envvar_name: str) -> bool:
    return  os.getenv(envvar_name, "True").lower() in ["false", "0", "f"]

SOCIAL_AUTH_VERIFY_SSL: bool = not envvar_is_false("SOCAL_AUTH_VERIFY_SSL")
```

hasys:
It's the best I can imagine right now:

```python
def environ_get_bool(variable_name: str, default_value: bool) -> bool:
    """Convert environment variable to boolean. If value is not recognizable as False
    returns default value. Method was inspired by check of SSL where it's better to
    enable SSL than disable if value is not recognizable."""
    env_var_value = os.getenv(variable_name, str(default_value)).lower()
    recognized_false = env_var_value in ("false", "f", "0", "-1")
    return False if recognized_false else default_value

SOCIAL_AUTH_VERIFY_SSL: bool = environ_get_bool("SOCIAL_AUTH_VERIFY_SSL", True)
ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL: bool = environ_get_bool(
    "ANSIBLE_AI_MODEL_MESH_API_VERIFY_SSL", True
)
```

But I am still not happy with the result. The thing is for SSL we want a specific case to cover: if it is not explicit that value is `False`, enable SSL, which means make the value of variable `True`. For some cases it can be reversed: if it is not explicit that the value is `True`, make it `False`. It means if we want generalize we need a second parse boolean method anyway and methods should be named like `environ_get_bool_make_it_true_if_value_is_not_recognized_as_false` `environ_get_bool_make_it_false_if_value_is_not_recognized_as_true`. I think it is all a big overkill for the case we are trying to solve.

PR #1359, opened 2024-10-22
Enable eslint on chatbot

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-33560>

## Description

Enable eslint on chatbot.  This PR also includes the updated chatbot TypeScript source files, whose lint rule violations were corrected.

### Steps to test
1. Pull down the PR
2. Change directory to `ansible_ai_connect_chatbot`
3. Run
```
npm install
npm run eslint
npm run coverage
```

### Scenarios tested

See above

## Production deployment

### DIFF
--- a/.github/workflows/npm_audit_chatbot.yml
+++ b/.github/workflows/npm_audit_chatbot.yml
@@ -1,0 +1,43 @@
+name: Chatbot - npm-audit
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ansible_ai_connect_chatbot/**
+  pull_request:
+    branches:
+      - main
+    paths:
+      - ansible_ai_connect_chatbot/**
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    env:
+      WORKING_DIRECTORY: ./ansible_ai_connect_chatbot
+
+    steps:
+      - name: Checkout commit
+        uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          cache: 'npm'
+          cache-dependency-path: ${{env.WORKING_DIRECTORY}}/package-lock.json
+
+      - name: Install Dependencies
+        run: npm install
+        working-directory: ${{env.WORKING_DIRECTORY}}
+
+      - name: Run npm audit
+        uses: oke-py/npm-audit-action@v2
+        with:
+          audit_level: low
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          create_issues: false
+          working_directory: ${{env.WORKING_DIRECTORY}}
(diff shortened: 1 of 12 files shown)


ansible_ai_connect_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -196,3 +196,4 @@ export const AnsibleChatbot: React.FunctionComponent = () => {
     </>
   );
 };
+AnsibleChatbot.displayName = "AnsibleChatbot";
```
omaciel, line 199: At some point we will have to check if this should be named Virtual Assistant?

.github/workflows/npm_audit_chatbot.yml
```diff
@@ -0,0 +1,43 @@
...
+      - main
+    paths:
+      - ansible_ai_connect_chatbot/**
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    env:
+      WORKING_DIRECTORY: ./ansible_ai_connect_chatbot
```
mabulgu, line 20: wow is this the chatbot UI path? Why we did not include a suffix as _ui ? (just a related question)

Makefile
```diff
@@ -178,3 +178,8 @@ chatbot-bundle:
 .PHONY: chatbot-test
 chatbot-test:
 	npm --prefix ./ansible_ai_connect_chatbot run coverage
+
+# Run lint checks for Admin Portal
+.PHONY: chatbot-lint
```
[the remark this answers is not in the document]
TamiTakamiya: I just followed the existing entries...  I read [this gnu make documentation](https://www.gnu.org/software/make/manual/make.html#Phony-Targets) and found it is possible to write

> .PHONY: cleanall cleanobj cleandiff

to combine multiple `.PHONY` lines in one, but I think the current implementation (one `.PHONY` per target) is easier to understand which target is `.PHONY` target, even though all targets are `.PHONY` targets in our case. What do you think?

ansible_ai_connect_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
TamiTakamiya: It will be addressed with [AAP-33903](https://issues.redhat.com/browse/AAP-33903). Thanks.

.github/workflows/npm_audit_chatbot.yml
TamiTakamiya: I think the name `ansible_ai_connect_chatbot` is sufficient since we know we implement serivice-related logic in the Python/Django code under `ansible_ai_connect`.

Makefile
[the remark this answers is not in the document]
omaciel: No concerns here :) Don't block on this comment.

mabulgu (approved):
LGTM

.github/workflows/npm_audit_chatbot.yml
mabulgu: not enough for "clean code" to me, but fair enough, thanks!

TamiTakamiya: Thanks!

PR #1364, opened 2024-10-23
Duplicated chatbot message box issue

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-34074>

## Description

The styles applied to the chatbot message box are set by using inline styles.  When the chatbot is loaded from the STAGE server, the following warning message shows up in the development console:

> index-CXRSxDSp.js:55 Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self' data:". Either the 'unsafe-inline' keyword, a hash ('sha256-lmsyJ3MLn/9zJPUSTRtlKyXRD2upigFYpVAUZs6+KS8='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.

Actually the same message shows up when the chatbot is loaded from a local server. However, when it was loaded from a local server, the styles are applied and no issues are found on UI.

This warning message can be suppressed by specifying the following setting for Django CSP:

```
CSP_STYLE_SRC = ("'self'", "'unsafe-inline'")
```
The string 'unsafe-inline' looks scary, but according to [the reply on this Stackoverflow page](https://stackoverflow.com/questions/68459561/content-security-policy-blocks-angular-styles), it should be fine for `CSP_STYLE_SRC`.  It is the change contained in this PR.

### Steps to test
1. Pull down the PR
2. Run the AI Connect service
3. Login through Red Hat SSO and open the chatbot page
4. Open Developer Console and make sure no warning message is found

### Scenarios tested

See above

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/settings/base.py
+++ b/ansible_ai_connect/main/settings/base.py
@@ -539,6 +539,7 @@ WCA_ENABLE_ARI_POSTPROCESS = os.getenv("WCA_ENABLE_ARI_POSTPROCESS", "False").lo
 CSP_DEFAULT_SRC = ("'self'", "data:")
 CSP_INCLUDE_NONCE_IN = ["script-src-elem"]
 CSP_CONNECT_SRC = "'self'"
+CSP_STYLE_SRC = ("'self'", "'unsafe-inline'")
 
 # Region for where the service is deployed. Used by the Health Check endpoint.
 DEPLOYED_REGION = os.getenv("DEPLOYED_REGION") or "unknown"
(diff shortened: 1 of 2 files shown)


ansible_ai_connect/main/settings/base.py
```diff
@@ -539,6 +539,7 @@
 CSP_DEFAULT_SRC = ("'self'", "data:")
 CSP_INCLUDE_NONCE_IN = ["script-src-elem"]
 CSP_CONNECT_SRC = "'self'"
+CSP_STYLE_SRC = ("'self'", "'unsafe-inline'")
```
mabulgu, line 542: Looks good to me as it seems to be a solution but for now I am not approving it till you undraft the issue.

TamiTakamiya:
This will be addressed by the PatternFly virtual assistant team with https://github.com/patternfly/virtual-assistant/issues/263

I am closing this without merging

PR #1365, opened 2024-10-23
AAP-27258: improve the visual when the Trial has expired

goneri:
Return a better message with an URL pointing on IBM's Lightspeed website.

image

image


omaciel (approved):
/lgtm

PR #1369, opened 2024-10-28
Allow test users to use chatbot

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-34721>

## Description

Allow test users to use chatbot in addition to RH employees so that we can start writing E2E/Performance test cases.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests are included in the PR

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/permissions.py
+++ b/ansible_ai_connect/main/permissions.py
@@ -11,22 +11,29 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from rest_framework.permissions import IsAuthenticated
 
 
-class IsAuthenticatedRHEmployeeOrTestUser(IsAuthenticated):
+class IsRHEmployee:
     """
-    Allow access only to users who are authenticated Red Hat employees or
-    test users.
+    Allow access only to users who are Red Hat employees
     """
 
-    code = "permission_denied__user_not_authenticated_rh_employee"
-    message = "The User is not an authenticated Red Hat employee."
+    code = "permission_denied__user_not_rh_employee"
+    message = "The User is not a Red Hat employee."
 
     def has_permission(self, request, view):
-        permitted = super().has_permission(request, view)
-        if not permitted:
-            return permitted
-
         user = request.user
-        return user.rh_employee or "test" in list(user.groups.values_list("name", flat=True))
+        return user.is_authenticated and user.rh_employee
+
+
+class IsTestUser:
+    """
+    Allow access only to test users, who are found in the "test" Django group
+    """
+
+    code = "permission_denied_user_not_test_user"
+    message = "The user is not a Test User"
+
+    def has_permission(self, request, view):
+        user = request.user
+        return user.is_authenticated and user.groups.filter(name="test").exists()
--- a/ansible_ai_connect/main/permissions.py
+++ b/ansible_ai_connect/main/permissions.py
@@ -11,9 +11,10 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from rest_framework.permissions import BasePermission
 
 
-class IsRHEmployee:
+class IsRHEmployee(BasePermission):
     """
     Allow access only to users who are Red Hat employees
     """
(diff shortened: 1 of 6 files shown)


ansible_ai_connect/main/permissions.py
```diff
@@ -28,4 +29,4 @@ def has_permission(self, request, view):
             return permitted
 
         user = request.user
-        return user.rh_employee
+        return user.rh_employee or "test" in list(user.groups.values_list("name", flat=True))
```
jameswnl, line 32: you can also use this expression to check group membership:
`user.groups.filter(name="test").exists()` 
which was used previously in https://github.com/ansible/ansible-ai-connect-service/pull/1341/files#diff-d8b9627bf3e9cc38433bebfcb64ea46bc5dee93c7eaa4e4fc62fe6c3fa1d4341

ansible_ai_connect/main/views.py
```diff
@@ -115,7 +115,7 @@ class ChatbotView(ProtectedTemplateView):
     template_name = "chatbot/index.html"
 
     permission_classes = [
-        IsAuthenticatedRHEmployee,
+        IsAuthenticatedRHEmployeeOrTestUser,
```
jameswnl, line 118: Please create this `TestUser` as a separate permission object and use bitwise `|` like https://stackoverflow.com/questions/35557156/how-to-get-or-permissions-instead-of-and-in-rest-framework

jameswnl (changes requested):

ansible_ai_connect/main/permissions.py
TamiTakamiya: Oh, that's neat. I'll update the code. Thanks!

ansible_ai_connect/main/views.py
TamiTakamiya: Yes. That's cleaner. Thanks.

ansible_ai_connect/users/views.py
```diff
@@ -88,7 +88,12 @@ def get_context_data(self, **kwargs):
         context["documentation_url"] = settings.COMMERCIAL_DOCUMENTATION_URL
 
         user = self.request.user
-        context["rh_employee"] = user.rh_employee if isinstance(user, User) else False
+        if isinstance(user, User):
+            context["rh_employee_or_test_user"] = user.rh_employee or "test" in list(
+                user.groups.values_list("name", flat=True)
```
jameswnl, line 93: Can you update this expression to be the same as the above one?

jameswnl (approved):

ansible_ai_connect/main/permissions.py
```diff
@@ -11,21 +11,30 @@
...
+    Allow access only to test users, who are found in the "test" Django group
+    """
+
+    code = "permission_denied_user_not_test_user"
+    message = "The user is not a Test User"
+
+    def has_permission(self, request, view):
         user = request.user
-        return user.rh_employee
+        return user.is_authenticated and user.groups.filter(name="test").exists()
```
goneri, line 40: The condition is wrong, by returning `False` if the user has no `test` group, it prevents the rh_employee users from connecting.

ansible_ai_connect/main/views.py
```diff
@@ -115,7 +115,7 @@ class ChatbotView(ProtectedTemplateView):
     template_name = "chatbot/index.html"
 
     permission_classes = [
-        IsAuthenticatedRHEmployee,
+        IsRHEmployee | IsTestUser,
```
jameswnl, line 118: @goneri this `|` condition

ansible_ai_connect/main/permissions.py
goneri: update: I'm wrong, cf. @jameswnl comments below.

ansible_ai_connect/users/templates/users/home.html
```diff
@@ -129,7 +129,7 @@ <h1 class="pf-c-title pf-m-lg">{{ project_name }}</h1>
                   {% if deployment_mode == 'saas' and user.is_authenticated and user.rh_user_is_org_admin %}
                   <a class="pf-l-level__item" href="/console"><span class="fas fa-solid fa-cog"></span> Admin Portal</a>
                   {% endif %}
-                  {% if chatbot_enabled and deployment_mode == 'saas' and user.is_authenticated and user.rh_employee %}
+                  {% if chatbot_enabled and deployment_mode == 'saas' and user.is_authenticated and rh_employee_or_test_user %}
```
goneri, line 132: You also need to adjust https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/users/templates/users/trial.html#L134-L136

PR #1327, opened 2024-09-27
Update test cases for AAP-32080

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-32080>

## Description

Updates test cases based on the comments I got for #1326. Except for one line to add a log at DEBUG level, all changes are in test codes.

### Steps to test
1. Pull down the PR
2. Run test cases

### Scenarios tested

This is a PR on test cases.

## Production deployment

- [X] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1373, opened 2024-10-31
AAP-33664: ModelPipelines: Make ModelPipelineFactory use class discovery

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-33664

## Description
Refactor the Model pipeline factory to avoid duplicate boiler plate code and improve extensibility.
 
## Testing
N/A. This PR is a refactoring exercice.

### Steps to test
N/A. This PR is a refactoring exercice.

### Scenarios tested
N/A. This PR is a refactoring exercice. Unit tests should continue to pass.

## Production deployment


ansible_ai_connect/ai/api/model_client/wca_client.py
manstis, line 1: This empty file mysteriously appeared following #1327

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
   'jwcrypto~=1.5.6',
   'jinja2~=3.1.3',
   'langchain~=0.2.5',
+  'langchain-community~=0.2.5',
```
manstis, line 30: There were a few warnings about using a deprecated import so I fixed them.

.gitignore
```diff
@@ -57,7 +57,7 @@ coverage.xml
 
 # Django stuff:
 *.log
-local_settings.py
+local_settings*.py
```
manstis, line 60: I'm using InteliJ 2024.2.3 which seems to favour providing local settings with a Python file.

I have numerous different configurations and this change therefore supports multiple local files.

ansible_ai_connect/ai/api/model_pipelines/__init__.py
manstis, line 1: This forces the different implementations to load (and hence self-register).

`# noqa` is to avoid Flake complaining about unused imports.

goneri (approved):

.gitignore
goneri: You can use instead your own user side `~/.gitignore`.

manstis: Agreed; however since `local_settings.py` was already _ignored_ I just expanded the filter.

PR #1375, opened 2024-11-01
AAP-32019: Enable Playbook generation/explanation endpoint in Lightspeed service for on-prem

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-32019

## Description
This PR adds support for Playbook Generation and Explanation for "on prem".

It is enabled/disabled with the environment variable `ENABLE_PLAYBOOK_ENDPOINT`

## Testing
TBD.

Once the container has been built by this PR I'll deploy an AAIC instance "on prem", test and document.

### Steps to test
1. Pull down the PR
2. Once the container has been built by this PR I'll deploy an AAIC instance "on prem", test and document.

### Scenarios tested
As above.

## Production deployment
- [ ] This code change is ready for production on its own
- [x] This code change requires the following considerations before going to production:

This PR depends on https://github.com/ansible/ansible-wisdom-ops/pull/1096

PR #1377, opened 2024-11-01
AAP-32019: Enable Playbook generation/explanation endpoint in Lightspeed service for on-prem

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-32019

## Description
This PR adds support for Playbook Generation and Explanation for "on prem".

It is enabled/disabled with the environment variable `ENABLE_PLAYBOOK_ENDPOINT`

## Testing
See JIRA for instructions.

### Steps to test
See JIRA for instructions.

### Scenarios tested
As above.

## Production deployment

This PR depends on https://github.com/ansible/ansible-wisdom-ops/pull/1096

### DIFF
--- a/ansible_ai_connect/main/settings/base.py
+++ b/ansible_ai_connect/main/settings/base.py
@@ -612,3 +612,14 @@ CHATBOT_URL = os.getenv("CHATBOT_URL")
 CHATBOT_DEFAULT_PROVIDER = os.getenv("CHATBOT_DEFAULT_PROVIDER")
 CHATBOT_DEFAULT_MODEL = os.getenv("CHATBOT_DEFAULT_MODEL")
 # ==========================================
+
+# ==========================================
+# Playbook Generation/Explanation endpoints
+# ------------------------------------------
+ANSIBLE_AI_ENABLE_PLAYBOOK_ENDPOINT = (
+    os.getenv(
+        "ANSIBLE_AI_ENABLE_PLAYBOOK_ENDPOINT", "False" if DEPLOYMENT_MODE == "on-prem" else "True"
+    ).lower()
+    == "true"
+)
+# ==========================================
--- a/ansible_ai_connect/main/settings/development.py
+++ b/ansible_ai_connect/main/settings/development.py
@@ -71,3 +71,5 @@ WCA_SECRET_DUMMY_SECRETS = os.getenv("WCA_SECRET_DUMMY_SECRETS", "")
 TELEMETRY_ADMIN_DASHBOARD_URL = (
     "https://console.stage.redhat.com/ansible/lightspeed-admin-dashboard"
 )
+
+ANSIBLE_AI_ENABLE_PLAYBOOK_ENDPOINT = True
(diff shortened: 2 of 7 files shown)


manstis:
I closed #1375 as the Konflux Build was repeatedly failing.. 

I suspect it had somehow entered an inconsistent state or something or other (as happens from time to time).

Let's see...

manstis:
<sigh>

image

manstis:
@goneri @jameswnl @hasys This is ready for review.

See the JIRA for instructions how to test.

ansible_ai_connect/main/settings/base.py
```diff
@@ -612,3 +612,9 @@ def is_ssl_enabled(value: str) -> bool:
 CHATBOT_DEFAULT_PROVIDER = os.getenv("CHATBOT_DEFAULT_PROVIDER")
 CHATBOT_DEFAULT_MODEL = os.getenv("CHATBOT_DEFAULT_MODEL")
 # ==========================================
+
+# ==========================================
+# Playbook Generation/Explanation endpoints
+# ------------------------------------------
+ENABLE_PLAYBOOK_ENDPOINT = os.getenv("ENABLE_PLAYBOOK_ENDPOINT", "False").lower() == "true"
```
goneri, line 619: I would prefer to prefix these envvars with `ANSIBLE_AI` (or some other prefix to discuss) to simplify the identification of the different keys, e.g: `ANSIBLE_AI_ENABLE_PLAYBOOK_ENDPOINT`.

Can you turn if on by default? I believe this is how we expect it be be initialized.

manstis: @goneri I'll change the name; no problem.

Changing the default behaviour to "enabled" is however more problematic.

An older version of the Ansible AI Connect Operator would not be aware of the new setting and if a User changes the definition of their "on prem" Ansible AI Connect instance to reference the latest service image they'd get Playbook Generation and Explanation "for free" (as it would default to enabled).

Keeping the default as disabled ensures we don't leak features for "on prem".

I did a [PR](https://github.com/ansible/ansible-wisdom-ops/pull/1096) for `ansible-wisdom-ops` to ensure Playbook Generation and Explanation are enabled for our SaaS offering (I will need to update with the name change) and the _development_ settings enable it too.

Do you see an alternative?

goneri: I understand, what about something like:

```
ENABLE_PLAYBOOK_ENDPOINT = os.getenv("ENABLE_PLAYBOOK_ENDPOINT", "False" if settings.DEPLOYMENT_MODE == "on-prem" else "True")).lower() == "true"
```

jameswnl (approved):

ansible_ai_connect/main/settings/development.py
```diff
@@ -71,3 +71,5 @@
 TELEMETRY_ADMIN_DASHBOARD_URL = (
     "https://console.stage.redhat.com/ansible/lightspeed-admin-dashboard"
 )
+
+ANSIBLE_AI_ENABLE_PLAYBOOK_ENDPOINT = True
```
goneri, line 75: :heart:

goneri (approved):

PR #1384, opened 2024-11-06
Set chatbot message timestamp when a message is received

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-35211>

## Description

Add a timestamp to each message when it is received. Othewise, timestamps are dynamically generated at runtime and vary at every re-draw.

### Steps to test
1. Pull down the PR
2. Run the chatbot locally and see message timestamps stay in fixed values.

## Production deployment


omaciel (approved):
/lgtm

PR #1387, opened 2024-11-08
pip-audit: ignore GHSA-32p4-gm2c-wmch

goneri:
The problem happens only if the `ansible.builtin.user` module is used,
which is not our case. We never call any ansible commands.

See: https://github.com/advisories/GHSA-32p4-gm2c-wmch


manstis (dismissed):
Approved, but it looks like you'll need to reference some more aliases?

manstis (approved):
LGTM :+1:

PR #1388, opened 2024-11-08
remove synk check for every PR update

goneri:
We are reaching the limit of what we can test and we already have similar
tests in place (`pip-audit` and SonarCloud).

The error we get:

```
You have reached your monthly limit of 200 private tests for your red-hat-ansible-lightspeed-service org.
To learn more about our plans and increase your tests limit visit https://snyk.io/plans.
```


manstis (approved):
💀 Bye, bye.

PR #1401, opened 2024-11-15
Chatbot debug mode

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-35616>

## Description

Enable "debug" chatbot UI that allows end-user to pick up a model to be used. This time it only allows to use the default `granite-8b` model and the `granite3-8b` model, which was installed with AAP-35616,

image

This is enabled only when the chatbot UI is loaded with `?debug=true` query parameter, e.g.

https://stage.ai.ansible.redhat.com/chatbot?debug=true

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit tests are included

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/views.py
+++ b/ansible_ai_connect/main/views.py
@@ -125,8 +125,6 @@ class ChatbotView(ProtectedTemplateView):
             and settings.CHATBOT_DEFAULT_MODEL
             and settings.CHATBOT_DEFAULT_PROVIDER
         ):
-            debug = request.GET.get("debug", "false")
-            self.debug = debug.lower() == "true"
             return super().get(request)
 
         # Otherwise, redirect to the home page.
@@ -138,7 +136,7 @@ class ChatbotView(ProtectedTemplateView):
         user = self.request.user
         if user and user.is_authenticated:
             context["user_name"] = user.username
-        context["debug"] = "true" if self.debug else "false"
+        context["debug"] = "true" if settings.CHATBOT_DEBUG_UI == "true" else "false"
 
         return context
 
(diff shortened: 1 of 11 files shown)


jameswnl (commented):
How about having the debug switch set in backend? And so we can turn on for stage only

TamiTakamiya:
@jameswnl Updated the code to enable the debug UI via a setting (`CHATBOT_DEBUG_UI`).  Also opened https://github.com/ansible/ansible-wisdom-ops/pull/1124 for enabling the debug UI in STAGE. Thanks.

ansible_ai_connect/main/views.py
```diff
@@ -136,6 +136,7 @@ def get_context_data(self, **kwargs):
         user = self.request.user
         if user and user.is_authenticated:
             context["user_name"] = user.username
+        context["debug"] = "true" if settings.CHATBOT_DEBUG_UI == "true" else "false"
```
manstis, line 139: This could simply be `context["debug"] = settings.CHATBOT_DEBUG_UI`.

manstis (dismissed):
LGTM 👍 

IDK if you'd like @jameswnl to review ad he requested the changes.

ansible_ai_connect/main/views.py
TamiTakamiya: Thanks for your comment. This code was wrong because `settings.CHATBOT_DEBUG_UI` is boolean and cannot have "true" as a value... I have fixed it in the newest commit.

manstis (approved):
LGTM 👍

PR #1403, opened 2024-11-18
Clear context button

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-34874>

## Description

Add "Clear context" button to Chatbot UI so that users can start a new conversation without reloading the browser window.
image

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit test cases are provided.

## Production deployment

### DIFF
--- a/ansible_ai_connect_chatbot/package-lock.json
+++ b/ansible_ai_connect_chatbot/package-lock.json
@@ -902,11 +902,10 @@
       }
     },
     "node_modules/@eslint/plugin-kit": {
-      "version": "0.2.2",
-      "resolved": "https://registry.npmjs.org/@eslint/plugin-kit/-/plugin-kit-0.2.2.tgz",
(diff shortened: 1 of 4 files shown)


mabashian (approved):
LGTM.  My only feedback would be to consider pulling the context button component definition out into a separate file.  I realize that we don't anticipate that component needing to be reused but my concern would be that the AnsibleChatbot component file continues to grow in size and becomes difficult to read/understand.  Nothing functionally wrong with the definition in the same file though so I'll leave it up to you @TamiTakamiya!

TamiTakamiya:
> LGTM. My only feedback would be to consider pulling the context button component definition out into a separate file. I realize that we don't anticipate that component needing to be reused but my concern would be that the AnsibleChatbot component file continues to grow in size and becomes difficult to read/understand. Nothing functionally wrong with the definition in the same file though so I'll leave it up to you @TamiTakamiya!

@mabashian  anticipate that this code will be removed soon when we start supporting the conversation history and PF6 virtual asistant already has its own "New chat" button implementation. Although I will keep it as it is, I appreciate your comment.

TamiTakamiya:
FYI.  When we add the conversation history support, we will use this conversation history drawer provided by PF6 virtual assistant:
Screenshot from 2024-11-17 20-03-36

PR #1409, opened 2024-11-20
Upgrade aiohttp to 3.10.11

manstis:

## Description
```
--- | --- | --- | --- | ---
aiohttp | 3.10.2 | GHSA-8495-4g3g-x7pr | 3.10.11 | ### Summary The Python parser parses newlines in chunk extensions incorrectly which can lead to request smuggling vulnerabilities under certain conditions.
```

## Production deployment


romartin (commented):
shouldn't we be updating also `pyproject.toml?`

PR #1418, opened 2024-11-22
rolegen: improve the outline sampe of the "dummy" pipeline

goneri:
Use a dedicated outline sample for the role generation.


manstis (approved):
LGTM :+1:

PR #1428, opened 2024-11-26
AAP-35857: The image push should not happen if the synk check is not successfull

mabulgu:
Jira Issue: <https://issues.redhat.com/browse/AAP-35857>

Instead of avoiding the build in case of snyk failure, we eitherway build the image but not tag it with the required tag. this way we can avoid the app being deployed as the argocd app checks for the `x.y.z.` tag to deploy the application. this is only for push pipeline.

While this issue only requires the push pipeline to be updated, I implemented the same for the PR pipeline because for the PRs it doesnt also make sense to tag the image with its own pr tag even if it is not deployed. this will also prevent any pr image to be deployed in case we use the "pr-deployment" tag.

## Testing
You can create an empty PR that is drived from this PR branch and see the results there, however you should mimic a failure on the snyk check. I mimiced it by changing the condition to sth irrelevant than "SUCCESS" to check if the rest of the pipeline is skipped if the condition doesnt match. I dont think there is an easy way to test this for PR case. For push case, the only way to test it to merge this PR.

## Production deployment


.tekton/ansible-ai-connect-service-push.yaml
```diff
@@ -271,7 +271,7 @@ spec:
       - name: TARGET_STAGE
         value: production
       - name: BUILD_ARGS
-        value: ["IMAGE_TAGS=latest 1.0.$(tasks.git-metadata.results.commit-timestamp)", "GIT_COMMIT=$(tasks.clone-repository.results.commit)"]
```
jameswnl, line 274: when will this tag be applied? we need this tag to be applied when snyk check passes so that the image autoupdates will pick this up.

.tekton/ansible-ai-connect-service-pull-request.yaml
```diff
@@ -240,7 +240,7 @@ spec:
       - name: TARGET_STAGE
         value: production
       - name: BUILD_ARGS
-        value: ["IMAGE_TAGS=pr-{{pull_request_number}} pr-{{pull_request_number}}.$(tasks.git-metadata.results.commit-timestamp)", "GIT_COMMIT=$(tasks.clone-repository.results.commit)"]
```
jameswnl, line 243: we don't need to change PR workflow since it is not related

mabulgu: see my notes in the PR explanation pls

.tekton/ansible-ai-connect-service-push.yaml
mabulgu: please see the `apply-tags` task and the condition added into it.

mabulgu:
@jameswnl thanks for your review. I responded on your comments as you might have missed some changes I made on the `apply-tags` task and my comment in the description of the PR for doing this for the PR pipeline as well for more consistency between the pipeline codes (it wont change the results for PR explanation anyways)

mabulgu:
@hasys would you mind taking a look at this an approve if possible. James seems to be quite busy so I need an approval to this one in order to move forward. Since now these things are also relate to you, I thought it would be a good start to work together:D

jameswnl:
Hi Aykut, i [slacked](https://redhat.enterprise.slack.com/archives/C07L4FBTGLE/p1733517860577969?thread_ts=1733517318.906719&channel=C07L4FBTGLE&message_ts=1733517860.577969) you yesterday 

basically my question is: your PR has removed the x.y.z format tag from the workflow files, how can that be applied to the image? Am i missing something here?

.tekton/ansible-ai-connect-service-push.yaml
hasys: @jameswnl in `apply-tags` section: 
```yaml
 - name: apply-tags
      [...]
      - name: ADDITIONAL_TAGS
        value: ["pr-{{pull_request_number}}", "pr-{{pull_request_number}}.$(tasks.git-metadata.results.commit-timestamp)"]
      [...]
      when:
        - input: $(tasks.sast-snyk-check.results.TEST_OUTPUT)
          operator: in
          values:
            - '"result":"SUCCESS"'

hasys (dismissed):
We had a conversation with @mabulgu, it is not completely make sense to me to block the pipeline if some issue is found in older code (which is the most common situation from my experience). Since we already have checks before PR merge, all new issues highly likely are found in already existing and deployed code (mostly already deployed dependencies).
image

So, not sure we should complicate automation and processes due to it.

But as @mabulgu pointed we have the similar approach in other repositories, so I think we should proceed consistently now, and fix the process in the future if needed.

Looks good to me thank you @mabulgu !

mabulgu:
> Hi Aykut, i [slacked](https://redhat.enterprise.slack.com/archives/C07L4FBTGLE/p1733517860577969?thread_ts=1733517318.906719&channel=C07L4FBTGLE&message_ts=1733517860.577969) you yesterday
> 
> basically my question is: your PR has removed the x.y.z format tag from the workflow files, how can that be applied to the image? Am i missing something here?

@jameswnl the `apply-tag` task does apply the image later on like I explained above. I will reach out to you for a f2f chat.

mabulgu:
Regarding our discussion with @jameswnl and @hasys, we decided to pivot this issue a bit and agreed on the following:

1- The pull request pipeline should stay as suggested in the PR as it already shows a failure in case of any problem within snyk check. the PR owner should take care of the issue then.
2- The push pipeline should be updated to not block the deployment, but send a slack message to the channel in case of a failure in snyk check. 

Item 2 should be implemented so I am working on this right now.

mabulgu:
Seems that we already have the mechanism for a slack message in case of a failurein the push pipeline. I reverted the push pipeline to not to block any deployment and I think we are good to go

hasys (approved):
Looks good to me, thank you @mabulgu

PR #1439, opened 2024-12-01
Update pip-compile tools

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-36625>

## Description

Currently `make pip-compile` runs (docker|podman)-compose command and updates requirements files both for x86_64 and aarch64 platforms. While using it, I sometimes see the following errors:
```
[pip-compile-x86_64]  | cannot open `/run/user/4212412/crun/4241da4bacfafe5feb3d5e540e45965f85efeaec73eec74fbeb3e67b3e5870b0/exec.fifo`: No such file or directory
[pip-compile-x86_64]  | Error: unable to start container 4241da4bacfafe5feb3d5e540e45965f85efeaec73eec74fbeb3e67b3e5870b0: `/usr/bin/crun start 4241da4bacfafe5feb3d5e540e45965f85efeaec73eec74fbeb3e67b3e5870b0` failed: exit status 1
```
While we need to use containers for supporting multiple platforms, there is no reason that we need to run them 
with (docker|podman)-compose command. This PR will the compose task into two separate make targets: `pip-compile-x86_64` and `pip-compile-aarch64`.  The `pip-compile` target is changed to depend on these two targets so that `make pip-compile` still generates requirements files with a singile command.

In addition to the changes in Makefile, two small changes were made on `pip-compile.sh` file:
1. Added a `pip install --upgrade pip` to upgrade the pip tool itself.
2. Added `-c requirements-${TARGET}.txt` to the `pip-compile` for processing development dependencies so that the same dependencies used for runtime are used for development as well.

### Steps to test
1. Pull down the PR
2. Run `make pip-compile` both on x86_64 Linux and ARM-based Mac.

### Scenarios tested

I tested this only on x86_64 Linux.

## Production deployment


TamiTakamiya:
I have decided to merge this into #1440.  Closing without merging.

PR #1443, opened 2024-12-05
AAP-37014: Make Wisdom's Chatbot interfaces available to RH employees only

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-37014>

## Description
- Make Wisdom's Chatbot interfaces available to RH employees only
- Assume a RH employee when it has the `redhat:employees` realm role and it's email is from `redhat.com` domain

## Testing
Tested locally against an employee user, and a non-employee user. 
Tested both chatbot view and API endpoint.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/views.py
+++ b/ansible_ai_connect/ai/api/views.py
@@ -98,6 +98,7 @@ from ansible_ai_connect.ai.api.utils.segment_analytics_telemetry import (
 )
 from ansible_ai_connect.users.models import User
 
+from ...main.permissions import IsRHEmployee, IsTestUser
 from ...users.throttling import EndpointRateThrottle
 from ..feature_flags import FeatureFlags
 from .data.data_model import ContentMatchPayloadData, ContentMatchResponseDto
@@ -1119,6 +1120,7 @@ class Chat(APIView):
     permission_classes = [
         permissions.IsAuthenticated,
         IsAuthenticatedOrTokenHasScope,
+        IsRHEmployee | IsTestUser,
     ]
     required_scopes = ["read", "write"]
     throttle_classes = [ChatEndpointThrottle]
--- a/ansible_ai_connect/ai/api/tests/test_views.py
+++ b/ansible_ai_connect/ai/api/tests/test_views.py
@@ -3975,6 +3975,7 @@ class TestChatView(WisdomServiceAPITestCaseBase):
         super().setUp()
         (org, _) = Organization.objects.get_or_create(id=123, telemetry_opt_out=False)
         self.user.organization = org
+        self.user.rh_employee = True
 
     @staticmethod
     def mocked_requests_post(*args, **kwargs):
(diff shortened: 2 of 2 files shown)


ansible_ai_connect/ai/api/views.py
```diff
@@ -1117,8 +1118,7 @@ class ChatEndpointThrottle(EndpointRateThrottle):
         scope = "chat"
 
     permission_classes = [
-        permissions.IsAuthenticated,
-        IsAuthenticatedOrTokenHasScope,
+        IsRHEmployee | IsTestUser,
```
romartin, line 1121: Guys, this means a token scoped session only cannot longer the chat API, is that correct? May it break integration tests?
@jameswnl @goneri @TamiTakamiya

TamiTakamiya: Please keep exiting two `permission_classes`
```
        permissions.IsAuthenticated,
        IsAuthenticatedOrTokenHasScope,
```
The first `permissions.IsAuthenticated` is actually redundunt since both `IsRHEmployee` and `IsTestUser` checks `user.is_authenticated` internally, but I think we'd better keep them to be consistent with other views and the redundancy is negligible.

TamiTakamiya (changes requested):
I left two comments. Pls take a look. Thanks!

ansible_ai_connect/ai/api/views.py
romartin: fixed, thanks!

ansible_ai_connect/users/pipeline.py
```diff
@@ -106,7 +111,7 @@ def redhat_organization(backend, user, response, *args, **kwargs):
     user.email = payload.get("email")
     user.email_verified = payload.get("email_verified")
     user.rh_user_is_org_admin = "admin:org:all" in roles
-    user.rh_employee = "redhat:employees" in roles
+    user.rh_employee = "redhat:employees" in roles and is_rh_email_domain(user.email)
```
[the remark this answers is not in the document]
romartin: Hey @TamiTakamiya ,I have added the email_verified check as well, but then I CAN'T access the chatbot link with my RH username. It looks or mine is not verified, which is weird, or we are not properly loading the value.

Can you please @TamiTakamiya try with your user??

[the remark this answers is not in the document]
romartin: fixed

TamiTakamiya (approved):
LGTM! Thanks!!

ansible_ai_connect/users/pipeline.py
```diff
@@ -84,6 +84,17 @@ def github_get_username(strategy, details, backend, user=None, *args, **kwargs):
         return get_username(strategy, details, backend, user, *args, **kwargs)
 
 
+def is_rh_user(user):
```
goneri, line 87: Unlike with the current user.rh_employee, the email may not have been validated. Anyone from outside Red Hat can an Org and manually create a user with a @redhat.com address.

romartin:
As I understand per @goneri 's feedback, we can just rely on the employee role, as other checks are done by the IDP. 
This way, just done the fix for the permission endpoint.
Thx!

goneri (approved):

romartin:
@mabulgu do you know why konflux build is failing here?? :(

romartin:
rebased

PR #1444, opened 2024-12-06
Rename env var used to configure ModelPipelines

manstis:

## Description
A simple internal refactor to rename an environment variable from what is implied WCA usage; to a general form.

## Testing
N/A

## Production deployment


manstis:
@hasys AFAIK you may be the only person (one of the only people) who configure the service locally for Ollama.

This may affect whatever you have local (but not merged?)

goneri (approved):

PR #1453, opened 2024-12-09
one_click_trial: non-rhsso users cannot get Trial access

goneri:
For instance, the AAP users should not get access to the Trial.


manstis (approved):
LGTM :+1: 

I suspect SonarCloud will want a test?

goneri:
> LGTM 👍
> 
> I suspect SonarCloud will want a test?
> 
> .. and no it does not 😸

It was really just to avoid the ugly backtrace when the user comes from AAP. Overall, nothing change from a user perspective.

PR #1456, opened 2024-12-10
pyproject.toml: upgrade the requirements

goneri:
Upgrade the requirements to get a recent version of langchain, the older versions won't work as expect.


manstis (approved):
LGTM 👍

PR #1457, opened 2024-12-11
help script to quickly prepare a test token

goneri:


tools/scripts/get-user-token.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+username=$1
+
+if [ -z "${username}" ]; then
+    echo "Usage ${0} my-username"
+fi
+
+podman --remote exec --user=root -it docker-compose_django_1 wisdom-manage createtoken --username ${username} --duration 6000000 --token-name my-dev-token
```
manstis, line 9: It might have been nice(r) to have done something like this:

https://github.com/ansible/ansible-ai-connect-service/blob/main/Makefile#L5-L10

manstis (approved):
LGTM :+1:

PR #1460, opened 2024-12-13
AAP-37349: ChatBot - Fix using the proper feedback endpoint.

romartin:
Sorry just a small fix that prevents to properly use the feedback button in the chatbot UI.


jameswnl (approved):
this is probably be set in an `env` var?
lgtm for now

romartin:
yes, the value is set via env, and this is just an small dev code piece thing that, once running the UI from local, it points some local mock server, otherwise, it points the "real" prod endpoints

PR #1461, opened 2024-12-13
Update trial messages in both vscode extension and trial web page

mabashian:
Jira Issue: <https://issues.redhat.com/browse/AAP-36175>

## Description
This updates the messages that we present to the user when a trial will expire in <= 90 days as well as when a trial is expired.

## Testing
I tested out all of the various scenarios by manipulating the expired_at timestamp on the user plan.  There are 4 options/scenarios:
1. A trial hasn't started yet
2. Expired (date set to something in the past)
3. About to expire (date set to something within the next 90 days)
4. Not about to expire (date set to something more than 90 days out).  Theoretically this shouldn't be possible but it's a scenario we already cover so I tested it anyway.

Screenshot 2024-12-13 at 1 43 13 PM
Screenshot 2024-12-13 at 1 42 46 PM
Screenshot 2024-12-13 at 1 42 12 PM
Screenshot 2024-12-13 at 1 41 05 PM

Screenshot 2024-12-13 at 2 51 27 PM
Screenshot 2024-12-13 at 2 50 54 PM

### Scenarios tested
See above

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/tests/test_views.py
+++ b/ansible_ai_connect/main/tests/test_views.py
@@ -201,6 +201,12 @@ class TestMarkdownMe(TestCase):
         Logged in as: test_user_name<br>
         Plan: trial of 90 days<br>
         Expiration: {expired_at}
+
+            <br>Accelerate Playbook creation with AI-driven content recommendations
+            from <b>IBM Watsonx Code Assistant for Red Hat Ansible Lightspeed</b>,
+            enabling faster, more efficient automation development. <a href=
+            "https://www.ibm.com/products/watsonx-code-assistant-ansible-lightspeed">
+            Learn more</a>.
         """
         self.assertEqual(dedent(expectation).strip(), content)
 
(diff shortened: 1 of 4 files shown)


manstis:
I'll have a better look Monday @mabashian but it looks like you'll need some unit tests for the logic. It'd be nice too, in my opinion, if the text could wrap on the HTML page. I also notice the three buttons/links are spread far apart. The _footer_ HTML is different on some of the other pages which makes the buttons group better. They're in the same container as the long text at the moment (The other _login_ pages uses a separate container IIRC).

mabashian:
@manstis 👍 I'll take a look at that formatting as well

mabashian:
@manstis OK I think I've got the footer spacing issue squared away:

Screenshot 2024-12-16 at 2 45 05 PM

Tell me more about `It'd be nice too, in my opinion, if the text could wrap on the HTML page` - I think the text is wrapping down to a new line when the screen is more narrow but maybe this isn't working for some browsers?

manstis (approved):
LGTM :+1: 

> I think the text is wrapping down to a new line when the screen is more narrow...

Yes it does. The button spacing is much better (and, IMO, more importantly consistent). The (long) length of the single line of text is just my opinion. There's nothing wrong with it.. so ignore me.

manstis:
This may help with the failing tests: https://github.com/ansible/ansible-ai-connect-service/pull/1466

PR #1462, opened 2024-12-13
pyproject: accept the regular flavor of psycopg

goneri:
Don't force the use of the `binary` package. It may not be available (e.g Python 3.13 Linux).


manstis (approved):
LGTM 👍

manstis:
"More stable" != "Perfect".. I've kicked the tests off again.

PR #1467, opened 2024-12-18
schema1: set the new plan_ids keys

goneri:
Amplitude cannot do a filter that match a single key in an array, if we want to
correctly identify an event's plan(s) we need to flatten the plan_id keys.


omaciel:
Needs tests :D

manstis (approved):
LGTM :+1:

PR #1469, opened 2024-12-19
Skip problematic chatbot app test suite

mabashian:

## Description
This particular test suite fails CI due to an uncaught error which is particularly tricky to track down.  To unblock unrelated PR's we're going to skip these tests.  This will be re-enabled when https://issues.redhat.com/browse/AAP-37878 is addressed.

## Testing
CI should run successfully

## Production deployment


manstis (approved):
LGTM 👍 

Sad we've not found a solution but this is the most pragmatic approach to unblock things (just before the Christmas holidays).

manstis:
Oh the beauty of 💚

PR #1470, opened 2024-12-19
AAP-37878: move Chatbot test to vitest-browser-react

goneri:
Right now, our tests are failing due to some problem with `happy-dom`, the fake webbrowser
used to run the tests by `react-testing-library`. `jsdom` is an alternative but we cannot
use it without adjusting our tests and facing similar issues or limitations.

By moving to `vitest-browser-react`, we are now able to run the tests against a real browser
like Chrome or Firefox. We still have the option to run the tests "headless" for a CI environment.

`vitest-browser-react` is directly based on `vitest-browser` and allow us to stay in a standard
`vite` ecosystem. It also reuses some elements from `react-testing-library`, which explains why
the PR is not that large.

Overall, this change reduces the differences between our tests and the production environment,
and allow us to limit our exposure to the limitations of these dummy browser implementations.

### DIFF
--- a/ansible_ai_connect_chatbot/src/App.test.tsx
+++ b/ansible_ai_connect_chatbot/src/App.test.tsx
@@ -1,403 +1,409 @@
+// NOTE: We don't need to import React, this is just to avoid the
+// following warning:
+// 'React' must be in scope when using JSX
+// react/react-in-jsx-scope
 import React from "react";
-import { vi } from "vitest";
-import { act, fireEvent, render, screen } from "@testing-library/react";
+
+// NOTE: ignoring because we non-use of "screen" is conistent with the
+// vitest-browser-react documentation
+/* eslint-disable testing-library/prefer-screen-queries */
+
+import { beforeEach, expect, test, vi } from "vitest";
+import { render } from "vitest-browser-react";
 import { MemoryRouter } from "react-router-dom";
 import { App } from "./App";
 import { ColorThemeSwitch } from "./ColorThemeSwitch/ColorThemeSwitch";
-import userEvent from "@testing-library/user-event";
+import { userEvent, page } from "@vitest/browser/context";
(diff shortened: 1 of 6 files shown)


manstis:
Super star 🦸 

I'll look tomorrow.

jameswnl (approved):

justjais (approved):
Goneri++

ansible_ai_connect_chatbot/src/App.test.tsx
```diff
@@ -1,403 +1,409 @@
...
-        ),
-      ).toBeInTheDocument();
-    } finally {
-      vi.useRealTimers();
-    }
-  });
+test("Chat service returns 429 Too Many Requests error", async () => {
+  mockAxios(429, true);
+  await renderApp();
+  const textArea = page.getByLabelText("Send a message...");
```
manstis, line 287: You could call `sendMessage("Hello")` and remove L#287-289.

ansible_ai_connect_chatbot/src/App.test.tsx
```diff
@@ -1,403 +1,409 @@
...
-    const textArea = await screen.findByLabelText("Send a message...");
-    await act(async () => userEvent.type(textArea, "Hello"));
-    const sendButton = await screen.findByLabelText("Send button");
-    await act(async () => fireEvent.click(sendButton));
-    expect(
-      await screen.findByText(
+test("Chat service returns an unexpected error", async () => {
+  mockAxios(-1, true);
+  const view = await renderApp();
+  const textArea = page.getByLabelText("Send a message...");
```
manstis, line 304: You could call `sendMessage("Hello")` and remove L#304-306.

goneri:
Great, thank you for rhe reviews.

TamiTakamiya:
@goneri @manstis Thanks for the change!!

PR #1477, opened 2025-01-03
Fix chatbot UI autoscrolling and insert position for bot message feedback

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-35711>
This also will fix the issue reported by https://github.com/ansible/ansible-lightspeed-va-feedback/issues/15

## Description

On the existing implementaiton, when a new bot message is received, chatbot window scrolls to the bottom of the message. It is changed to scroll to the top of the last message instead.

Also when a ThumbsUp/Down icon is clicked for a bot message, which is NOT the last one, the feedback message is added to the bottom and chatbot window scrolls to the bottom (it is the issue reported by https://github.com/ansible/ansible-lightspeed-va-feedback/issues/15). With this PR, the feedback message will be added right after the bot message, to which the feedback was sent, and chatbot window scrolls to the top of the feedback message.

### Steps to test
1. Pull down the PR
2. Run the chatbot
3. Send more than one queries
4. Click ThubsUp/Down icon for a bot message, which is NOT the last one, and see the window scrolls to the feedback message ("Thank you for your feedback...")

### Scenarios tested

Manually tested as described above. See the video attached below.

https://github.com/user-attachments/assets/e3c62973-7bcf-4179-a681-eda84cfa9220

## Production deployment

### DIFF
--- a/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
+++ b/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
@@ -155,7 +155,7 @@ export const useChatbot = () => {
       newMessage.scrollToHere = true;
       let inserted = false;
       for (const msg of msgs) {
-        delete msg.scrollToHere;
+        msg.scrollToHere = false;
         newMsgs.push(msg);
         if (msg === addAfter) {
           newMsgs.push(newMessage);
--- a/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
+++ b/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
@@ -146,25 +146,45 @@ export const useChatbot = () => {
   >(undefined);
   const [selectedModel, setSelectedModel] = useState("granite3-8b");
 
-  const addMessage = (newMessage: ExtendedMessage) => {
-    setMessages((msgs: ExtendedMessage[]) => [...msgs, newMessage]);
+  const addMessage = (
+    newMessage: ExtendedMessage,
+    addAfter?: ExtendedMessage,
+  ) => {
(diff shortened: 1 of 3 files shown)


ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -146,25 +146,45 @@ export const useChatbot = () => {
...
+  const addMessage = (
+    newMessage: ExtendedMessage,
+    addAfter?: ExtendedMessage,
+  ) => {
+    setMessages((msgs: ExtendedMessage[]) => {
+      const newMsgs: ExtendedMessage[] = [];
+      newMessage.scrollToHere = true;
+      let inserted = false;
+      for (const msg of msgs) {
+        delete msg.scrollToHere;
```
manstis, line 158: It might be _nicer_ to set to `false` rather than delete from the object?

manstis (approved):
Minor comment.

Untried or tested but changes seem reasonable to me.

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
TamiTakamiya: I thought having one `scrollToHere` flag in the whole messages would be simpler, but it is true that changing a boolean value should be less expensive then deleting the attribute... I will update the code as you suggested.

TamiTakamiya: Updated the code in the latest commit. Thanks!

manstis (approved):
LGTM :+1:

PR #1478, opened 2025-01-03
AAP-33915: Use the new exception handler for the Chat view

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-33915

## Description
Internal refactoring of existing code.

## Testing
N/A. Internal refactoring of existing code.

### Steps to test
N/A. Internal refactoring of existing code.

### Scenarios tested
N/A. Internal refactoring of existing code.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/telemetry/schema1.py
+++ b/ansible_ai_connect/ai/api/telemetry/schema1.py
@@ -33,7 +33,7 @@ logger = logging.getLogger(__name__)
 version_info = VersionInfo()
 
 
-def _anonymize_struct(value):
+def anonymize_struct(value):
     return anonymizer.anonymize_struct(value, value_template=Template("{{ _${variable_name}_ }}"))
 
 
@@ -167,7 +167,7 @@ class ChatBotResponseDocsReferences:
 
 
 @define
-class ChatBotBaseEvent:
+class ChatBotBaseEvent(Schema1Event):
     chat_prompt: str = field(validator=validators.instance_of(str), converter=str, default="")
     chat_system_prompt: str = field(
         validator=validators.instance_of(str), converter=str, default=""
@@ -183,21 +183,15 @@ class ChatBotBaseEvent:
         converter=str,
         default=settings.CHATBOT_DEFAULT_PROVIDER,
     )
-    modelName: str = field(
-        validator=validators.instance_of(str), converter=str, default=settings.CHATBOT_DEFAULT_MODEL
-    )
-    rh_user_org_id: int = field(validator=validators.instance_of(int), converter=int, default=-1)
-    timestamp: str = field(
-        default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
-    )
 
     def __attrs_post_init__(self):
-        self.chat_prompt = _anonymize_struct(self.chat_prompt)
-        self.chat_response = _anonymize_struct(self.chat_response)
+        self.chat_prompt = anonymize_struct(self.chat_prompt)
+        self.chat_response = anonymize_struct(self.chat_response)
 
 
 @define
 class ChatBotFeedbackEvent(ChatBotBaseEvent):
+    event_name: str = "chatFeedbackEvent"
     sentiment: int = field(
         validator=[validators.instance_of(int), validators.in_([0, 1])], converter=int, default=0
     )
@@ -205,7 +199,4 @@ class ChatBotFeedbackEvent(ChatBotBaseEvent):
 
 @define
 class ChatBotOperationalEvent(ChatBotBaseEvent):
-    req_duration: float = field(
-        validator=[validators.instance_of(float)], converter=float, default=0
-    )
-    exception: str = field(validator=validators.instance_of(str), converter=str, default="")
+    event_name: str = "chatOperationalEvent"
--- a/ansible_ai_connect/ai/api/tests/test_views.py
+++ b/ansible_ai_connect/ai/api/tests/test_views.py
@@ -4072,7 +4072,6 @@ class TestChatView(WisdomServiceAPITestCaseBase):
     ):
         if user is None:
             user = self.user
-        self.client.force_authenticate(user=user)
         with (
             patch.object(
                 apps.get_app_config("ai"),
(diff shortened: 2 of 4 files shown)


ansible_ai_connect/ai/api/telemetry/schema1.py
```diff
@@ -167,7 +167,7 @@ class ChatBotResponseDocsReferences:
 
 
 @define
-class ChatBotBaseEvent:
+class ChatBotBaseEvent(Schema1Event):
```
manstis, line 170: Are Chat Bot events considered "Schema 1"?

ansible_ai_connect/ai/api/telemetry/schema1.py
```diff
@@ -180,29 +180,20 @@ class ChatBotBaseEvent:
...
     )
-    modelName: str = field(
-        validator=validators.instance_of(str), converter=str, default=settings.CHATBOT_DEFAULT_MODEL
-    )
-    rh_user_org_id: int = field(validator=validators.instance_of(int), converter=int, default=-1)
-    timestamp: str = field(
-        default=Factory(lambda self: timezone.now().isoformat(), takes_self=True)
-    )
 
     def __attrs_post_init__(self):
```
manstis, line 184: This function isn't really used after the refactor as the properties are not set at the time of instantiation.

The `attrs` library calls this after the `__init__` function.

ansible_ai_connect/ai/api/views.py
```diff
@@ -176,9 +176,18 @@
...
 
 class AACSAPIView(APIView):
+    def __init__(self, **kwargs):
+        super().__init__(kwargs)
+        self.event = None
+        self.model_id = None
+        self.exception = None
+
     def initialize_request(self, request, *args, **kwargs):
+        # Call super first to ensure request object is correctly
```
manstis, line 186: There was a problem with the tests and use of `force_authenticate`. Django had not setup the _forced_ user when we were calling `self.event.set_request` and hence the event contained an `AnonymousUser`.

ansible_ai_connect/ai/api/views.py
```diff
@@ -176,9 +176,18 @@
 
 
 class AACSAPIView(APIView):
+    def __init__(self):
```
manstis, line 179: This just removes some warnings about setting properties later, outside the `__init__` function.

ansible_ai_connect/ai/api/telemetry/schema1.py
romartin: actually yes, schema1 only

ansible_ai_connect/ai/api/telemetry/schema1.py
romartin: I am not sure, I think @goneri implemented this, let's see if he can give a pointer here... :)

ansible_ai_connect/ai/api/tests/test_views.py
```diff
@@ -4212,9 +4211,8 @@ def test_operational_telemetry_error(self):
                 123,
             )
             self.assertEqual(
-                segment_events[0]["properties"]["exception"],
-                "An exception <class 'ansible_ai_connect.ai.api.exceptions."
-                "ChatbotInvalidResponseException'> occurred during a chat generation",
+                segment_events[0]["properties"]["problem"],
```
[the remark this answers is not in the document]
romartin: I think so, well at least if he's using this field as for reports in Amplitude. Although this looks not a blocker, we just need to let him know in order to "refactor", if necessary, any query.

Also @manstis this is a good candidate to make the integration tests to fail :) I think we're doing assertions over this field, it worth checking too, IMO :)

romartin (commented):
@manstis code looks good to me, just tried to answer some of your Qs.

ansible_ai_connect/ai/api/telemetry/schema1.py
goneri: - `timestamp` is set line 90
- `rh_user_org_id` is set by `set_user()`
- `modelName` is a bit special because it should be set from `AACSAPIView` ([See](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/views.py#L226)). So it's better to set if from the view as an attribute.

ansible_ai_connect/ai/api/tests/test_views.py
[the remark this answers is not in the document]
goneri: note: I knew the way we use this field is not always consistent depending on the view and I would like to converge to the new approach.

Feel free to object/cry/shout :-).

ansible_ai_connect/ai/api/views.py
```diff
@@ -176,9 +176,18 @@
...
+        super().__init__()
+        self.event = None
+        self.model_id = None
+        self.exception = None
+
     def initialize_request(self, request, *args, **kwargs):
+        # Call super first to ensure request object is correctly
+        # initialised before instantiating event
+        _request = super().initialize_request(request, *args, **kwargs)
```
goneri, line 188: I've a light objection regarding the use of the `_`, if we want to flag our private variables like this and be consistent, we will need to re-factorize the whole code base. I would prefer we keep the current approach (no `_` prefix).

ansible_ai_connect/ai/api/views.py
```diff
@@ -1152,105 +1162,69 @@ def __init__(self):
...
-            )
 
-            start = timeit.default_timer()
+        if not self.chatbot_enabled:
+            raise ChatbotNotEnabledException()
 
-            llm: ModelPipelineChatBot = apps.get_app_config("ai").get_model_pipeline(
-                ModelPipelineChatBot
-            )
+        if not request_serializer.is_valid():
```
goneri, line 1169: Note: this will later be done from within `AACSAPIView` and we will get a generic exception (not view specific).

ansible_ai_connect/ai/api/views.py
manstis: Renamed.

ansible_ai_connect/ai/api/tests/test_views.py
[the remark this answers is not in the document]
manstis: @romartin I could only find [these](https://github.com/ansible/ansible-wisdom-testing/blob/main/tests/api/test_chat.py) tests. This seems to only be a basic _sanity_ check. Do you know of any others?

goneri (approved):

ansible_ai_connect/ai/api/tests/test_views.py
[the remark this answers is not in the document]
romartin: oh @manstis , correct! Sorry for the confusion.

romartin (approved):

PR #1480, opened 2025-01-06
Add system_prompt to chat endpoint

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-38211>

## Description

Add system_prompt to `POST /chat` endpoint.

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit test cases are included.

## Production deployment

- [X] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1482, opened 2025-01-07
Add System Prompt modal to chatbot debug UI

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-38211>

## Description

Add UI for allowing user to specify a custom system prompt in debug mode (STAGE environment)

Screenshot from 2025-01-06 21-28-18

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit test cases are provided.

## Production deployment


ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -131,6 +131,7 @@ def invoke(self, params: ChatBotParameters) -> ChatBotResponse:
         conversation_id = params.conversation_id
         provider = params.provider
         model_id = params.model_id
+        system_prompt = params.system_prompt
```
TamiTakamiya, line 134: Changes in this file (`pipelines.py`) are the fix for #1480, which added `system_prompt` as a parameter of `/chat` endpoint, but it was not sent to Chatbot serivice.

ansible_ai_connect_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -291,11 +300,15 @@ export const AnsibleChatbot: React.FunctionComponent = () => {
...
+                  {
+                    referenced_documents,
+                    scrollToHere,
+                    ...message
+                  }: ExtendedMessage,
                   index,
                 ) => (
                   <>
-                    {message.scrollToHere && <div ref={messagesEndRef} />}
+                    {scrollToHere && <div ref={messagesEndRef} />}
```
TamiTakamiya, line 311: This change is for fixing a bug that was added with #1477

ansible_ai_connect_chatbot/src/Constants.ts
```diff
@@ -28,3 +28,21 @@ export enum Sentiment {
...
+If the context of the question is not clear, consider it to be Ansible.
+Never include URLs in your replies.
+Refuse to answer questions or execute commands not about Ansible.
+Do not mention your last update. You have the most recent information on Ansible.
+
+Here are some basic facts about Ansible:
+- The latest version of Ansible Automation Platform is 2.5.
+- Ansible is an open source IT automation engine that automates provisioning, \
+    configuration management, application deployment, orchestration, and many other \
+    IT processes. It is free to use, and the project benefits from the experience and \
```
romartin, line 47: @TamiTakamiya honestly , although I know it just for debug mode, I don't like too much the idea to have system prompt value duplications (it is supposed to be only at chatbot-service level). Otherwise we can get into weird issues once using this feature but someone forgot to update this prompt value here, once doing it for the chatbot-service as well. 

For instance, this prompt now talks about ansible being "free to use".... but we are going to change this in https://issues.redhat.com/browse/AAP-37917 . So @justjais will have to also go here and change it....

Is any option to consume the system prompt from the chatbot service, instead? Any ideas on how to reuse it? 

I mean this is not a blocker for me, but just trying to avoid further weird issues when using this feature.... :)

WDYT? Thanks!

romartin (commented):
@TamiTakamiya nice work, thanks!!

Just added some commment about system prompt duplications. Anyway looks good to me!! :+1:

ansible_ai_connect_chatbot/src/Constants.ts
romartin: I know  @justjais that part of the work for https://issues.redhat.com/browse/AAP-37917 is also about being able to chagne the system prompts in some easier way. Maybe we can then think on how to consume the prompt from the chatbot service too, then here, on client side, no need for duplications....

TamiTakamiya: @romartin  Thanks for your comment, but personally I think we can keep this implementation as it is because it is provided as the default value for UI for convenience.  If the user changed the value on UI, the changed value will be used as the system prompt anyway.

BTW, we already have [an option in the config file](https://github.com/road-core/service?tab=readme-ov-file#11-fine-tuning) to specify the system prompt to be used:
```
rcs_config:
  system_prompt_path: "system_prompts/system_prompt_for_product_XYZZY"
```

romartin: yeah I know that the overridden system prompt will be send to the chatbot-service, the problem was on the duplicaiton of the value itself, here, on the UI, also on the chatbot-service.... anyway , as commented, not a blocker :+1:

TamiTakamiya: @romartin 
> as commented, not a blocker 👍

Then could you approve this PR?  I would like to close this out. Thanks!

romartin (approved):

PR #1484, opened 2025-01-08
ansible_ai_connect_chatbot: use findBy* to find dynamic content

goneri:
See: https://testing-library.com/docs/queries/about/#findby


manstis (approved):
LGTM 👍

PR #1485, opened 2025-01-08
AAP-30619: migrate PlaybookGeneration to the new Schema1 object

goneri:
Use an object defined in ansible_ai_connect/ai/api/telemetry/schema1.py to prepare the Schema1 payload.

### DIFF
--- a/ansible_ai_connect/ai/api/views.py
+++ b/ansible_ai_connect/ai/api/views.py
@@ -185,7 +185,7 @@ class AACSAPIView(APIView):
         # should exposed as self.validated_data.model, or using a special
         # method.
         # See: https://github.com/ansible/ansible-ai-connect-service/pull/1147/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9R143-R151  # noqa: E501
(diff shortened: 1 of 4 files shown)


ansible_ai_connect/ai/api/views.py
```diff
@@ -820,195 +823,67 @@ class GenerationPlaybook(APIView):
...
+                text=text,
+                custom_prompt=custom_prompt,
+                create_outline=create_outline,
+                outline=outline,
+                generation_id=generation_id,
+                model_id=model_id,
             )
-            raise WcaInstanceDeletedException(cause=e)
+        )
+        self.event.playbook_length = len(playbook)
```
manstis, line 863: Should this be moved lower down and use `self.event.playbook_length = len(anonymized_playbook)`?

The event [used to contain](https://github.com/ansible/ansible-ai-connect-service/pull/1485/files#diff-ecfb6919dfd8379aafba96af7457b253e4dce528897dfe6bfc207ca2b3b2ada9L969) the length of the anonymized playbook.

manstis (commented):
Minor comment; otherwise LGTM :+1:

PR #1486, opened 2025-01-09
AACSAPIView: validate the payload in the class

goneri:
Validate the view parameter from within the `AACSAPIView` class as soon as the `request_serializer_class` attribute is set with a Serializer class.

Also introduce minor cosmetic changes in Serializer class definitions to set some default values.
These changes are optional but the ultimate goal is to ensure all the fields are set in the `validated_data`
structure. This to reduce the use of `.get()`, which can silently hide an actually problem. This is also a way to
ensure the type of the field remains the same.

### DIFF
--- a/ansible_ai_connect/ai/api/views.py
+++ b/ansible_ai_connect/ai/api/views.py
@@ -34,7 +34,6 @@ from rest_framework.views import APIView
 from ansible_ai_connect.ai.api.aws.exceptions import WcaSecretManagerError
 from ansible_ai_connect.ai.api.exceptions import (
     BaseWisdomAPIException,
-    ChatbotInvalidRequestException,
     ChatbotInvalidResponseException,
     ChatbotNotEnabledException,
     FeedbackInternalServerException,
@@ -179,8 +178,9 @@ class AACSAPIView(APIView):
     def __init__(self):
         super().__init__()
         self.event = None
-        self.model_id = None
+        self.req_model_id = None
         self.exception = None
+        self.validated_data = {}
 
     def initialize_request(self, request, *args, **kwargs):
         # Call super first to ensure request object is correctly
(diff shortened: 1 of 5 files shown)


ansible_ai_connect/ai/api/views.py
```diff
@@ -189,15 +189,19 @@ def initialize_request(self, request, *args, **kwargs):
...
+            self.load_parameters(request)
 
         return initialised_request
 
+    def load_parameters(self, request) -> Response:
+        if hasattr(self, "request_serializer_class"):
+            request_serializer = self.request_serializer_class(data=request.data)
+            request_serializer.is_valid(raise_exception=True)
+            self.validated_data = request_serializer.validated_data
+            if req_model_id := self.validated_data.get("mode_id"):
```
manstis, line 202: `model_id`

ansible_ai_connect/ai/api/views.py
```diff
@@ -763,32 +767,20 @@ class Explanation(AACSAPIView):
...
-                model_id=self.model_id,
+                content=self.validated_data["content"],
+                custom_prompt=self.validated_data["customPrompt"],
+                explanation_id=self.validated_data["explanationId"],
+                model_id=self.validated_data["model"],
             )
         )
-        self.event.playbook_length = len(playbook)
-        self.event.explanationId = explanation_id
+        self.event.playbook_length = self.validated_data["content"]
```
manstis, line 782: This doesn't look _right_. `playbook_length = validated_data["content"]`?

ansible_ai_connect/ai/api/views.py
goneri: Since we have different `model_id` ( or `modelName` ) variables, I use `req_model_id` internally to store the value the user requested in the payload.

manstis: Yes, but you've typed `mode_id` not `model_id`.

goneri: oh you mean `model_id`, not `mode_id` :D. Thanks!

ansible_ai_connect/ai/api/views.py
```diff
@@ -832,43 +828,24 @@ class GenerationPlaybook(AACSAPIView):
...
-                model_id=model_id,
+                text=self.validated_data["text"],
+                custom_prompt=self.validated_data["customPrompt"],
+                create_outline=self.validated_data["createOutline"],
+                outline=self.validated_data["outline"],
+                generation_id=self.validated_data["generationId"],
+                model_id=self.req_model_id,
             )
         )
+        self.event.generationId = self.validated_data["generationId"]
```
manstis, line 847: See my earlier comment about initialising the event as soon as possible.

ansible_ai_connect/ai/api/views.py
```diff
@@ -1083,7 +1040,7 @@ def post(self, request) -> Response:
                 request=request,
                 query=req_query,
                 system_prompt=req_system_prompt,
-                model_id=req_model_id,
+                model_id=self.req_model_id,
```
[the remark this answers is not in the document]
goneri: Thanks.

ansible_ai_connect/ai/api/views.py
```diff
@@ -763,32 +771,20 @@ class Explanation(AACSAPIView):
...
-                model_id=self.model_id,
+                content=self.validated_data["content"],
+                custom_prompt=self.validated_data["customPrompt"],
+                explanation_id=self.validated_data["explanationId"],
+                model_id=self.validated_data["model"],
             )
         )
-        self.event.playbook_length = len(playbook)
-        self.event.explanationId = explanation_id
+        self.event.playbook_length = len(self.validated_data["content"])
```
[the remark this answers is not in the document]
goneri: Indeed, good idea.

manstis (approved):
Thanks. LGTM 👍

PR #1487, opened 2025-01-10
AAP-36732: ModelPipelines: Configuration improvements: Remove env vars

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-36732

## Description
This PR removes the discrete (legacy) environment variables previously used to configure LLM services (WCA, Ollama, GRPC etc). All configuration is now controlled by a JSON (or YAML) block, stored in a `ANSIBLE_AI_MODEL_MESH_CONFIG` environment. Various examples are given in this PR.

I have included support for the legacy use of environment variables.

These are used if `ANSIBLE_AI_MODEL_MESH_CONFIG` is not set.

This is predominantly for the use by the Operator; where an older version uses the latest service version.

## Testing
N/A. Internal refactoring of existing code.

### Steps to test
N/A. Internal refactoring of existing code.

### Scenarios tested
N/A. Internal refactoring of existing code.

## Production deployment


README-ANSIBLE_AI_MODEL_MESH_CONFIG.md
```diff
@@ -0,0 +1,40 @@
+# Example `ANSIBLE_AI_MODEL_MESH_CONFIG` configuration
+
+Pay close attention to the formatting of the blocks.
+
+Each ends with `}},` otherwise conversion of the multi-line setting to a `str` can fail.
```
goneri, line 5: YAML would maybe be more elegant here?

manstis: @goneri I've added support for loading _either_ JSON or YAML and updated the README's.

manstis: "Pay close attention to the formatting of the blocks".

This was caused by (me) trying to define JSON as a multi-line String in a `.env` file.

Moving it to a Django/Python settings file and re-configuring my local "launch" configuration fixed things.

goneri:
Work as expected with the following configuration:

```
export ANSIBLE_AI_MODEL_MESH_CONFIG="
---
ModelPipelineCompletions:
  provider: dummy
ModelPipelineContentMatch:
  provider: dummy
ModelPipelinePlaybookGeneration:
  provider: dummy
ModelPipelineRoleGeneration:
  provider: dummy
ModelPipelinePlaybookExplanation:
  provider: dummy
ModelPipelineChatBot:
  provider: dummy
"
```

The part is dislike I found a bit annoying is that if my YAML syntax is incorrect, the exception is silently ignored.

goneri (dismissed):

manstis:
@goneri 

> The part is dislike I found a bit annoying is that if my YAML syntax is incorrect, the exception is silently ignored.

That'll be caused by [this](https://github.com/ansible/ansible-ai-connect-service/pull/1487/files#diff-971ea3504dafb989bdbbf91b141278b61244d622c0bdaee489ffc4e6b4042935R37-R49) where _exceptions_ are effectively ignored. 

Let me tinker a little before this is merged.

manstis:
@goneri 

> Actually my configuration doesn't work because the `config` keys are missing.

Excellent find. Thanks.

manstis:
@goneri PR updated to better handle exceptions (and fixed `KeyError: 'config'`).

.github/workflows/pre-commit.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Set up python3
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.11'
```
manstis, line 33: This aligns the Python version with the other GHA definition.

`3.11` is needed to support `ExceptionGroup` (used in this PR). It is also the version used to [build](https://github.com/ansible/ansible-ai-connect-service/blob/main/wisdom-service.Containerfile#L24) the container.

manstis:
See https://issues.redhat.com/browse/AAP-38734 re: failing "PR Deployment Testing" check.

goneri (approved):

PR #1489, opened 2025-01-13
Chat History Frontend-only Implementation

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-38324>

## Description

Implement chat history UI.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests + manual tests by running a local AI Connect server.

## Production deployment


romartin (approved):
Tested and looking good, thanks!!

Just one comment - the model selector does not change as with the conversation/tab, but I think that's only for debug mode, correct?

TamiTakamiya:
> Just one comment - the model selector does not change as with the conversation/tab, but I think that's only for debug mode, so on prod users will not see it, correct?

Yes, that's correct.

PR #1500, opened 2025-01-20
Skip seated user check for chat segment events

TamiTakamiya:
Jira Issue: n/a

## Description

Skip "seated user check" for Chat-related segment events because Chat is irrelevant to "seated users" or WCA.

I added two new entries to `allow_list`:
``` 
    "chatFeedbackEvent": {
        "*": None,
    },
    "chatOperationalEvent": {
        "*": None,
    },
```
The line` "*": None,` implies all properties are allows.

The problem was found on the `chatOperationalEvent`, but I have added a fix for the `chatFeedbackEvent` as well.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests + manual tests by running local AI Connect server.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/utils/segment.py
+++ b/ansible_ai_connect/ai/api/utils/segment.py
@@ -181,6 +181,10 @@ def redact_seated_users_data(event: Dict[str, Any], allow_list: Dict[str, Any]) 
     Returns:
     - dict: A new dictionary containing only the allowed nested keys from the source dictionary.
     """
+    # If an asterisk is found in the allow list, just return the given event dict to allow all keys.
+    if "*" in allow_list:
+        return event
+
     redacted_event = {}
     for key, sub_whitelist in (
         allow_list.items() if isinstance(allow_list, dict) else allow_list[0].items()
--- a/ansible_ai_connect/ai/api/tests/test_views.py
+++ b/ansible_ai_connect/ai/api/tests/test_views.py
@@ -1968,6 +1968,8 @@ class TestFeedbackView(WisdomServiceAPITestCaseBase):
                 "sentiment": "1",
             }
         }
+        self.user.rh_user_has_seat = True
+        self.user.organization = Organization.objects.get_or_create(id=1)[0]
         self.client.force_authenticate(user=self.user)
         with self.assertLogs(logger="root", level="DEBUG") as log:
             r = self.client.post(reverse("feedback"), payload, format="json")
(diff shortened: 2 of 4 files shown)


ansible_ai_connect/ai/api/utils/segment.py
```diff
@@ -181,6 +181,10 @@ def redact_seated_users_data(event: Dict[str, Any], allow_list: Dict[str, Any])
     Returns:
     - dict: A new dictionary containing only the allowed nested keys from the source dictionary.
     """
+    # If an asterisk is found in the allow list, just return the given event dict to allow all keys.
+    if "*" in allow_list:
```
manstis, line 185: It'd be nice to have a test for these with your new check?

I _know_ SonarCloud says you have 100%... but is that a false positive? I don't _see_ any.

romartin (dismissed):
Approving although agree with @manstis , tests for new wildcard would be great, that's an important feature... Anyway I have tested same fix locally and that's the right fix. Thanks @TamiTakamiya !

ansible_ai_connect/ai/api/utils/segment.py
TamiTakamiya: I have added two lines to test_operational_telemetry()`
```
        self.user.rh_user_has_seat = True
        self.user.organization = Organization.objects.get_or_create(id=1)[0]
```
which forces to run the test case with a "seated" user.  Without this allow_list change, the test case failed. Isn't this  change enough for the test coverage?

TamiTakamiya: > Isn't this change enough for the test coverage?

I found there were some test cases on redact operations in `test_segment.py`.  I'll add two test cases to it.

manstis: Yes, that's fine @TamiTakamiya 

It wasn't clear from your PR.

manstis (approved):
Thanks for the explanation. 

LGTM 👍

TamiTakamiya:
@manstis @romartin I have added two simple test cases to `test_segment.py`.  Pls re-review this. thanks!

manstis (approved):
LGTM :+1:

ansible_ai_connect/ai/api/utils/segment.py
hasys: Hey @TamiTakamiya, @manstis, sorry for late response, just my 2c the idea behind allow list agains block list in this case is making sure you are logging only something you really need to log, and avoid situation we are storing things we are not supposed to store.

Whit the asterisk it is easier to make a mistake and store unintended data.

manstis: @hasys You're never late to the party and, as author of the original white listing, your opinion is important.. When you wrote the enhancement you no doubt had a clear vision in mind and it should not be _bastardized_ to fit other use-cases (as this is how we end up with spaghetti code!)

What would you have proposed?

IIUC the issue was that the Chat Bot events were not recognised as "valid" events and hence filtered out.

@TamiTakamiya's change means they are no longer blocked by the filter and obviates the need to list all fields.

PR #1502, opened 2025-01-22
feedback: allow events coming from roleGen

goneri:
Accept the action and feeback events coming from the extension for the role generation.

### DIFF
--- a/ansible_ai_connect/ai/api/serializers.py
+++ b/ansible_ai_connect/ai/api/serializers.py
@@ -262,6 +262,7 @@ class IssueFeedback(serializers.Serializer):
     )
 
 
+# TODO: we don't use this class, we can remove it
 class PlaybookGenerationFeedback(serializers.Serializer):
     USER_ACTION_CHOICES = (("0", "ACCEPTED"), ("1", "REJECTED"), ("2", "IGNORED"))
 
@@ -274,10 +275,21 @@ class PlaybookGenerationFeedback(serializers.Serializer):
     )
 
 
-class PlaybookGenerationAction(serializers.Serializer):
-    ACTIONS = (("0", "OPEN"), ("1", "CLOSE_CANCEL"), ("2", "TRANSITION"), ("3", "CLOSE_ACCEPT"))
+class GenerationActionEnum(serializers.ChoiceField):
+    def __init__(self):
+        super().__init__(
+            choices=(
+                ("0", "OPEN"),
+                ("1", "CLOSE_CANCEL"),
+                ("2", "TRANSITION"),
+                ("3", "CLOSE_ACCEPT"),
+            ),
+            required=True,
+        )
 
-    action = serializers.ChoiceField(choices=ACTIONS, required=True)
+
+class PlaybookGenerationAction(serializers.Serializer):
+    action = GenerationActionEnum()
     wizardId = serializers.UUIDField(
         format="hex_verbose",
         required=True,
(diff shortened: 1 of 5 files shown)


ansible_ai_connect/ai/api/serializers.py
```diff
@@ -394,9 +386,9 @@ class FeedbackRequestSerializer(Metadata):
     issueFeedback = IssueFeedback(required=False)
     metadata = Metadata(required=False)
     model = serializers.CharField(required=False)
-    playbookExplanationFeedback = PlaybookExplanationFeedback(required=False)
-    playbookGenerationFeedback = PlaybookGenerationFeedback(required=False)
+    playbookExplanationFeedback = PlaybookExplanationAction(required=False)
```
manstis, line 389: Any reason **not** to rename the variable, `playbookExplanationFeedback` too?

> We use the same naming convention in `vscode-ansible`.

Do you?

I see it's still [referred](https://github.com/ansible/vscode-ansible/blob/main/src/extension.ts#L624) to as `...Feedback`. I guess keeping this as `...Feedback` preserves backwards compatibility with older VSCode Extension versions.

tools/openapi-schema/ansible-ai-connect-service.yaml
```diff
@@ -705,6 +705,18 @@ components:
         * `0` - ACCEPTED
         * `1` - REJECTED
         * `2` - IGNORED
+    Action693Enum:
```
manstis, line 708: It's a shame we have to have a _system generated_ name for the enum.

Looks like we used to have `PlaybookGenerationActionActionEnum` (which was _slightly_ better).

ansible_ai_connect/ai/api/views.py
```diff
@@ -450,6 +455,31 @@ def write_to_segment(
...
+            wizard_id = str(role_generation_action_data.get("wizardId", ""))
+            event = {
+                "action": action,
+                "wizardId": wizard_id,
+                "fromPage": from_page,
+                "toPage": to_page,
+                "modelName": model_name,
+            }
+            send_segment_event(event, "roleGenerationAction", user)
+            if False and from_page > 1 and action in [1, 3]:
```
manstis, line 471: Is this deliberate? You're not sending Segment Events for page transitions.

manstis (commented):
Other than a couple of niggles.. but I think I answered one myself; LGTM :+1: 

I'd more interested in your `if False...` when sending page transition events.

ansible_ai_connect/ai/api/views.py
goneri: Yes, Aparna needs to approve the new event first.

mabashian (approved):

ansible_ai_connect/ai/api/tests/test_feedback_view.py
```diff
@@ -380,26 +379,36 @@ def test_feedback_explanation(self):
             properties = segment_events[0]["properties"]
             self.assertEqual(properties["action"], "1")
 
-    @skip("Schema2 event is not enabled yet")
     def test_feedback_generation(self):
+        action = {
```
manstis, line 383: I'm struggling to understand why the properties in this structure are duplicated.

If it's important that it's duplicated perhaps add a comment so somebody doesn't think to "clean it up" in the future.

Please explain yourself :laughing:

manstis (commented):
On the whole LGTM :+1: 

I just don't understand however why the test data is duplicated.

Am I missing something obvious?!?

ansible_ai_connect/ai/api/tests/test_feedback_view.py
goneri: Oops, it's me. Let's fix that.

manstis (approved):
LGTM :+1: 

LGTM :+1: 

:rofl:

PR #1503, opened 2025-01-22
Move chat view tests to their own file

mabashian:

## Description
Moves chat view tests out to their own file reducing the size of test_views.py

## Testing


manstis (approved):
LGTM 👍 

Thank-you.

PR #1504, opened 2025-01-22
Move completion view tests to their own file

mabashian:

## Description
Moves completion view tests out to their own file reducing the size of test_views.py

## Testing


manstis (approved):
LGTM 👍 (I didn't check in detail, I assume it's cut/paste).

Thank-you.

PR #1505, opened 2025-01-23
Move completion wca view tests to their own file

mabashian:

## Description
Moves completion wca view tests out to their own file reducing the size of test_views.py

Note that I also removed some leftover unused classes in test_views.py that are unrelated to the completion wca view tests

## Testing


manstis (dismissed):
LGTM :+1:

goneri (approved):
:partying_face:

PR #1509, opened 2025-01-23
config: failback on "nop" if provider is not available

goneri:
Failback on the `nop` provider if the requested provider is not available. This way,
the service can start despite the misconfiguration.

Another approch is to "fail fast" and raise a critical exception with a clear message
to let the user fix the configuration.

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/config_providers.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/config_providers.py
@@ -16,6 +16,7 @@ from typing import Dict
 from ansible_ai_connect.ai.api.model_pipelines.config_pipelines import (
     PipelineConfiguration,
 )
+from ansible_ai_connect.ai.api.model_pipelines.registry import get_registry_entry
 
 
 class Configuration(Dict):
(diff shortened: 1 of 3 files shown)


goneri:
Draft because there is no test yet and overall I want to discuss this with @manstis first.

manstis:
Hmm. I'll look properly tomorrow but the fallback to `NOP` should have already been working. How could I replicate the problem? Simply exclude configuration for a pipeline?

manstis:
If I remember correctly I (should be) creating the `NOP`'s early when the configuration is first loaded. `REGISTRY` should then be fully populated so the later lookups can't fail.... or so I believe. I'll check tomorrow.

manstis:
@goneri if you can find time to mention how to reproduce the issue I'd like to look before you arrive tomorrow. Thanks.

goneri:
> @goneri if you can find time to mention how to reproduce the issue I'd like to look before you arrive tomorrow. Thanks.

It happens if I see the provider of ChatBot to `dummy`.

manstis:
@goneri OK.. so I looked and looked.. but could not find anything that was, or could have been, doing what I thought.

Your PR therefore seems fine to me - thanks for identifying the problem and providing a fix.

I did create a [similar fix](https://github.com/ansible/ansible-ai-connect-service/tree/fix_missing_pipeline_implementations) that ensures people won't need to worry about the _safety_ of `REGISTRY`.

I'll leave it to you to decide which you prefer (but if you could fix [this](https://github.com/ansible/ansible-ai-connect-service/commit/b6e1179a52e23fecd5f86fd6dc9d4340d6fb5613#diff-4129ccbef47b027f5061a080ab4f8b40c484b87c6a4374b4f5678d1315049a0fR43) in your PR it'd be appreciated).

goneri:
Thank you @manstis

PR #1513, opened 2025-01-29
Lint the results from role generation

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-37200

## Description
This PR does two things:

1. Introduces a linting post-processing step to the role gen file content.
2. Accounts for warnings to be returned by wca api which was included in their demo last week.  This should line up with warnings that come back from playbook generation

## Testing
Unit test coverage added for new scenarios

### Scenarios tested
None beyond the automated tests

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
@@ -242,6 +242,11 @@ class WCABasePipeline(
         wca_codegen_playbook_retry_counter.inc()
 
     @staticmethod
+    def on_backoff_codegen_role(details):
+        WCABasePipeline.log_backoff_exception(details)
+        wca_codegen_playbook_retry_counter.inc()
+
+    @staticmethod
     def on_backoff_explain_playbook(details):
         WCABasePipeline.log_backoff_exception(details)
         wca_explain_playbook_retry_counter.inc()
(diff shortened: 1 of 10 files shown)


ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
```diff
@@ -507,7 +512,70 @@ def __init__(self, config: WCA_PIPELINE_CONFIGURATION):
         super().__init__(config=config)
 
     def invoke(self, params: RoleGenerationParameters) -> RoleGenerationResponse:
-        raise NotImplementedError
+        request = params.request
```
goneri, line 515: I believe this should only be implemented in `pipelines_saas.py` since we don't want to provide this feature yet with the onPrem offert. @manstis WDYT?

manstis: @goneri @mabashian Yes, if this is only available for SaaS then that would be a better place for it. However please put a comment saying it should be moved back to the base WCA class when it becomes available for on-prem 😁 (otherwise we'll either have to remember or risk the duplication).

manstis: (I was going to review this PR tomorrow)

mabashian: Thanks @manstis @goneri - I've moved this logic out to the saas pipeline

manstis (approved):
LGTM :+1:

goneri (approved):

PR #1514, opened 2025-01-30
Fix missing pipeline implementations.

manstis:

## Description
It was possible for a User to configure a pipeline to use a provider that did not support a given feature.

In such scenarios `get_model_pipeline(<feature>)` returned `None` leading to RTEs.

This PR ensures a `NOP` implementation is returned if an explicit implementation does not exist.

## Testing
See "Steps to test"

### Steps to test
1. Configure `ANSIBLE_AI_MODEL_MESH_CONFIG` to include:
```
ModelPipelineChatBot:
  provider: dummy
```
2. Start the service
3. Try to load the Home Page.

### Scenarios tested
See "Steps to test"

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/__init__.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/__init__.py
@@ -17,3 +17,6 @@ import ansible_ai_connect.ai.api.model_pipelines.wca.configuration_saas  # noqa
 import ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_dummy  # noqa
 import ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_onprem  # noqa
 import ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_saas  # noqa
+from ansible_ai_connect.ai.api.model_pipelines.registry import set_defaults
+
+set_defaults()
--- a/ansible_ai_connect/ai/api/model_pipelines/README.md
+++ b/ansible_ai_connect/ai/api/model_pipelines/README.md
@@ -1,0 +1,110 @@
+# Model Pipelines
+
+Ansible AI Connect is becoming feature rich.
+
+It supports API for the following features:
+- Code completions
+- Content match
+- Playbook Generation
+- Role Generation
+- Playbook Explanation
+- Chat Bot
+
+"Model Pipelines" provides a mechanism to support different _pipelines_ and configuration for each of these features for different providers. Different providers require different configuration information.
+
+## Pipelines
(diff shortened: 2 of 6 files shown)


ansible_ai_connect/ai/api/model_pipelines/tests/test_factory.py
```diff
@@ -91,6 +99,20 @@ def test_default_fallback_playbook_explanation(self):
                 log,
             )
 
+    # 'dummy' does not have a ChatBot pipeline
+    @override_settings(ANSIBLE_AI_MODEL_MESH_CONFIG=json.dumps(CHATBOT))
```
goneri, line 103: We can do without the `json.dumps` call here.

ansible_ai_connect/ai/api/model_pipelines/__init__.py
```diff
@@ -17,3 +20,26 @@
...
+            logger.warning(
+                f"'{k.alias()}' is not available for provider '{pipeline_provider}',"
+                " failing back to 'nop'"
+            )
+            return defaults[k]
+        return v
+
+    return {k: v_or_default(k, v) for k, v in REGISTRY[pipeline_provider].items()}
+
+
```
goneri, line 43: Since this runs during the module loading, it will be a bit tricky to properly unit-test it. I feel like it would be better to isolate this in a dedicated function.

ansible_ai_connect/ai/api/model_pipelines/tests/test_factory.py
manstis: `ANSIBLE_AI_MODEL_MESH_CONFIG` must be a `string` containing either JSON or YAML.

Using a `dict` is not supported (remember, the value is loaded from an environment variable).

ansible_ai_connect/ai/api/model_pipelines/__init__.py
manstis: Looks like you're right (I've added a test but it's not being detected as covering this code).

I'll move it out.

ansible_ai_connect/ai/api/model_pipelines/tests/test_factory.py
goneri: I mean, `CHATBOT` can just be a plain YAML string :-)

goneri: I mean, `CHATBOT` can just be a plain YAML string :-)

goneri (approved):

PR #1515, opened 2025-01-30
podman-compose cannot recognize relative paths correctly

TamiTakamiya:
Jira Issue: n/a

## Description

podman-compose 1.20 has an issue to recognize relative paths given with `-f` parameter. ([reference](https://github.com/containers/podman-compose/issues/1059))
Since it still can process absolute paths, we'll update our `Makefile` to use absolute paths by adding `${PWD}` in front of relative paths

### Steps to test
1. Pull down the PR
2. Run some make targets that uses `podman-compose`, like `make start-backends`.

### Scenarios tested

Manually ran `make` targets.

## Production deployment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,27 +67,27 @@ run-server-containerized:
 
 .PHONY: docker-compose
 docker-compose:
-	${COMPOSE_RUNTIME} -f tools/docker-compose/compose.yaml up --remove-orphans
+	${COMPOSE_RUNTIME} -f ${PWD}/tools/docker-compose/compose.yaml up --remove-orphans
 
 .PHONY: start-db
 # Run db in container for running Django application from source
 start-db:
-	${COMPOSE_RUNTIME} -f tools/docker-compose/compose-db.yaml up --remove-orphans -d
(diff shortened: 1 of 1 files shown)


manstis (approved):
LGTM :+1:

Makefile
```diff
@@ -67,27 +67,27 @@ run-server-containerized:
...
 # Run db in container for running Django application from source
 start-db:
-	${COMPOSE_RUNTIME} -f tools/docker-compose/compose-db.yaml up --remove-orphans -d
+	${COMPOSE_RUNTIME} -f ${PWD}/tools/docker-compose/compose-db.yaml up --remove-orphans -d
 
 .PHONY: stop-db
 # stop db container
 stop-db:
-	${COMPOSE_RUNTIME} -f tools/docker-compose/compose-db.yaml up down
+	${COMPOSE_RUNTIME} -f ${PWD}/tools/docker-compose/compose-db.yaml down
```
TamiTakamiya, line 80: This was not related to the `podman-compose` issue.  But it did not work unless I removed `up` :-)

omaciel (approved):
/lgtm

PR #1520, opened 2025-02-04
Schema1: enable playbook and role geneation events

goneri:


manstis (commented):
Does this need unit testing?

manstis (approved):
> Does this need unit testing?

Clearly not :partying_face:

PR #1523, opened 2025-02-05
PlaybookExplanationFeedback: do not accept action == 3

goneri:
Partial revert of b2e04ee to be sure we don't accept an `PlaybookExplanationFeedback` event when action == 3 and fix the name
of the actions in the OpenAPI schema.


manstis (approved):
LGTM :+1:

goneri:
Sorry, I missed that part. I just reopened @mabashian's PR.

PR #1526, opened 2025-02-10
Add scaffolding for role explanation endpoint

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-39792

## Description
An attempt to scaffold in the role explanation endpoint

### Steps to test
1. Pull down the PR
2. Check out the swagger api schema and compare against proposal in Jira
3. Ensure unit tests run successfully

## Production deployment

### DIFF
--- a/README-ANSIBLE_AI_MODEL_MESH_CONFIG.md
+++ b/README-ANSIBLE_AI_MODEL_MESH_CONFIG.md
@@ -31,6 +31,11 @@ ANSIBLE_AI_MODEL_MESH_CONFIG="{
         "config": {
             "inference_url": "http://host.containers.internal:11434",
             "model_id": "mistral:instruct"}},
+    "ModelPipelineRoleExplanation": {
+        "provider": "ollama",
+        "config": {
+            "inference_url": "http://host.containers.internal:11434",
+            "model_id": "mistral:instruct"}},
     "ModelPipelineChatBot": {
         "provider": "http",
         "config": {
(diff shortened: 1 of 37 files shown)


ansible_ai_connect/ai/api/model_pipelines/langchain/pipelines.py
```diff
@@ -443,3 +446,18 @@ def self_test(self) -> Optional[HealthCheckSummary]:
 
     def get_chat_model(self, model_id):
         raise NotImplementedError
+
+
+class LangchainRoleExplanationPipeline(
```
manstis, line 451: Do you need this one?

It'd fall back to the `NOP` implementation if not explicitly implemented.

The `NOP` [implementation](https://github.com/ansible/ansible-ai-connect-service/pull/1526/files#diff-6550699009921b7b336d5b303016e012e80d6913762e34a227718bcb64bea840R129) raises `NotImplementedError`.

manstis (approved):
LGTM :+1: (A small comment, but harmless nonetheless).

You're the first developer to make new changes using:

- Versioned API
- `AACSAPIView` base class

You're also one of the first to make new changes using:

- `ai/api/model_pipelines`

What was your experience?

ansible_ai_connect/ai/api/model_pipelines/langchain/pipelines.py
mabashian: 👍 I dropped this class in favor of using NopRoleExplanationPipeline where needed

mabashian:
@manstis most (really almost all) of this work was simply following the patterns already established but I didn't find it too cumbersome to work with this architecture.  If I'm honest I'm still a bit confused on whether this endpoint should be added to v0.  It's currently included in the v0 urls but should it only be available in v1?

goneri (approved):

PR #1534, opened 2025-02-13
dummy playbookgen: don't return Markdown tags

goneri:
Don't wrap the YAML answer in between Markdown tags. Those breaks the
`YAML` structure and should not be part of the answer.


manstis (approved):
LGTM 👍

PR #1535, opened 2025-02-13
Chatbot UI update

TamiTakamiya:
Jira Issue: n/a

## Description

Minor Chatbot UI update on following three points:

1. Update PF6 chatbot library to 2.2.0-prerelease.21, which includes the fix for https://github.com/patternfly/chatbot/issues/416
   * Also added @patternfly/react-table and updated eslint-plugin-testing-library and typescript-eslint 
2. Added a small change on the timing to stop showing the waiting icon. Before the change, the waiting icon was deleted when the very first response was received from the chatbot service.  It was changed to delete when the first visible data is received. There is a gap between the first response and the first visible data. 
    * Aslo added the Accept header for streaming chat.  It is needed when the chatbot UI is connected to the new streaming chat endpoint, which is being developed on AI Connect service.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests + manual test by running the local vite server.

## Production deployment

### DIFF
--- a/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
+++ b/ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
@@ -369,13 +369,12 @@ export const useChatbot = () => {
             method: "POST",
             headers: {
               "Content-Type": "application/json",
+              Accept: "application/json,text/event-stream",
               "X-CSRFToken": csrfToken!,
             },
             body: JSON.stringify(chatRequest),
             async onopen(resp: any) {
-              if (resp.ok && resp.status === 200) {
-                setIsLoading(false);
-              } else if (
+              if (
                 resp.status >= 400 &&
                 resp.status < 500 &&
                 resp.status !== 429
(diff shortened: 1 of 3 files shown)


ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -369,13 +369,12 @@ export const useChatbot = () => {
             method: "POST",
             headers: {
               "Content-Type": "application/json",
+              Accept: "application/json,text/event-stream",
```
TamiTakamiya, line 372: This was not needed when the UI was connected directly to the Chatbot service, but when it is connected to the streaming chat endpoint on the AI Connect service (<-- still under development), UI did not connect without this.

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -369,13 +369,12 @@ export const useChatbot = () => {
...
               "X-CSRFToken": csrfToken!,
             },
             body: JSON.stringify(chatRequest),
             async onopen(resp: any) {
-              if (resp.ok && resp.status === 200) {
-                setIsLoading(false);
```
TamiTakamiya, line 377: This line was used to remove the loading icon when the stream is opened. It is moved to line 397 when the first character is received.

romartin (approved):

PR #1542, opened 2025-02-28
make pyright happy again

ldjebran:
## Description
make pyright happy again

## Production deployment

### DIFF
--- a/ansible_ai_connect/healthcheck/backends.py
+++ b/ansible_ai_connect/healthcheck/backends.py
@@ -32,7 +32,7 @@ class HealthCheckSummaryException:
     exception: HealthCheckException
     cause: Exception
 
-    def __init__(self, exception: HealthCheckException, cause: Exception = None) -> None:
+    def __init__(self, exception: HealthCheckException, cause: Exception | None = None) -> None:
         self.exception = exception
         self.cause = cause
 


goneri (approved):

omaciel (approved):
/lgtm

ansible_ai_connect/healthcheck/backends.py
```diff
@@ -32,7 +32,7 @@ class HealthCheckSummaryException:
     exception: HealthCheckException
     cause: Exception
 
-    def __init__(self, exception: HealthCheckException, cause: Exception = None) -> None:
+    def __init__(self, exception: HealthCheckException, cause: Exception | None = None) -> None:
```
manstis, line 35: I'm not sure this is the correct fix.

When I wrote this I expect an exception.

IIRC it is used elsewhere.

ldjebran: but we are assigning to None , so the type should be None or Exception

ldjebran: or we have not to assign None, and always provide an exception

ldjebran: anyway the fix for pyright is OK, if we need to fix to always have an exception we have to review all other code that is initiating this class, self_test is not providing a second argument for sure.

PR #1543, opened 2025-02-28
reduce the severity of the "is not available for provider blabla" messages

goneri:
These messages are flooding our logs in production and pretty normal
until we finally migrate to `ANSIBLE_AI_MODEL_MESH_CONFIG`.


manstis (approved):
LGTM 👍 I was thinking much the same.

PR #1544, opened 2025-03-03
Disable chat streaming temporarily

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-39044>

## Description

Chat streaming was enabled with #1527, but it is not working.  AI Connect Service logs following error:
```
django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: 'daphne'.
```
and `400` error is sent to chatbot UI.  Until a solution is found, we need to let chatbot UI to use non-streaming `/chat` endpoint.

### Steps to test
n/a

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/views.py
+++ b/ansible_ai_connect/main/views.py
@@ -160,7 +160,8 @@ class ChatbotView(ProtectedTemplateView):
         if user and user.is_authenticated:
             context["user_name"] = user.username
         context["debug"] = "true" if settings.CHATBOT_DEBUG_UI else "false"
-        context["stream"] = "true" if self.streaming_chatbot_enabled else "false"
+        # context["stream"] = "true" if self.streaming_chatbot_enabled else "false"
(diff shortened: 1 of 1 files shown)


jameswnl (approved):
Lgtm

but can we use ops repo / env to toggle this?

TamiTakamiya:
@jameswnl Streaming can be enabled/disabled with `ANSIBLE_AI_MODEL_MESH_CONFIG` envvar, which will replace many envvars used for defining pipelines. As far as I looked at the ops repo,  `ANSIBLE_AI_MODEL_MESH_CONFIG` was not used yet to define pipelines.

TamiTakamiya:
@jameswnl I think one possible sollution for this. I will create another PR.

manstis:
@goneri mentioned that he might be (starting) on changing Lightspeed's AWS Secret Manager to set `ANSIBLE_AI_MODEL_MESH_CONFIG` instead of the discrete environment variables.

PR #1549, opened 2025-03-04
AAP-40235: Improve error notification of features that are not available

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-40235

## Description
Change use of `NotImplementedError` to `FeatureNotAvailable` when a pipeline does not implement a feature.

## Testing
- Run `ansible-ai-connect-service` locally with `ANSIBLE_AI_MODEL_MESH_CONFIG="{}"`.
- Run the Ansible VSCode extension configured to use the local `ansible-ai-connect-service`. 
- Attempts to use "Completions", "Content Match", "Playbook Generation" etc should result in a "The requested action is not available in your environment" message.

### Steps to test
As above.

### Scenarios tested
As above.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_onprem.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_onprem.py
@@ -224,9 +224,6 @@ class WCAOnPremRoleExplanationPipeline(
     def __init__(self, config: WCAOnPremConfiguration):
         super().__init__(config=config)
 
-    def invoke(self, params: RoleExplanationParameters) -> RoleExplanationResponse:
-        raise NotImplementedError
-
     def self_test(self) -> Optional[HealthCheckSummary]:
         raise NotImplementedError
 
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_saas.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_saas.py
@@ -427,9 +427,6 @@ class WCASaaSRoleExplanationPipeline(
     def __init__(self, config: WCASaaSConfiguration):
         super().__init__(config=config)
 
-    def invoke(self, params: RoleExplanationParameters) -> RoleExplanationResponse:
-        raise NotImplementedError
-
     def self_test(self) -> Optional[HealthCheckSummary]:
         raise NotImplementedError
 
--- a/ansible_ai_connect/ai/api/pipelines/completion_stages/inference.py
+++ b/ansible_ai_connect/ai/api/pipelines/completion_stages/inference.py
@@ -25,6 +25,7 @@ from prometheus_client import Histogram
 from ansible_ai_connect.ai.api.data.data_model import ModelMeshPayload
 from ansible_ai_connect.ai.api.exceptions import (
     BaseWisdomAPIException,
+    FeatureNotAvailable,
     ModelTimeoutException,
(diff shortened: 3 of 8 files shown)


ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_onprem.py
```diff
@@ -224,9 +224,6 @@ class WCAOnPremRoleExplanationPipeline(
     def __init__(self, config: WCAOnPremConfiguration):
         super().__init__(config=config)
 
-    def invoke(self, params: RoleExplanationParameters) -> RoleExplanationResponse:
```
manstis, line 227: This overrode the super class's implementation and was not needed.

ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_saas.py
```diff
@@ -427,9 +427,6 @@ class WCASaaSRoleExplanationPipeline(
     def __init__(self, config: WCASaaSConfiguration):
         super().__init__(config=config)
 
-    def invoke(self, params: RoleExplanationParameters) -> RoleExplanationResponse:
```
manstis, line 430: This overrode the super class's implementation and was not needed.

ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_onprem.py
mabashian: 👍

ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_saas.py
mabashian: 👍

goneri (approved):

mabashian (approved):
lgtm, thank you!

ansible_ai_connect/ai/api/pipelines/completion_stages/inference.py
```diff
@@ -221,6 +222,13 @@ def process(self, context: CompletionContext) -> None:
             detail = exception.json_response["detail"]
             raise WcaValidationFailureException(cause=e, detail={"reason": detail})
 
+        except FeatureNotAvailable as e:
+            exception = e
+            logger.exception(
+                f"Completions feature is not available for suggestion {payload.suggestionId}."
+            )
```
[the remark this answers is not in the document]
manstis: Funny enough.. I did that originally; however our `middleware.py` was then (for some reason) failing to extract a `model_id` from the underlying exception.. Effectively wrapping it and re-raising solved the problem (I suspect the issue was much more nuanced. _Most_ of our existing exception handling catches one type of exception and raises a another (for HTTP transport). `FeatureNotAvailable` is already the _outer_ type of exception (i.e. specific to HTTP) and is used from the existing internals of some (disabled) pipelines. I followed the same approach in this PR. 

In an ideal world we could/should raise an _internal_ `FeatureNotImplemented` exception; catch it in our "[exception boundaries](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/views.py#L218-L243)" and raise an HTTP specific flavour.. Hmm.. the more I've typed this the more I think this would be the better/consistent approach... Hmm.. WDYT? Before I merge... fix here, fix in a new "tech debt" JIRA or not fix.

PR #1550, opened 2025-03-06
update LLM model/provider defaults defined in chatbot UI

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-40371>

## Description

Change the default LLM defined in chatbot UI to granite 3.1. It is used only in the debug mode, which is enabled in STAGE.  In non-debug mode, which is used in PROD, there would be no changes in the behaviors.

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit tests.

## Production deployment


manstis (approved):
LGTM 👍

PR #1554, opened 2025-03-07
WCABaseConfiguration: honor the inference_url key

goneri:
Properly expose the `inference_url` value that was used during the
instance creation.


manstis (approved):
LGTM 👍 

Great catch!

PR #1558, opened 2025-03-10
pipeline/http: self_test() with a 5s timeout

goneri:
Ensure the `self_test()` won't take too much time and penalize the other
checks. `requests` doesn't have any default timeout and relies on the
settings of the system.

See: https://github.com/psf/requests/issues/3070

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
@@ -140,7 +140,7 @@ class HttpChatBotMetaData(HttpMetaData):
         )
         try:
             headers = {"Content-Type": "application/json"}
-            r = requests.get(self.config.inference_url + "/readiness", headers=headers)
+            r = requests.get(self.config.inference_url + "/readiness", headers=headers, timeout=5)
             r.raise_for_status()
 
             data = r.json()


ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -140,7 +140,7 @@ def self_test(self) -> Optional[HealthCheckSummary]:
         )
         try:
             headers = {"Content-Type": "application/json"}
-            r = requests.get(self.config.inference_url + "/readiness", headers=headers)
+            r = requests.get(self.config.inference_url + "/readiness", headers=headers, timeout=5)
```
manstis, line 143: The pipeline configuration has a timeout property. Did you want to consider using that? Or did you prefer a dedicated timeout for self test?

manstis (approved):
LGTM 👍 

Comment for consideration.

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
goneri: The two timeouts target two different things. Here we are just ensure the service is alive, and a 1s timeout should just fine.

PR #1564, opened 2025-03-12
Update files in /aap_chatbot

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-41008>

## Description

Update the files under `/aap_chatbot` to be consistent with the files under `/ansible_ai_connect_chatbot` with keeping the changes made for AAP-UI integration.

This PR also includes a new constant `AAP_UI`, which is set to `true`.  Where it is possible (not mandatory), please use it to implement AAP UI-specific changes like
```
AAP_UI ? ChatbotDisplayMode.fullscreen : ChatbotDisplayMode.default,
```

### Steps to test
1. Pull down the PR
2. Run unit tests

## Production deployment


romartin (approved):
Thanks @TamiTakamiya ! I undertand most of this work is about porting changes from one directory to another, so it looks good, as changes have been already approved on the original directory.
 Also the AAP_UI flag looks good to me. Thanks!

PR #1566, opened 2025-03-14
evaluate service_type value

ldjebran:
Jira Issue: <https://issues.redhat.com/browse/AAP-NNNN>

## Description
evaluate the service_type is the resource_server settings is defined the service_type is "lightspeed" otherwise it's "aap"

this will allow to not fail in environment that are not using the integration and using old ansible-django-base

## Production deployment


manstis (approved):
LGTM 👍

PR #1569, opened 2025-03-17
AAP-42164: roleGen: properly pass the "name" parameter

goneri:
The `name` parameter was ignored. We need it if the user wants to overwrite
the default name returned by the LLM.

See: https://github.com/ansible/vscode-ansible/pull/1888

### DIFF
--- a/ansible_ai_connect/ai/api/views.py
+++ b/ansible_ai_connect/ai/api/views.py
@@ -998,6 +998,7 @@ class GenerationRole(AACSAPIView):
         roles, files, outline, warnings = llm.invoke(
             RoleGenerationParameters.init(
                 request=request,
+                name=self.validated_data["name"],
                 text=self.validated_data["text"],
                 outline=self.validated_data["outline"],
                 model_id=self.req_model_id,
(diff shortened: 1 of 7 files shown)


ansible_ai_connect/ai/api/views.py
```diff
@@ -1021,7 +1023,7 @@ def post(self, request) -> Response:
         )
 
         answer = {
-            "role": anonymized_role,
+            "name": anonymized_role,
```
manstis, line 1026: It seems a little _weird_ that `name` is the anonymised `role`.. what's the reasoning?

manstis (dismissed):
LGTM 👍 

Just one comment about something I clearly don't fully understand.

ansible_ai_connect/ai/api/views.py
goneri: `name` is the name of the role, `role` was the former name of the key. I renamed it `name` for consistency since it's also the name of the WCA key.
The key is used as parameter and in the answer.

manstis (dismissed):
LGTM 👍

manstis (dismissed):
LGTM 👍

manstis (approved):
LGTM :+1:

PR #1578, opened 2025-03-19
AAP-41234: Automate AAP Chatbot UI versioning and publishing.

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-41234>

## Description
Adding a couple of GH workflows that automate versioning and publishing of the AAP Chatbot UI.

**ACTION**: A new  `NPM_TOKEN` secret must be set in this repository in order to run the publish action.

### Scenarios tested
Tested in my own fork.

## Production deployment
- [] This code change requires the following considerations before going to production:


.github/workflows/ui_publish_aap_chatbot.yml
```diff
@@ -0,0 +1,43 @@
+name: Chatbot - Build and Test
```
manstis, line 1: Probably want to change this name (to avoid duplication with the Action that does the build and test and is called the same)

manstis (approved):
LGTM 👍 

I haven't tested them. 

Just a comment you might want to consider.

manstis (dismissed):
LGTM 👍

romartin:
I added also a test run to ensure we are all good passing before releasing / publishing as well.

manstis (approved):
LGTM 👍

PR #1579, opened 2025-03-20
AAP-42324: Legacy configuration: Defaults are used when ANSIBLE_AI_MODEL_MESH_CONFIG is an empty string

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-42324

## Description
When `ANSIBLE_AI_MODEL_MESH_CONFIG` is provided, but an empty string use it to define Model Pipeline configuration.

## Testing
Run the service with `ANSIBLE_AI_MODEL_MESH_CONFIG` set to `""`.

All features should show as disabled in `/check/status`.

### Steps to test
As above.

### Scenarios tested
As above.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/config_loader.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/config_loader.py
@@ -37,7 +37,7 @@ def load_config_from_str(model_pipeline_config: str) -> Configuration:
     result = {}
     errors: [Exception] = []
 
-    if model_pipeline_config != "":
+    if model_pipeline_config:
         result = load_json(model_pipeline_config)
         if isinstance(result, Exception):
             errors.append(result)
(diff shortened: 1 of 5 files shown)


ansible_ai_connect/ai/api/model_pipelines/config_loader.py
```diff
@@ -28,41 +28,48 @@
...
     # yaml.safe_load(..) seems to also support loading JSON. Nice.
     # However, try to load JSON with the correct _loader_ first in case of corner cases
+    result = {}
     errors: [Exception] = []
-    result = load_json()
-    if isinstance(result, Exception):
-        errors.append(result)
-        result = load_yaml()
+
+    if model_pipeline_config != "":
```
goneri, line 40: you can just do `if model_pipeline_config:` Since `""` is considered `False` and this allow you to handle the case value is `0`, `[]`, `{}` or `None`, all of them being cases you want to avoid..

goneri (approved):

ansible_ai_connect/ai/api/model_pipelines/config_loader.py
manstis: Added as new commit @goneri 

I'll squash the commits when merging.

goneri (approved):

PR #1580, opened 2025-03-20
AAP Auth issue with local setup

TamiTakamiya:
Jira Issue:  n/a

## Description

This issue was found in the AAP Gateway integration work performed by @ldjebran.

In local AAP setup, Lightspeed OAuth authentication code makes a call to retrieve the license information. The received data contains license information in the following form:
```json
    "license_info": {
        "license_type": "open",
        "valid_key": true,
        "subscription_name": "OPEN",
        "product_name": "AWX"
    },
```
The current AAP Authentication code assumes that `license_info` always contains the key `date_expired` and fails
in the `user_has_valid_license()` method of the `AAPOAuth2` class.

According to [AWX source code](https://github.com/ansible/awx/blob/devel/awx/main/access.py#L355-L358), when `license_type` is `open`, we can assume that the user has a valid license. We also need to modify our logic to accept `license_info` with  `license_type` = `open` as a valid one.

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit tests + manual test w/ local AAP setup

## Production deployment

### DIFF
--- a/ansible_ai_connect/users/auth.py
+++ b/ansible_ai_connect/users/auth.py
@@ -69,7 +69,12 @@ class AAPOAuth2(BaseOAuth2):
     def user_has_valid_license(self, access_token):
         url = self.get_config_endpoint(settings.AAP_API_URL)
         data = self.get_json(url, headers={"Authorization": f"bearer {access_token}"})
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM 👍

PR #1585, opened 2025-03-25
[DRAFT] streaming_agent endpoint

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-NNNN>

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
@@ -21,8 +21,10 @@ import aiohttp
 import requests
 from django.http import StreamingHttpResponse
 from health_check.exceptions import ServiceUnavailable
+from llama_stack_client.types.agents import AgentTurnResponseStreamChunk
 
 from ansible_ai_connect.ai.api.exceptions import (
+    AgentInternalServerException,
     ChatbotForbiddenException,
     ChatbotInternalServerException,
     ChatbotPromptTooLongException,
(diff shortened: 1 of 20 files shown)


ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -309,3 +329,146 @@ async def async_invoke(self, params: StreamingChatBotParameters) -> AsyncGenerat
                     }
                     yield json.dumps(error).encode("utf-8")
                     return
+
+
+@Register(api_type="http")
```
manstis, line 334: I was thinking @TamiTakamiya 

Use of llama-stack is really a different type of _provider_ and not a new _feature_.

Therefore, the should be a new `t_model_mesh_api_type` [here](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/main/settings/types.py#L22-L32).

You could copy/paste the entire `http` folder, renaming it something like `llamastack` and move this pipeline implementation to it. L#334 would be `@register(api_type="llamastack")` etc. The [`provider`](https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/main/settings/legacy.py#L187) in `ANSIBLE_AI_MODEL_MESH_CONFIG` for ChatBot would be `llamastack`.

Does this make sense to you?

TamiTakamiya: Yes, it does and it looks cleaner. My original intension was just to have the llama stack pipeline with keeping the existing streaming chat pipeline as it is.

manstis:
Replaced by https://github.com/ansible/ansible-ai-connect-service/pull/1587/

Keeping open (for now) for @TamiTakamiya

manstis:
Closing @TamiTakamiya request.

PR #1573, opened 2025-03-18
PR to update the Chatbot UI removing the display mode from UI

justjais:
Jira Issue: <https://issues.redhat.com/browse/AAP-41008>

## Description

PR to update the Chatbot UI removing the display mode from UI

### Steps to test
1. Pull down the PR
2. Run the Chatbot UI integrated with AAP UI
3. Display mode button should be removed from Chatbot UI and by default the display mode should be in full screen.

### Scenarios tested

NA

## Production deployment

- [x] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1586, opened 2025-03-26
Update PF6 chatbot to 2.2.0-prerelease.37

TamiTakamiya:
Jira Issue: n/a

## Description

Update PF6 chatbot to 2.2.0-prerelease.37

### Steps to test
1. Pull down the PR
2. Run unit tests.

~~**NOTE** Currently unit tests in `/aap_chatbot` are failing. It's under investigation.~~ Fixes for unit test failures are also included in this PR.

## Production deployment


aap_chatbot/src/App.test.tsx
```diff
@@ -554,28 +554,28 @@ test("Color theme switch", async () => {
...
-  await expect
-    .element(
-      page.getByText(
-        "In Ansible, the precedence of variables is determined by the order...",
-      ),
-    )
-    .toBeVisible();
-  await expect.element(page.getByText("Create variables")).toBeVisible();
-});
+// test("Debug mode test", async () => {
```
manstis, line 557: Why does the test need removing?

manstis (commented):
LGTM :+1: 

I just want to understand why the test is removed.

Does it not work with the newer PatternFly?

aap_chatbot/src/App.test.tsx
TamiTakamiya: We have removed some UI elements that are shown only in Debug mode with #1573 and the test references one of those removed elements. I commented out the test so that I can uncomment it later when we merge `/aap_chatbot` and `/ansible_ai_connect_chatbot`.

manstis (approved):
Thank-you for the explanation @TamiTakamiya 

LGTM 👍

PR #1590, opened 2025-03-27
remove the grpc pipeline

goneri:
This is pipeline than we haven't used for years now and the maintain is a source
of unnecessary efforts.


manstis (approved):
LGTM :+1: 

(and good riddance)

PR #1591, opened 2025-03-27
get_api_key(): remove the second parameter

goneri:
The second parameter of `get_api_key()` was redondant and already exposed through the `user.organization.id` key.
By removing it, we simplify our tests and how we call the function itself.


manstis (approved):
Thank-you for cleaning up more of my mess.

LGTM :+1:

goneri:
> Thank-you for cleaning up more of my mess.
> 
> LGTM 👍

I believe these design decisions are from the dark ages of the pre-Michael-Anstis era!

manstis (approved):
LGTM :+1:

PR #1593, opened 2025-03-30
Update disclaimer text

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-41821>

## Description

Update disclaimer text. Change codes are:

1. Updated the disclaimer text that is displayed when the chatbot window is opened.
2. Updated the label text that is displayed on the link at the bottom of the chatbot window.
|
image

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit test + manual test on local server with AAP-UI integration

## Production deployment

### DIFF
--- a/aap_chatbot/src/App.test.tsx
+++ b/aap_chatbot/src/App.test.tsx
@@ -293,7 +293,7 @@ test("Basic chatbot interaction", async () => {
     .not.toBeInTheDocument();
 
   const footNoteLink = page.getByText(
-    "Lightspeed uses AI. Check for mistakes.",
+    "Always review AI-generated content prior to use.",
   );
   await footNoteLink.click();
   await expect
@@ -301,7 +301,7 @@ test("Basic chatbot interaction", async () => {
     .toBeVisible();
   await page.getByText("Got it").click();
   await expect
-    .element(view.getByText("While Lightspeed strives for accuracy,"))
+    .element(page.getByText("While Lightspeed strives for accuracy,"))
     .not.toBeVisible();
 
   await textArea.fill("Tell me about Ansible.");
--- a/aap_chatbot/src/useChatbot/useChatbot.ts
+++ b/aap_chatbot/src/useChatbot/useChatbot.ts
@@ -60,12 +60,11 @@ export const inDebugMode = () => {
 
 export const isStreamingSupported = () => {
   // For making streaming mode debug easier.
-  //   if (!import.meta.env.PROD && import.meta.env.MODE.includes("stream")) {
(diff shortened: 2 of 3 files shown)


manstis (approved):
LGTM 👍

TamiTakamiya:
@manstis I have re-read the doc on the legal requrement and found the "Always review AI-generated content prior to use" message should be

> Persistent message in interface or for each Output from LLM 

which means that as long as the message is persistent at the bottom of the chatbot window, we do not show the message for each output from LLM.  So I have updated the code and the change is now much simpler...

aap_chatbot/src/App.test.tsx
```diff
@@ -293,15 +293,15 @@ test("Basic chatbot interaction", async () => {
...
+    "Always review AI-generated content prior to use.",
   );
   await footNoteLink.click();
   await expect
     .element(page.getByText("While Lightspeed strives for accuracy,"))
     .toBeVisible();
   await page.getByText("Got it").click();
   await expect
-    .element(view.getByText("While Lightspeed strives for accuracy,"))
+    .element(page.getByText("While Lightspeed strives for accuracy,"))
```
TamiTakamiya, line 304: It was a bug in the original code.

aap_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -60,12 +60,11 @@ export const inDebugMode = () => {
 
 export const isStreamingSupported = () => {
   // For making streaming mode debug easier.
-  //   if (!import.meta.env.PROD && import.meta.env.MODE.includes("stream")) {
```
TamiTakamiya, line 63: This change was added for hack-a-thon version. We need to wait until we can  fully enable streaming with AAP-UI integration. The original logic disables streaming by default.

aap_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -363,8 +362,7 @@ export const useChatbot = () => {
         chatRequest.media_type = "application/json";
         await fetchEventSource(
           import.meta.env.PROD
-            ? //             ? "/api/lightspeed/v1/ai/streaming_chat/"
-              "/api/lightspeed/v1/ai/streaming_agent/"
```
TamiTakamiya, line 367: This was also added for hack-a-thon and even it is not correct for the final version of hack-a-thon code. I revert the change to point to the streaming_chat endpoint.

manstis (approved):
LGTM 👍

PR #1595, opened 2025-04-01
UI improvement for agent workflow

TamiTakamiya:
Jira Issue: n/a

## Description

Small UI improvement PoC for agent workflow. Intermediate results are displayed when the workflow is being processed and they are hidden in a collapsed area when the final result is obtained. 

https://github.com/user-attachments/assets/44c8e705-e7b8-44df-8af6-1d3b2fb9040c

### Steps to test
1. Pull down the PR
2. Run unit test

### Scenarios tested

Unit test + manual test

## Production deployment

### DIFF
--- a/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
+++ b/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
@@ -1,5 +1,9 @@
 import React, { useEffect, useRef, useState } from "react";
-import { Content, ContentVariants } from "@patternfly/react-core";
+import {
+  Content,
+  ContentVariants,
+  ExpandableSection,
+} from "@patternfly/react-core";
 import ChatbotContent from "@patternfly/chatbot/dist/dynamic/ChatbotContent";
 import ChatbotWelcomePrompt from "@patternfly/chatbot/dist/dynamic/ChatbotWelcomePrompt";
 import ChatbotFooter, {
@@ -292,6 +296,7 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
                       {
                         referenced_documents,
                         scrollToHere,
+                        collapse,
                         ...message
                       }: ExtendedMessage,
                       index,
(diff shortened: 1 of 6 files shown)


aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -303,14 +308,27 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
...
-                          <Message key={`m_msg_${index}`} {...message} />
-                          <ReferencedDocuments
-                            key={`m_docs_${index}`}
-                            caption="Refer to the following for more information:"
-                            referenced_documents={referenced_documents}
-                          />
-                        </div>
+                        {collapse ? (
+                          <div key={`m_div_${index}`}>
+                            <ExpandableSection toggleText="Show more">
```
manstis, line 313: You might like to follow the example [here](https://www.patternfly.org/components/expandable-section/).

By wrapping an `ExpandableSection` in your own (for exmaple) `ExpandableSectionBasic` would allow you to toggle the text between "Show more" (when collapsed) and "Show less" (when expanded). 

At the moment your text is always "Show more" even when expanded.

Just a thought.

TamiTakamiya: Initially I thought it would be an easy change, but I saw the sample code on the PatternFly page used React' "state". Since there might be multiple ExpandableSection's, there would be multiple states and I do not want to have many states on a page. We may be able to have a different approach. I will merge this as it is and open a lower priority JIRA ticket for that.

manstis: Ah, but the state is encapsulated in an instance of `ExpandableSectionBasic` so you can repeat as many times as you like without fear of maintaining multiple states at the _page_ level.

manstis: You'd use your `ExpandableSectionBasic` and not PF's `ExpandableSection` directly.

TamiTakamiya: https://issues.redhat.com/browse/AAP-43218 was opened for this.

PR #1598, opened 2025-04-02
Turn on streaming chat only when streaming chat is enabled on lightspeed

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-43200>

## Description

When the chatbot is running in the integrated environment with AAP, we want to enable chat streaming only when the lightspeed is configured to support chat streaming.  If only non-streaming mode is enabled on lightspeed, UI will use the non-streaming mode.

### Steps to test
1. Pull down the PR
2. Run uni tests

### Scenarios tested

Unit test and manual test on local AAP integration environment.

## Production deployment


romartin (approved):
LGTM. Thanks @TamiTakamiya

PR #1616, opened 2025-04-10
AAP chatbot - Sign commits when releasing.

romartin:
Just an small fix for signing the commits when creating the PR for releasing  AAP chatbot.


omaciel (approved):
Good to go

PR #1617, opened 2025-04-10
Fix the push pipeline by updating the newly added SAST checks

mabulgu:
No jira

## Description
see this conversation: https://redhat-internal.slack.com/archives/C04G3TZBGHJ/p1744298725638969

## Testing
We can only test this after the merge unfortunately.

## Production deployment


mabulgu:
@TamiTakamiya the move is just for visual concerns. the actual change is: I was using a reference for a task that doesnt exist. I had to use "build-container" instead and run these after it just like sast-snyk check. thanks for your review!

PR #1624, opened 2025-04-16
AAP-42800: trial: give better 403 errors

goneri:
Return better 403 errors to help the user to understand why the Trial
was refused.

image


manstis (dismissed):
LGTM 👍 

Although if you want these to be better handled in VSCode you _might_ need to think about writing one of _our_ exception classes (with `code` in it).. Not sure... Just a thought.

goneri:
> LGTM 👍
> 
> Although if you want these to be better handled in VSCode you _might_ need to think about writing one of _our_ exception classes (with `code` in it).. Not sure... Just a thought.

This is not an API end-point, just a regular one.

manstis (approved):
LGTM 👍

PR #1625, opened 2025-04-21
Chat copy button is not working in streaming mode

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-44267>

## Description

The copy icon on the Chatbot UI was not working when the chatbot was running in the streaming mode because the code was expecting that response text was included in the API response while the response text is sent through a stream in the streaming mode.

This PR also contains a change in the logic to copy referenced documents. In the existing code, only document URLs were copied. With this PR, both title and URL are copied in the markdown format.

Another small change in PR is the removal of  "action-button-clicked", which was introduced for ThumbsUp/Down icons what are not clicked more than once. Since the copy icon may be clicked more than once, it should not be applied to it.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests + manual tests with local servers both in AAP-integration and standalone environments.

## Production deployment

### DIFF
--- a/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
+++ b/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
@@ -28,7 +28,11 @@ import {
   ChatbotToggle,
   Conversation,
 } from "@patternfly/chatbot";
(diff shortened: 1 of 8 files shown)


justjais (approved):
LGTM

PR #1626, opened 2025-04-22
AAP-40232: Operator: auth_config_secret_name configuration is optional

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-40232

## Description
If Django's Social Authentication mechanism has not been properly configured then don't show the "Login" button.

The change here is triggered by the Operator, `ansible-ai-connect-operator`, [changes](https://github.com/ansible/ansible-ai-connect-operator/pull/517). 

When `auth_config_secret_name` is not provided no AAP authentication is configured. With AAP 2.5+ and AAP 2.6 onwards authentication is moving to the APP Gateway and separate configuration becoming obsolete. There are also discussions about removing the UI from `ansible-ai-connect-service` all together when deployed as part of AAP. The "Login" screen, when deployed "onprem", is therefore likely to be killed off completely at some point in time.
 
## Testing
Deploy the service without any Authentication configuration.

For example:
```
DEPLOYMENT_MODE = "saas"
# SOCIAL_AUTH_OIDC_OIDC_ENDPOINT = "something"
# SOCIAL_AUTH_OIDC_KEY = "something"
# SOCIAL_AUTH_OIDC_SECRET = "something"
```
Navigate to the _homepage_ that will no longer have a "Login" button.

See JIRA for screenshot.

### Steps to test
As above.

### Scenarios tested
As above.

## Production deployment


ansible_ai_connect/users/views.py
```diff
@@ -59,6 +59,26 @@ def __init__(self, *args, **kwargs):
...
+            case "onprem":
+                self.is_auth_configured = (
+                    settings.AAP_API_URL
+                    and settings.SOCIAL_AUTH_AAP_KEY
+                    and settings.SOCIAL_AUTH_AAP_SECRET
+                )
+            case "upstream":
+                self.is_auth_configured = (
+                    settings.SOCIAL_AUTH_GITHUB_KEY or settings.SOCIAL_AUTH_GITHUB_TEAM_KEY
+                )
```
goneri, line 80: I believe it would be more elegant to have a `case _:` and set `self.is_auth_configured = True` there.

hasys:
Code looks good, works good to me as well. Thank you Michael!

image
image

hasys (approved):

ansible_ai_connect/users/views.py
goneri: Also, it's a situation that should never happen. Instead of trying to deal with it, we should IMO instead raise a `ValueError` exception instead.

goneri (approved):
Looks good, to minor comments.

ansible_ai_connect/users/views.py
manstis: @goneri Let me try to understand.

Is `DEPLOYMENT_MODE == "upstream"` obsolete?

Are the various `SOCIAL_AUTH_GITHUB_xxx` settings obsolete?

I'd rather create a _maintenance_ JIRA to remove it entirely than raise an exception.

manstis:
@goneri @hasys Could you please re-approve?

I added a new commit tweaking a couple of points raised by @goneri 

I'll squash commits when merging.

hasys (approved):

goneri (approved):

PR #1627, opened 2025-04-23
Feedback Github issue contains zero UUID

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-44465>

## Description

When a user clicks the thumbs up/down feedback icons after the chatbot provides a response, they are redirected to GitHub with a pre-populated feedback form. One of the fields, Conversation ID, is expected to be populated with a unique UUID identifying the specific chat session. However, the field is currently being populated with a default all-zero UUID, which prevents proper identification and traceability of the feedback.

This occurs in streaming mode only. Previously, the code assumed a conversation id is included in the API response. It is no longer true in streaming mode. In streaming mode, a conversation id is sent in a stream and stored in a React state.

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit test + manual tests using local server (both standalone and AAP-UI integration environments)

## Production deployment

### DIFF
--- a/aap_chatbot/src/useChatbot/useChatbot.ts
+++ b/aap_chatbot/src/useChatbot/useChatbot.ts
@@ -75,7 +75,10 @@ export const fixedMessage = (content: string): MessageProps => ({
   timestamp: getTimestamp(),
 });
 
-export const feedbackMessage = (f: ChatFeedback): MessageProps => ({
+export const feedbackMessage = (
+  f: ChatFeedback,
+  conversation_id: string,
+): MessageProps => ({
   role: "bot",
   content:
     f.sentiment === Sentiment.THUMBS_UP
@@ -94,7 +97,9 @@ export const feedbackMessage = (f: ChatFeedback): MessageProps => ({
             content: "Sure!",
             id: "response",
             onClick: () =>
-              window.open(createGitHubIssueURL(f), "_blank")?.focus(),
+              window
+                .open(createGitHubIssueURL(f, conversation_id), "_blank")
+                ?.focus(),
           },
         ],
 });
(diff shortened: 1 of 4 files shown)


aap_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -140,6 +148,20 @@ export const useChatbot = () => {
   const [conversationId, setConversationId] = useState<
     string | null | undefined
   >(undefined);
+
+  // Workaround for the lag issue of the conversation_id state value.
+  const getConversationId = () => {
+    let id;
+    setConversationId((value) => {
```
manstis, line 155: Out of curiosity why do you use the `setConversationId(lamda)` approach vs just using `this.conversationId` on the `useState` hook?

manstis (approved):
LGTM :+1: 

I would however like to understand your approach better :smile:

aap_chatbot/src/useChatbot/useChatbot.ts
TamiTakamiya: I initially wrote:
```
const getConversationId = () => conversationId ? conversationId : "00000000-0000-0000-0000-000000000000";
```
and found `conversationId` contained `undefined` even if after `setConversationId` is called with a valid value returned in the `start` stream data.

I learned React's state updates the vaule is not updated immediately and the update occurs after rendering. Even though I thought that rendering should have occurred, it seems true that there is a time lag between `setConversationId` is called and the value is reflected in `conversationId`.

However, there is a way to retrieve the value that was set most recently and it is the way to call `setConversationId` with a lamba function, i.e., following is called,
```
setConversationId((prevValue) => { return newValue; });
```
`prevValue` contains the value most recently set.  I think it is provided for implementing features like counters, which depend on the previous values.

In the following code,
```
const getConversationId = () => {
    let id;
    setConversationId((value) => {
      id = value;
      return value;
    });
    if (!id) {
      id = "00000000-0000-0000-0000-000000000000";
    }
    return id;
  };
```
`setConversationId` receives `value` and returns `value` without changes, i.e. it does not change `conversationId`, but `id` is set to `value`.  Even if `conversationId` still set to `undefined`, `id` can access to the new conersation ID value, if `setConversationId` was called.  That's why I wrote the code in that way.

PR #1630, opened 2025-04-23
Show 429 message in streaming chat mode

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-44458>

## Description

When the chatbot API received too many requests in a certain period of time, a 429 status code is returned and a message is displayed in the non-streaming chat mode. However, the logic is not working in the streaming mode.  This PR is for enabling the logic in the streaming code.

### Steps to test
1. Pull down the PR
2. Run unit test.

### Scenarios tested

Unit test + Manual test

## Production deployment


aap_chatbot/src/App.test.tsx
```diff
@@ -822,3 +822,22 @@ test("Chat streaming error in status check", async () => {
   const textContent = alert?.textContent;
   expect(textContent).not.contain("mocked error");
 });
+
+test("Chat service returns 429 Too Many Requests error in streaming", async () => {
+  mockAxios(200, false, false, referencedDocumentExample, true);
```
manstis, line 827: I'm confused :smile: 

The _streaming_ version of the Chatbot seems to use [`fetchEventSource(..)`](https://github.com/ansible/ansible-ai-connect-service/blob/4996c458cfb50f87256378ce217fa20c4dda8076/aap_chatbot/src/useChatbot/useChatbot.ts#L443) whereas this test is using `mockAxiox(....., true)`. I'd have expected this test to mocked [`mockFetchEventSource()`](https://github.com/ansible/ansible-ai-connect-service/blob/4996c458cfb50f87256378ce217fa20c4dda8076/aap_chatbot/src/App.test.tsx#L336C5-L336C45).

The code looks fine.. I'm just confused as to why you mock one thing; but the code appears to use something different.

manstis (commented):
@TamiTakamiya Before approving; I'd like to understand something a bit more.

aap_chatbot/src/App.test.tsx
TamiTakamiya: I admit it's very confusing. 

It is for mocking [this health check call](https://github.com/ansible/ansible-ai-connect-service/blob/main/aap_chatbot/src/useChatbot/useChatbot.ts#L175).  The communication receiving chatbot data is mocked with `mockFetchEventSource()`.

manstis (approved):
Thanks for the explanation @TamiTakamiya 

LGTM :+1:

PR #1632, opened 2025-04-24
AAP-44422: Chatbot history not cleared across logout/login as different user.

romartin:
Jira Issue: <https://issues.redhat.com/browse/AAP-44422>

## Description
The goal is to clean the conversation state once un-mounting the component, which will happen on AAP-UI once either closing panel, logging out or refreshing the page.

Initially I tried to move the `conversationList` and `conversationStore` into the `AnsibleChatbot` as managed state, but I got into weird issues, so just decided to create a function to clean those objects once un-mounting the component.  Works properly. Test added.

## Testing
Tested both `ansible_ai_connect_chatbot` and `aap_chatbot` apps, by using a local chatbot service. 

### Scenarios tested
Tested also the AAP-UI integration, conversations are cleared after logging out.

## Production deployment


TamiTakamiya (approved):
LGTM. Thanks!

PR #1633, opened 2025-04-25
Hide action icons while streaming

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-44421>

## Description

Hide action icons (ThumbsUp/Down, Copy) while streaming. Even though it does not provide any new functions, message rendering looks smoother. 

### Steps to test
1. Pull down the PR
2. Run unt tests

### Scenarios tested

Unit + manual tests

## Production deployment

### DIFF
--- a/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
+++ b/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
@@ -253,7 +253,11 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
                         {collapse ? (
                           <div key={`m_div_${index}`}>
                             <ExpandableSection toggleText="Show more">
-                              <Message key={`m_msg_${index}`} {...message} />
+                              <Message
+                                key={`m_msg_${index}`}
+                                {...message}
+                                isLoading={isLoading && !message.content}
+                              />
                               <ReferencedDocuments
                                 key={`m_docs_${index}`}
                                 caption={REFERENCED_DOCUMENTS_CAPTION}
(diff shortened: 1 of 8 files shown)


aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -274,7 +278,7 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
                       </div>
                     ),
                   )}
-                  {isLoading ? (
+                  {messages.at(-1)?.role === "user" && isLoading ? (
```
TamiTakamiya, line 281: There is a small time period when `isLoading` is set and a bot message is already added to the last of messages list. In such a case, we do not need to the loading animation icon.

aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -253,7 +253,11 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
                         {collapse ? (
                           <div key={`m_div_${index}`}>
                             <ExpandableSection toggleText="Show more">
-                              <Message key={`m_msg_${index}`} {...message} />
+                              <Message
+                                key={`m_msg_${index}`}
+                                {...message}
+                                isLoading={isLoading && !message.content}
+                              />
```
TamiTakamiya, line 260: If `isLoading` is set and message content is still empty, display the loading animation icon instead of showing an empty string.

romartin (approved):

PR #1637, opened 2025-04-25
IFrame tooltip fix

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-43962>

## Description

PR to fix Chatbot header menu tooltip slipping out of iframe and getting displayed over AAP UI.  Extends the #1636 changes.

### Steps to test
1. Pull down the PR
2. Run chatbot in AAP-integration env and test manually

### Scenarios tested

Manual test only.  This requires Iframe, but currently our test automation does not run chatbot in an iframe.

## Production deployment

### DIFF
--- a/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
+++ b/aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
@@ -13,7 +13,7 @@ import ChatbotHeader, {
 } from "@patternfly/chatbot/dist/dynamic/ChatbotHeader";
 
 import "./AnsibleChatbot.scss";
-import { inDebugMode, useChatbot } from "../useChatbot/useChatbot";
+import { bodyElement, inDebugMode, useChatbot } from "../useChatbot/useChatbot";
 import { ReferencedDocuments } from "../ReferencedDocuments/ReferencedDocuments";
 
 import type { ExtendedMessage } from "../types/Message";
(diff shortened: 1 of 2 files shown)


aap_chatbot/src/AnsibleChatbot/AnsibleChatbot.tsx
```diff
@@ -319,6 +315,16 @@ export const AnsibleChatbot: React.FunctionComponent<ChatbotContext> = (
                   hasAttachButton={false}
                   hasStopButton={hasStopButton}
                   handleStopButton={handleStopButton}
+                  buttonProps={{
+                    //@ts-ignore
```
TamiTakamiya, line 319: These @ts-ignore tags are a workaround for https://github.com/patternfly/chatbot/issues/536

aap_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -139,6 +139,9 @@ const createGitHubIssueURL = (
   return url.toString();
 };
 
+// For fixing tooltips that pops up from an iframe
+export let bodyElement = document.body;
```
manstis, line 143: I wonder whether it's better to return `bodyElement` from the hook vs adding this global? We've already had problems with global (and user state/messages). Are you certain this works ok? Will it still have a value when the `AnsibleChatbot` unmount?

romartin (approved):

manstis (commented):
IIUC you're passing a reference to the outer parent body to the inner child components in the `IFRAME`? I'm just concerned about use of another global. Do you perceive any issues?

PR #1642, opened 2025-04-30
AAP-45044: Chatbot: Add new status check endpoint

manstis:
Jira Issue: https://issues.redhat.com/browse/AAP-45044

## Description
This PR adds a dedicated endpoint to check the Chatbot status/health.

The existing code uses the `/health/status` endpoint that returns the status of *all* components.

If one of these is mis-configured the endpoint returns HTTP500 and the Chatbot UI code assumes it is unavailable.

### Testing the service
1. Run `ansible-ai-connect-service` locally
2. Check the `api/v1/health/status/chatbot` endpoint.

Example response:
```
HTTP/1.1 200 OK
Content-Type: application/json

{
  "chatbot-service": "disabled",
  "streaming-chatbot-service": "ok"
}
```

### Testing AAP UI
1. `cd aap-ui`
2. `npm ci`
3. You will need to change `frontend/chatbot/ChatbotState.tsx`:
```
...
const LIGHTSPEED_HEALTH_STATUS_PATH = `/api/lightspeed/v1/health/status/chatbot/`;
  ...
  const setChatbotEnablementState = useCallback(async () => {
    try {
      const lightspeedStatusResponse = await get(LIGHTSPEED_HEALTH_STATUS_PATH, {});
      const state: string = lightspeedStatusResponse['chatbot-service'];
      if (state === 'disabled') {
        setChatbotState(ChatbotStateEnum.Disabled);
      } else if (state === 'ok') {
        setChatbotState(ChatbotStateEnum.Closed);
      } else {
        setChatbotState(ChatbotStateEnum.Disabled);
      }
    } catch (e) {
      setChatbotState(ChatbotStateEnum.Disabled);
    }
  }, [get, setChatbotState]);
  ...
```
4. You will need to change `frontend/chatbot/interfaces/LightspeedStatus.tsx`:
```
export interface LightspeedStatusResponse {
  'chatbot-service': string;
  'streaming-chatbot-service': string;
}
```
5. `export PLATFORM=<See JIRA for the URL>`
6. `cd ./platform`
7. `npm run start`
8. Login as `manstis` (See JIRA for password)

The above AAP instance uses the Lightspeed service container built by this PR.

A PR to `aap-ui` is in progress.. but I don't have access to create a PR yet.

### Steps to test
As above.

### Scenarios tested
As above.

## Production deployment

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/types.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/types.py
@@ -1,0 +1,18 @@
+#  Copyright Red Hat
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import TypeVar
+
+PIPELINE_PARAMETERS = TypeVar("PIPELINE_PARAMETERS")
+PIPELINE_RETURN = TypeVar("PIPELINE_RETURN")
+PIPELINE_TYPE = TypeVar("PIPELINE_TYPE")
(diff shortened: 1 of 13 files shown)


ansible_ai_connect/ai/api/model_pipelines/types.py
manstis, line 1: I moved these to a new file as it better supports their re-use.

Having them in the `pipelines.py` file led to circular dependencies.

manstis:
@TamiTakamiya FYI the issue arose in the AAP 2.5 testing.

Christian Adams had accidentially mis-configured WCA/`model_config_secret` and so the "models" health check was failing.

The existing endpoint returns HTTP500 *if* any component check fails.

This led to the Chatbot UI thinking it was unavailable.

I chatted with @ldjebran and we agreed the _cleanest_ solution was to have a dedicated endpoint.

TamiTakamiya (approved):
Looks good to me. Thanks!

PR #1644, opened 2025-05-02
type hints: cover main/settings/legacy.py

goneri:


ansible_ai_connect/main/settings/legacy.py
```diff
@@ -60,8 +60,8 @@ def load_from_env_vars():
 
     # Model API Timeout (in seconds). Default is None.
     model_service_timeout = (
-        int(os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT"))
-        if os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT")
+        cast(int, os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT"))
```
manstis, line 63: Are these still used?

They should be part of the JSON configuration.

manstis (dismissed):
LGTM 👍 (but I'd like to think it's not used).

ansible_ai_connect/main/settings/legacy.py
goneri: Yes, I had to, because `os.getenv()` return a string and pyright is not able (yet?) to take the `isnumeric()` into consideration to figure out that it's a `int`.

ansible_ai_connect/main/settings/legacy.py
```diff
@@ -60,8 +60,8 @@ def load_from_env_vars():
 
     # Model API Timeout (in seconds). Default is None.
     model_service_timeout = (
-        int(os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT"))
-        if os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT")
+        int(cast(int, os.getenv("ANSIBLE_AI_MODEL_MESH_API_TIMEOUT")))
```
manstis, line 63: This is getting awkward. Why the need to pass the result of a cast (to `int`) to the `int` constructor?! 

That said, I trust your judgement, LGTM 👍

manstis (approved):

PR #1650, opened 2025-05-15

goneri:


ansible_ai_connect/users/wca_auth.py
manstis, line 1: To be honest I don't understand why you've moved this out of the `ModelPipeline` `MetaData` where you still retrieve the Token in the `View`.

goneri: I was more like a hack at this stage.

PR #1651, opened 2025-05-15
remove the deprecated CIAMCheck class from the authz_check

goneri:
We've never used it in production and the CIAM team doesn't support it anymore.

See: AAP-20347


manstis (approved):
LGTM 👍 

🧹 🧹 🧹

PR #1652, opened 2025-05-16
requirements: bump Django to 4.2.21

goneri:
This to address a regression caused by the CVE-2025-26699 fix and CVE-2025-32873.

See: https://docs.djangoproject.com/en/5.2/releases/4.2.21/


manstis (approved):
LGTM 👍

PR #1656, opened 2025-05-22

goneri:


ansible_ai_connect/users/wca_auth.py
manstis, line 1: In all honesty I don't understand what moving this code achieves. You're passing a `provider_name` to presumably support different scenarios so why not have `get_api_key` and `get_model_id` on the `MetaData` interface and provide implementations as appropriate? IMO you're deconstructing polymorphic behavior into procedural logic. Sorry.

goneri: This was not really mean to be reviewed.

manstis: I guessed that when I saw your other PR... I never sleep. Always vigilant, always a nosey blighter.

PR #1657, opened 2025-05-22

goneri:


manstis (dismissed):
LGTM 👍 

🥳

ansible_ai_connect/users/views.py
```diff
@@ -188,6 +193,34 @@ def retrieve(self, request, *args, **kwargs):
...
+
+    def get_object(self):
+        return self.request.user
+
+    @method_decorator(cache_per_user(ME_USER_CACHE_TIMEOUT_SEC))
+    def get(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        config = apps.get_app_config("ai").get_model_pipeline(MetaData)
+        api_key = config.get_api_key(instance)
```
manstis, line 218: This is going to break if the _provider_ is not `wca`, `wca-onprem` or `wca-dummy`.

I know you protect the endpoint with `ANSIBLE_AI_ENABLE_WCA_TOKEN_GEN` however it might be safer if you moved `get_api_key` and `get_token` to `MetaData` with the default implementation to raise an exception that could be caught here to return something like HTTP 501? You could then simply override these methods in the WCA related implementations.

WDYT?

manstis (commented):
Overall this LGTM :+1: 

I have made a few comments it'd be good to hear your opinion on before approving.

ansible_ai_connect/users/views.py
goneri: I've defined `get_token()` and `get_api_token()` in `MetaData`, they just raise a `NotImplementedError`.

I don't want to over think this for now.

PR #1662, opened 2025-05-29
Add user uuid to WCA request headers

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-46219

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment


goneri:
@mabashian I believe you can inject header at this level https://github.com/ansible/ansible-ai-connect-service/blob/main/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_saas.py#L231-L245

This way we won't have to adjust all the views (and test) like this.

PR #1668, opened 2025-06-04
Upgrade llama-stack-client to 0.2.9

TamiTakamiya:
Jira Issue: n/a

## Description

Upgrade llama-stack-client library to version 0.2.9

### Steps to test
1. Pull down the PR
2. Run unit tests

### Scenarios tested

Unit tests

## Production deployment


manstis (approved):
Good catch @TamiTakamiya 

LGTM :+1: (and yes, I'll re-approve when it's converted from _draft_).

manstis:
@TamiTakamiya Why revert the x86 changes? Surely we'd need `llama-stack` upgraded on both architectures.

TamiTakamiya:
> @TamiTakamiya Why revert the x86 changes? Surely we'd need `llama-stack` upgraded on both architectures.

It's just for checking which change is causing the test failure. ..

TamiTakamiya:
[These lines](https://github.com/meta-llama/llama-stack-client-python/pull/230/files#diff-8b7e3d7f61a369533d971784a423a7488f71ffa782c5d951419ccb93cd70071bR6-R9) added with https://github.com/meta-llama/llama-stack-client-python/pull/230 seems to be causing this issue.  The problem disappears if they are commented out.

PR #1671, opened 2025-06-05
WCA/self_test(): ensure infer_from_parameters has the headers defined

goneri:


manstis (approved):
LGTM :+1: .... and 100% code coverage... hmmmm :smile_cat:

PR #1685, opened 2025-06-12
[DO NOT MERGE] Force 'ok' health response for llama-stack health check

manstis:
**=== DO NOT MERGE ===**

PR to create a container image with the `llama-stack` health check returning `'ok'`.


manstis:
No longer required.

PR #1687, opened 2025-06-12
AAP-34013: WCA: use a config key to disable anonymizer

goneri:
`enable_anonymization` is a new configuration key that can be
used to enable or disable the Anonymizer. The default is `True`.

To get anonymization you need:
- `enable_anonymization=True` at provider level (default is `True`)
- `enable_anonymization=True` at organization level (default is `True`)

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/tests/__init__.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/tests/__init__.py
@@ -131,6 +131,7 @@ def mock_pipeline_config(pipeline_provider: t_model_mesh_api_type, **kwargs):
                 one_click_default_model_id=extract(
                     "one_click_default_model_id", "one-click-default-model-id", **kwargs
                 ),
+                enable_anonymization=extract("enable_anonymization", True, **kwargs),
             )
         case "wca-onprem":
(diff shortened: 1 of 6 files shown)


goneri:
Actually, I'm wondering if it would not be more elegant to expose a new configuration key to the WCA provider :thinking:.

manstis:
> Actually, I'm wondering if it would not be more elegant to expose a new configuration key to the WCA provider 🤔.

Well... I'm not a fan of using environment variables when it only applies in certain circumstances. If anonymisation only applies to WCA then perhaps a configuration change is preferable... but can't anonymisation apply equally to any provider (now or in the future)?

goneri:
> > Actually, I'm wondering if it would not be more elegant to expose a new configuration key to the WCA provider 🤔.
> 
> Well... I'm not a fan of using environment variables when it only applies in certain circumstances. If anonymisation only applies to WCA then perhaps a configuration change is preferable... but can't anonymisation apply equally to any provider (now or in the future)?

Yes, I'm rewriting my PR. Regarding the other question, for now we only cover WCA, we may want to extend that later, it's not a difficult change.

manstis (approved):
Super. LGTM :+1:

PR #1688, opened 2025-06-13
Do not query vector DB for general greetings

TamiTakamiya:
Jira Issue:  n/a

## Description

Add a sentence to the instruction given to llama stack ChatAgent so that no vector DB queries take place for general geetings, such as "Hello" or "How are you?".

This PR also contains a fix for the issue that Granite 3.3 occasionally does not create a list of JSON object and returns a JSON object for one tool.  One sentence is added for forcing LLM (Granite) to create a JSON list even if a single tool is discovered. Without the change, the query `Are there any videos that explain AAP?` caused the problem.

### Steps to test
1. Pull down the PR
2. Run Lightspeed server and llama stack
3. Open chatbot UI on browser and type in "Hello" to verify no vector DB query is made.

### Scenarios tested

Manual testing only.

## Production deployment


manstis (approved):
Great to see you improving the prompt and tool handling. It's something we need to do and was to be the subject of discussion in our next sync meeting.

LGTM 👍

PR #1689, opened 2025-06-13
Do not display duplicated inference output

TamiTakamiya:
Jira Issue: n/a

## Description

When the agentic chatbot makes tool calls, the inference API is called to process outputs from tool calls and the last output is displayed as the final output.

With the current implementation of our chatbot UI, intermediate outputs are hidden in a collapible section.  However, when the section is expanded, the last inference output shows up as the duplicated information. We want to get rid of that.
Screenshot From 2025-06-13 12-11-26
Screenshot From 2025-06-13 12-08-49

When the agentic chat does not make tool calls, a single inference call generates the final output. In such cases, we do not need to have a collapsible section. We'd like to take it out as well.  Also sometimes Granite model inserts the string `<noinput>` and we want to remove it.
Screenshot From 2025-06-13 12-10-41
Screenshot From 2025-06-13 12-07-14

### Steps to test
1. Pull down the PR
2. ...
3. ...

## Production deployment

### DIFF
--- a/aap_chatbot/src/App.test.tsx
+++ b/aap_chatbot/src/App.test.tsx
@@ -249,7 +249,7 @@ function mockFetchEventSource() {
     },
     {
       event: "token",
-      data: { id: 0, token: "Let me search for " },
+      data: { id: 0, token: "\n\n`inference>`Let me search for " },
     },
     {
       event: "token",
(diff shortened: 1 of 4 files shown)


TamiTakamiya:
> We [have](https://github.com/lightspeed-core/lightspeed-providers/tree/main/lightspeed_stack_providers/providers/inline/agents/lightspeed_inline_agent). Perhaps create a JIRA so this discussion/requirement is not forgotten?

Opened [AAP-47742](https://issues.redhat.com/browse/AAP-47742)  for that change. Thanks!

TamiTakamiya:
@justjais Added a comment on RegEx options. Would you take a look?

PR #1711, opened 2025-07-16
Open /chatbot view to users whose organization has an AAP sub

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-49639

Assisted-by: Cursor (gemini-2.5-pro-preview)

## Description
We already had a permission class (IsOrganisationLightspeedSubscriber) which keyed off of the user's `rh_org_has_subscription` attribute.  I'm leveraging that class and adding it to the mix of permission classes already applied to the chatbot view as well as the chatbot-related endpoints.  This change should only add to the existing matrix of scenarios that allow a user to access the chatbot (and see the chatbot link on the home page).

## Testing
Tested locally by programmatically hard coding my user to fit this new scenario (ie, ensuring that user.rh_internal is false and user.rh_org_has_subscription is true) so that I could verify that the chatbot link is present on the home page and the chatbot renders when the user navigates to /chatbot.

### Scenarios tested
Unit tests added around testing the presence of the link as well as the ability to navigate to the chatbot page

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/tests/test_views.py
+++ b/ansible_ai_connect/main/tests/test_views.py
@@ -33,6 +33,7 @@ from ansible_ai_connect.ai.api.model_pipelines.nop.pipelines import (
 )
 from ansible_ai_connect.main.settings.base import SOCIAL_AUTH_OIDC_KEY
 from ansible_ai_connect.main.views import LoginView
+from ansible_ai_connect.organizations.models import Organization
 from ansible_ai_connect.test_utils import (
     APIVersionTestCaseBase,
     create_user_with_provider,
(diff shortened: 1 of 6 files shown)


manstis (approved):
LGTM :+1: 

Not assisted-by: Cursor (gemini-2.5-pro-preview)

ansible_ai_connect/main/tests/test_views.py
```diff
@@ -258,6 +259,7 @@ class TestChatbotView(TestCase):
 
     def setUp(self):
         super().setUp()
+        self.organization = Organization.objects.create(id=12345)
```
goneri, line 262: We've a function called `create_user_with_provider()` in `ansible_ai_connect/test_utils.py` to provision test users.

And if you're test class depends on `WisdomServiceAPITestCaseBaseOIDC`, you get a test user created for Free! And the Teardown is also included :-)

goneri (changes requested):
Sorry @mabashian for what may look like nitpicking, but our tests are already a bit out of control, and I don't want to add what will some be an extra layer of technical debts.

goneri (approved):

PR #1714, opened 2025-07-18
Remove chatbot link and related access logic from home page

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-49873

This pull request removes support for displaying and testing the "Chatbot" feature in the application. The changes primarily involve removing the logic, templates, and tests related to the chatbot functionality.

### Removal of Chatbot Functionality:

* **Tests removed**: Deleted multiple test cases related to verifying the presence and behavior of the chatbot link under various user conditions in `test_views.py`. These tests included scenarios for anonymous users, non-RH users, RH users, and users with different chatbot settings.

* **Template update**: Removed the conditional block in `home.html` that displayed the chatbot link in the UI when the chatbot feature was enabled and accessible to the user.

* **View logic update**: Eliminated the `can_access_chatbot` context variable from the `get_context_data` method in `views.py`, which determined user access to the chatbot link based on authentication and group membership.

## Production deployment

### DIFF
--- a/ansible_ai_connect/users/templates/users/trial.html
+++ b/ansible_ai_connect/users/templates/users/trial.html
@@ -143,9 +143,6 @@
                   {% if deployment_mode == 'saas' and user.is_authenticated and user.rh_user_is_org_admin %}
                     <a class="pf-l-level__item" href="/console"><span class="fas fa-solid fa-cog"></span> Admin Portal</a>
                   {% endif %}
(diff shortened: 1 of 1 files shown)


goneri (changes requested):
Can you do the same clean up in `ansible_ai_connect/users/templates/users/trial.html`?

goneri (approved):

PR #1715, opened 2025-07-21
req: upgrade langchain

goneri:
Also, remove `langchain_community` from the dev requirements. We don't need it.


manstis (approved):
LGTM :+1:

PR #1720, opened 2025-07-24
To remove the chat_prompt from Chatbot being sent to segment for logs analysis

justjais:
Jira Issue: <https://issues.redhat.com/browse/AAP-39508>

Assisted-by: Cursor AI 
Generated by: Unit tests

## Description

To remove the chat_prompt from Chatbot being sent to segment for logs analysis. Confirmed from Marty that we can disable the chat_prompts from being sent to segment and thus Amplitude to avoid leaking any user sensitive information. Please look at the latest comment over the issue: AAP-39508, for more reference

### Steps to test
1. Verified running the unit tests

### Scenarios tested

NA

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/settings/base.py
+++ b/ansible_ai_connect/main/settings/base.py
@@ -213,6 +213,9 @@ ANALYTICS_MIN_ANSIBLE_EXTENSION_VERSION = os.environ.get(
     "ANALYTICS_MIN_ANSIBLE_EXTENSION_VERSION", "v2.12.143"
 )
 
+# Control what data is sent to telemetry, by default we exclude chat_prompt
+SEGMENT_EXCLUDE_CHAT_PROMPTS = os.environ.get("SEGMENT_EXCLUDE_CHAT_PROMPTS", "true")
+
 OAUTH2_PROVIDER = {
     "SCOPES": {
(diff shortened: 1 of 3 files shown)


ansible_ai_connect/main/settings/base.py
```diff
@@ -213,6 +213,7 @@ def is_ssl_enabled(value: str) -> bool:
     "ANALYTICS_MIN_ANSIBLE_EXTENSION_VERSION", "v2.12.143"
 )
 
+
```
manstis, line 216: Well.. it does no harm; but does no good either.

Do we have to keep it?

ansible_ai_connect/wildcard_oauth2/test.yaml
manstis, line 1: What's this file for?

manstis (commented):
Just a couple of nit-picks.

manstis (approved):
LGTM :+1: 

@goneri did you have anything to add?

I know you're not a fan.. but this is at least consistent with other event filtering.

justjais:
I am going ahead with the merge, as updating the implementation around ALLOW_LIST if needed will need to be handled separately

PR #1725, opened 2025-08-01
Revert "containerFile: only use the requirements of the TARGETARCH (#1717)"

goneri:
`TARGETARCH` is not defined by default with `docker`. This ends up creating more troubles than expected.

This reverts commit 2624b982d9dc6fbd28c146cb8465dbd5126b768e.


manstis (approved):
LGTM :+1:

PR #1728, opened 2025-08-04
http/check_test: force timeout to 1s

goneri:
Avoid using the `self.timeout()` method which can lead to a sisution
where we set `timeout=None`.

Also rename the `self.timeout()` method to `task_gen_timeout()` to
improve the readability.

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
@@ -69,7 +69,7 @@ class HttpMetaData(MetaData[HttpConfiguration]):
         i = self.config.timeout
         self._timeout = int(i) if i is not None else None
 
-    def timeout(self, task_count=1):
+    def task_gen_timeout(self, task_count=1):
         return self._timeout * task_count if self._timeout else None
 
 
@@ -94,7 +94,7 @@ class HttpCompletionsPipeline(HttpMetaData, ModelPipelineCompletions[HttpConfigu
                 self._prediction_url,
                 headers=self.headers,
                 json=model_input,
-                timeout=self.timeout(task_count),
+                timeout=self.task_gen_timeout(task_count),
                 verify=self.config.verify_ssl,
             )
             result.raise_for_status()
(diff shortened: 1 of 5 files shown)


justjais (dismissed):
LGTM

manstis (commented):
WDYT?

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -144,7 +144,7 @@ def self_test(self) -> HealthCheckSummary:
             r = requests.get(
                 self.config.inference_url + "/readiness",
                 headers=headers,
-                timeout=self.timeout(1),
+                timeout=1,
```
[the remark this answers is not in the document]
goneri: I think we should not have `task_count` here at all. `task_count` is a parameter that only makes sense for multiline code gen. I'm actually updating my patch to rename the `timeout()` method to `task_gen_timeout()`.

manstis, line 147: I'd still like to be able to set the `timeout` for chat requests.

`ANSIBLE_AI_MODEL_MESH_CONFIG` supports configuring a `timeout` for the "Chatbot" pipelines.

This change negates the configuration option.

manstis (commented):
Whilst agreeing with renaming the `timeout` method to `task_gen_timeout` for `/completions` and related operations; you've removed the use of the `timeout` configuration parameter for `/chat` and `/streaming_chat` operations.

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
goneri: Thank you for catching this. It's definitely a mistake.

goneri: Oh crap, indeed. I just pushed an update, stay with `task_gen_timeout()` for now.

goneri:
> I think you missed a couple of updates?

... yes, I've also missed some in the test-suites :-).

PR #1729, opened 2025-08-05
Adjust project name in login view for chatbot route

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-50250

Assisted-by: claude-code

## Description
This is a bit of a messy "problem" to solve.  The project name comes in via a setting which we append to the end of "Log in to " on the login page.  The approach I've taken is to check to see if the ?next query param starts with /chatbot and then find/replace the intended text with an empty string.  I don't think this is a particularly clever approach nor is it robust.  If the setting (ANSIBLE_AI_PROJECT_NAME) string changes just a bit, this logic may break.  I also don't love hard-coding references to WCA in the view.  I'm open to suggestions for alternative ways to achieve this (or push back saying that this is a bad idea and we shouldn't do it).

## Testing
1. Pull down the PR
2. Make sure that ANSIBLE_AI_PROJECT_NAME is set to 'Ansible Lightspeed with IBM watsonx Code Assistant' (note that I also had to add - ANSIBLE_AI_PROJECT_NAME=${ANSIBLE_AI_PROJECT_NAME} to compose.yml)
3. Run the ai-connect-service
4. Navigate to /login (should see the full project name)
5. Navigate to /chatbot unauthenticated (should be redirected to /login and see the name without the WCA bits)

### Scenarios tested
1. Login screen renders correctly when no "next" param present
2. Login screen renders correctly when "next" param is something other than /chatbot
3. Login screen renders correctly when "next" param starts with /chatbot

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/views.py
+++ b/ansible_ai_connect/main/views.py
@@ -54,11 +54,11 @@ class LoginView(auth_views.LoginView):
         next_url = self.request.GET.get("next") or "/"
         context["next"] = next_url
         context["deployment_mode"] = settings.DEPLOYMENT_MODE
-        
+
         project_name = settings.ANSIBLE_AI_PROJECT_NAME
         if next_url and next_url.startswith("/chatbot"):
             project_name = project_name.replace(" with IBM watsonx Code Assistant", "")
-        
+
         context["project_name"] = project_name
         context["aap_api_provider_name"] = settings.AAP_API_PROVIDER_NAME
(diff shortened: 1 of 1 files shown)


manstis:
Have you thought of an inverse approach, i.e. add " with WatsonX blah blah" if Wisdom Service is configured to use WCA? All AI features are defined in "ModelPipelines" and each has a Provider ID. It'd be quite simple to iterate them and check for `wca` or `wca-onprem` and add the extra text.

I'm not at my laptop now but can talk more later if you're interested.

PR #1734, opened 2025-08-06
Append WCA suffix to project title based on service configuration

mabashian:
Jira Issue: https://issues.redhat.com/browse/AAP-50250

Assisted-by: claude-code

## Description
An alternative approach to https://github.com/ansible/ansible-ai-connect-service/pull/1729

This approach is a bit of the inverse of the other approach.  With these changes, the WCA bits are appended to the end of the project name and displayed on the login screen so long as one of the pipelines is configured with a wca provider.  This means that the project name no longer needs to contain the WCA bits as the suffix will be added programatically.  I set the default suffix in a new environment variable called `ANSIBLE_AI_PROJECT_WCA_SUFFIX`.  We could hard code the suffix in the application code but I thought it would be nice to have the ability to overwrite it with a variable if needed (but this can be easily changed).

This PR also adds the /chatbot "next" query parameter logic so that the suffix is not added if the user is attempting to navigate straight to the chatbot (which was the original issue outlined in https://issues.redhat.com/browse/AAP-50250

Note that I did add logic to NOT append the suffix if the suffix is already present in the project name.  That means that (I think) these changes should not impact any current deployments of the ai-connect-service.  In order to FULLY address AAP-50250 we'll need to make sure that any lightspeed service with an `ANSIBLE_AI_PROJECT_NAME` that includes the WCA suffix is updated to have the WCA suffix removed (so that the application code begins adding it dynamically).

## Testing
1. Pull down the PR
2. Make sure that ANSIBLE_AI_MODEL_MESH_CONFIG has a pipeline in it (like ModelPipelineCompletions) with a wca provider 
3. Set ANSIBLE_AI_PROJECT_NAME to something arbitrary
4. Start the service, navigate to the login page and you should see " with IBM watsonx Code Assistant" appended to the end of the project name

You can also remove the wca pipeline from your config and the suffix should go away

You can also navigate straight the chatbot endpoint (/chatbot) unauthenticated and the suffix should not be present on the login screen

## Production deployment


goneri:
In the future, it could be nice to have a commercial name field in the provider configuration, and we would just inject that in the final string.

```
  "ModelPipelineRoleExplanation": {
    "provider": "wca",
    "config": {
      "inference_url": "https://api.dataplatform.test.cloud.ibm.com",
      "commercial_name": "WatsonX by IBM"
    }
  }
```

PR #1735, opened 2025-08-06
Remove redundant granite3-1-8b that is unsupported with lightspeed-stack

manstis:

## Description
`granite3-1-8b` is not supported by our `ansible-chatbot-stack` backend and can be removed.

## Testing
N/A

### Steps to test
N/A

### Scenarios tested
N/A

## Production deployment


ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -33,7 +33,6 @@ const botName =
   ANSIBLE_LIGHTSPEED_PRODUCT_NAME;
 
 export const modelsSupported: LLMModel[] = [
-  { model: "granite3-1-8b", provider: "my_rhoai_g31" },
```
manstis, line 36: Our _local_ Wisdom Service Chatbot UI can run in "debug mode" but we only support one model with `lightspeed-stack`. There is a [JIRA](https://issues.redhat.com/browse/AAP-44959) to support multiple, like `road-core`, but it's not been worked on.

TamiTakamiya: I configure Granite (w/ vllm provider) and two Gemini models (2.5-pro and 2.5-flash) in my run.yaml and can switch between them using my locally built chatbot debug UI. I guess it;s working now.

goneri:
Note for the future: We also need to do a similar change in `ansible-wisdom-ops`: https://github.com/ansible/ansible-wisdom-ops/pull/1478

PR #1737, opened 2025-08-08
Support bypass tools option on lightspeed-stack /streaming_query API

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-51359>

Assisted-by: n/a
Generated by: n/a

## Description

Support the "no_tools" option on lightspeed-stack's `/streaming_query` and `/query` APIs.

This option is available when the chatbot UI is running in the DEBUG mode.  The new "Bypass tools" checkbox
is added on the UI for setting System Prompt Override.

image

### Steps to test
1. Pull down the PR
2. Run Unit tests

### Scenarios tested

Unit tests + manual tests using local server

## Production deployment

### DIFF
--- a/aap_chatbot/src/useChatbot/useChatbot.ts
+++ b/aap_chatbot/src/useChatbot/useChatbot.ts
@@ -33,7 +33,9 @@ const botName =
   ANSIBLE_LIGHTSPEED_PRODUCT_NAME;
 
 export const modelsSupported: LLMModel[] = [
-  { model: "granite-3.3-8b-instruct", provider: "rhoai" },
+  { model: "granite-3.3-8b-instruct", provider: "my_rhoai_dev" },
+  { model: "gemini/gemini-2.5-flash", provider: "gemini" },
+  { model: "gemini/gemini-2.5-pro", provider: "gemini" },
 ];
 
 export const readCookie = (name: string): string | null => {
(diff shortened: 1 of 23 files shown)


aap_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -33,7 +33,9 @@ const botName =
   ANSIBLE_LIGHTSPEED_PRODUCT_NAME;
 
 export const modelsSupported: LLMModel[] = [
-  { model: "granite-3.3-8b-instruct", provider: "rhoai" },
+  { model: "granite-3.3-8b-instruct", provider: "my_rhoai_dev" },
+  { model: "gemini/gemini-2.5-flash", provider: "gemini" },
```
manstis, line 37: Whilst these other models will work with presumably your AI Installer backend they will break LSIA. Are you planning on updating the SaaS backend to support these too?

manstis (approved):
LGTM 👍

aap_chatbot/src/useChatbot/useChatbot.ts
TamiTakamiya: Ok. I will revert the line 36.

I thought it needed to be consistent with our (road-core based) chatbot configuration ([this line](https://github.com/ansible/ansible-wisdom-ops/blob/bebe922fabcb325b88b6a172b85207ab59a81da4/applications/wisdom-chatbot/deployment.yaml#L16)), and that's the reason that I changed it to `my_rhoai_dev`, but the operator (this line) assumes `rhoai` as the provider name for `granite-3.3-8b-instruct`, I think I need to modify https://github.com/ansible/ansible-wisdom-ops/pull/1483 to change the provider name.

TamiTakamiya: (additional update) ... Though It is possible to have use provider names in `/aap_chatbot` and `/ansible_ai_connect_chatbot`, I want make these code as common as possible...

manstis: @TamiTakamiya Operator deployments are different to what we have in `ansible-wisdom--ops`. It was the Gemini models that'd cause problems as the SaaS backend only supports Granite in `llama-stack`s run configuration.

PR #1743, opened 2025-08-14
Containerfile: use ubi-minimal

goneri:
Use `ubi9/ubi-minimal` to be closed to the downstream set-up.


jameswnl:
PR Deployment created in namespace wisdom-pr-1743

PR #1746, opened 2025-08-14
Stop sending chat_response to telemetry

TamiTakamiya:
Jira Issue: <https://issues.redhat.com/browse/AAP-51559>

Assisted-by: n/a
Generated by: n/a

## Description

AI Installer will be deployed in PROD by reusing the PROD chatbot.  Because the complexity of the data processing AI Installer does, we want to enable streaming to provide better user experience.  Streaming has been disabled on the PROD chatbot.

Streamning chatbot sends a different event (`streamingChatOperationalEvent`) and we want to suppress sending responses (messages from chatbot) to telemetry as those responses may contain some sentive data like IP addresses or passwords.

This PR takes out `chat_response` from `streamingChatOperationalEvent` and `chatOperationalEvent`. ~~Also this PR remove the "seated user" check before filtering out sensitive properties. The "seated user" concept is no longer used and we should consider all users are "seated" now. It will ensure `chat_respponse` is taken out regardles of the value of the "seated_user" flag.~~ 

Update:
Removing the "seated user" check broke more test cases than I thought. I will open a separate Jira issue for that and limit the scope of this PR to just removing `chat_response` from `streamingChatOperationalEvent` and `chatOperationalEvent`

### Steps to test
1. Pull down the PR
2. Run Unit test

## Production deployment

### DIFF
--- a/ansible_ai_connect/main/templates/chatbot/index.html
+++ b/ansible_ai_connect/main/templates/chatbot/index.html
@@ -9,8 +9,8 @@
       content="{{bot_name}}"
     />
     <title>{{bot_name}}</title>
-    <script type="module" crossorigin src="/static/chatbot/index-D9V6sBiE.js"></script>
-    <link rel="stylesheet" crossorigin href="/static/chatbot/index-DCfh_ZDh.css">
+    <script type="module" crossorigin src="/static/chatbot/index-ecUyj60O.js"></script>
+    <link rel="stylesheet" crossorigin href="/static/chatbot/index-Dm-_Ojks.css">
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>


ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -261,6 +261,9 @@ def send_schema1_event(self, ev):
         # the AppRegistryNotReady exception
         from ansible_ai_connect.ai.api.utils.segment import send_schema1_event
 
+        # Set the "seated user" flag to ensure prompt/response are not sent to telemetry
+        ev.rh_user_has_seat = True
```
goneri, line 265: This is not super elegant :-). I would really prefer we completely remove the prompt/repsonse from the original payload. Is this an option?

PR #1747, opened 2025-08-15
chatbot: only keep gemini/gemini-2.5-flash for now

goneri:


omaciel (approved):
As long as we don't break anything when promoting to prod accidentally :)

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -33,7 +33,6 @@ const botName =
   ANSIBLE_LIGHTSPEED_PRODUCT_NAME;
 
 export const modelsSupported: LLMModel[] = [
-  { model: "granite-3.3-8b-instruct", provider: "my_rhoai_dev" },
   { model: "gemini/gemini-2.5-flash", provider: "gemini" },
```
omaciel, line 36: Maybe swap the order so that gemini is set by default just to avoid losing the configuration?

goneri:
> @goneri line 181 define the default model and it also needs to be updated

Done, thank you @TamiTakamiya

TamiTakamiya (approved):

PR #1753, opened 2025-08-28
Adding UI support for Vertex AI API.

romartin:
Just adding Vertex AI API model in the UI dropdown for chatbot.


jameswnl (approved):

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.ts
```diff
@@ -33,6 +33,7 @@ const botName =
   ANSIBLE_LIGHTSPEED_PRODUCT_NAME;
 
 export const modelsSupported: LLMModel[] = [
+  { model: "gemini-2.5-flash", provider: "gemini" },
```
manstis, line 36: How is this different to L#37?

`llama-stack` wants `provider_id`\`model_id` as the model name. You might see warnings in `ansible-chatbot-stack` logs using this. It'll become an exception in newer `llama-stack` releases.

PR #1754, opened 2025-09-02
To introduce env based cert discovery

justjais:
Jira Issue: <https://issues.redhat.com/browse/AAP-50667>

Test Assisted-by: Cursor IDE
Test Generated by: Sonnet

## Description

To introduce environment based cert discovery and also eliminates conditional SSL logic complexity that could be reason of failure when `ca_cert_file` is an empty string or invalid path.

### Steps to test
1. Pull down the PR
2. Build the container, push the image to quay
3. Use the image to test the operator for communication b/w service and chatbot, and it does communicates using https and port 8321

### Scenarios tested

NA

## Production deployment

- [x] This code change is ready for production on its own
- [ ] This code change requires the following considerations before going to production:

PR #1756, opened 2025-09-03
To fix the SSL issue for TLS enablement b/w service and chatbot

justjais:
Jira Issue: <https://issues.redhat.com/browse/AAP-50667>

Test Assisted-by: Cursor
Test Generated by: Claude

## Description

To fix the WCA SSL issue observed after merging of PR #1754. PFA the logs attached [here](https://paste.opendev.org/show/bk1igT0hrhxfZ9G97EBl/)

### Steps to test
1. Pull down the PR
2. Tested the fix in on-prem test cluster
3. WCA health status is Ok now, previous to the fix was resulting into error state

### Scenarios tested

Added unit tests for the scenario, to be caught early and not create regression.

## Production deployment


ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -78,26 +78,36 @@ def task_gen_timeout(self, task_count=1):
...
-                os.environ.setdefault("SSL_CERT_FILE", service_ca)
-                logger.info("Configured SSL context to use mounted service-ca certificate")
+                # Store the service CA path for session-level use
+                # instead of setting global environment variables
+                self._service_ca_path = service_ca
+                logger.info(
+                    "Found mounted service-ca certificate for HTTP pipeline SSL verification"
+                )
+            else:
+                self._service_ca_path = None
```
manstis, line 96: Do you need to set it to `None`?

You seem to check if `self` has attribute `_service_ca_path` elsewhere.

For example [here](https://github.com/ansible/ansible-ai-connect-service/pull/1756/files#diff-b6c900f30e6213219c6ddcd3af4d9405581fee6029a05e493bdd4d9613268578R105).

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
```diff
@@ -78,26 +78,36 @@ def task_gen_timeout(self, task_count=1):
...
+        else:
+            self._service_ca_path = None
```
manstis, line 98: Likewise.. you check `hasattr(self, "_service_ca_path")` elsewhere.

manstis (commented):
@justjais I'm still not clear on _why_ the WCA changes are needed.

From your PR I _guess_ that the TLS configuration in the `http` pipelines caused conflict with the TLS configuration in the `wca` pipelines and your workaround (particularly in `pipelines_base.py`) is to force using the correct configuration for the correct target endpoint. 

Before approving I'd like:

- To better understand the issue faced and why you've taken the approach in this PR
- Appreciate reviews from other team members who may know TLS much better than me (@goneri , @ldjebran?, @jameswnl?). 

I simply don't know enough about the issue that is being faced; nor enough about _how_ TLS should be configured to feel confident in approving this PR "as is".

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
justjais: Yes, I've set to None as it directs that "no service CA path available". 

And, for:
 `if self.config.verify_ssl and hasattr(self, "_service_ca_path") and self._service_ca_path:`

I can update to:
 `if self.config.verify_ssl and self._service_ca_path:`
to remove the redundant comparison

ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
justjais: I'll remove the `hasattr(self, "_service_ca_path")` comparison to make it consistent and non-redundant.

goneri (approved):

goneri (changes requested):
I approved the PR, but I haven't had the time to test it myself. I would appreciate another +1 before merging.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,181 @@
...
+                )
+                # Don't raise exception, just fall back to system defaults
+                # session.verify defaults to True
+        else:
+            session.verify = False
+            logger.debug("SSL Manager: SSL verification disabled")
+
+        return session
+
+    def get_ssl_context(self, verify_ssl: bool = True) -> Optional[ssl.SSLContext]:
```
goneri, line 133: I don't see the value of the `verify_ssl` parameter considering you only test the value once and don't really use it. It would be more elegant to just remove the parameter and do the `verify_ssl` check before.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,181 @@
...
+                            + " falling back to system defaults",
+                            ca_bundle_path,
+                        )
+                        # session.verify defaults to True, which uses system CAs
+                else:
+                    logger.debug("SSL Manager: Session using system default CA verification")
+                    # session.verify defaults to True, which uses system CAs
+
+            except (OSError, AttributeError) as e:
+                logger.warning(
```
goneri, line 122: This should IMO by fatal. If the TLS configuraiton is wrong, we should not try to make it works.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_requests_session(self, verify_ssl: bool = True) -> requests.Session:
+        """Get requests session with infrastructure-managed SSL configuration.
+        Args:
+            verify_ssl: Whether SSL verification should be enabled
+        Returns:
+            Configured requests session with proper SSL settings.
+            Falls back to system defaults if custom CA bundles are not accessible.
+        """
+        session = requests.Session()
+
```
goneri, line 99: Moving the following block at the top of the function will reduce the complexity, like you did here https://github.com/ansible/ansible-ai-connect-service/pull/1756/files#diff-5383513b0ec56a18282de119eaa5dfb5eda4512a099be14316686c5f365989eaR142:

```python
if not verify_ssl:
    session.verify = False
    logger.info("SSL Manager: SSL verification disabled")
    return session
```

This way you can remove one level of indentation for the rest of the logic.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,181 @@
...
+            except (OSError, AttributeError) as e:
+                logger.warning(
+                    "SSL Manager: SSL configuration issue (non-fatal): %s, using system defaults", e
+                )
+                # Don't raise exception, just fall back to system defaults
+                # session.verify defaults to True
+        else:
+            session.verify = False
+            logger.debug("SSL Manager: SSL verification disabled")
```
goneri, line 129: This should IMO be a `logger.warning()`. This is not something we want to encourage.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,190 @@
...
+                        + " falling back to system defaults",
+                        ca_bundle_path,
+                    )
+            else:
+                # session.verify defaults to True, which thus uses system CAs
+                logger.debug("SSL Manager: Session using system default CA verification")
+
+        except (OSError, AttributeError) as e:
+            # session.verify defaults to True
+            logger.warning("SSL Manager: SSL configuration issue: %s, using system defaults", e)
```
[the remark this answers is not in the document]
justjais: done, thanks!

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+        except (OSError, AttributeError) as e:
+            # SSL configuration errors should be fatal
+            error_msg = f"SSL Manager: Fatal SSL configuration error: {e}"
+            logger.error(error_msg)
+            raise OSError(error_msg) from e
+
+        return session
+
+    def get_ssl_context(self, verify_ssl: bool = True) -> Optional[ssl.SSLContext]:
```
goneri, line 148: verify_ssl is always True, please remove the parameter.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,220 @@
...
+        """Get SSL context for aiohttp with infrastructure-managed certificates.
+        Args:
+            verify_ssl: Whether SSL verification should be enabled
+        Returns:
+            SSLContext configured with CA bundle, or None if verification disabled
+        Raises:
+            ssl.SSLError: If SSL context creation fails
+            OSError: If SSL configuration fails
+        """
+        if not verify_ssl:
```
goneri, line 158: Please remove the block

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,180 @@
...
+        legacy_service_ca = self._get_legacy_service_ca_path()
+        if legacy_service_ca and os.path.exists(legacy_service_ca):
+            logger.debug("SSL Manager: Using legacy service CA: %s", legacy_service_ca)
+            return legacy_service_ca
+
+        # No custom CA bundle found - use system defaults
+        logger.debug("SSL Manager: No custom CA bundle found, using system defaults")
+        return None
+
+    def get_requests_session(self) -> requests.Session:
```
goneri, line 98: We can actually prepare a session that would reuse `get_ssl_context()`. I've prepared an example here for verify=False:

https://goneri.lebouder.net/allow-broken-certificate-for-a-specific-host-with-pythons-requests/

We can easily reuse the logic to prepare another Adapter for the other cases.

ansible_ai_connect/main/ssl_manager.py
```diff
@@ -0,0 +1,190 @@
...
+            else:
+                # session.verify defaults to True, which thus uses system CAs
+                logger.debug("SSL Manager: Session using system default CA verification")
+
+        except (OSError, AttributeError) as e:
+            # session.verify defaults to True
+            logger.warning("SSL Manager: SSL configuration issue: %s, using system defaults", e)
+
+        return session
+
```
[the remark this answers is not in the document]
goneri: yup :-)

goneri (changes requested):

goneri (approved):

PR #1771, opened 2025-09-25
AAP-54064 clear RBAC data in migration step

AlanCoding:
This is split from https://github.com/ansible/ansible-ai-connect-service/pull/1770 to be the minimum changes to facilitate the fix.

Jira Issue: <https://issues.redhat.com/browse/AAP-54064>

Assisted-by: Claude (kind of, it wasn't that much help for this)

## Description
Without this, we can hit an upgrade error for the 2.5->2.6 line. This is because of the DAB 2.5 behavior of auto-registering the `Team` model. No models are intended to be registered, and you can see this in the setting in the `base.py` settings module. So there is no hazard in delete all rows in RBAC, as this does. RBAC is intended to be _enabled_ and also _unused_ and we will continue with that expectation.

## Testing
TBD - do upgrade testing

### Steps to test
Do full deployments in OCP, I don't have a local or minimal version of the fix. Although it would be possible.

### Scenarios tested
TBD

## Production deployment

Needs testing in production.

### DIFF
--- a/ansible_ai_connect/organizations/migrations/0005_migration_data_fix.py
+++ b/ansible_ai_connect/organizations/migrations/0005_migration_data_fix.py
@@ -1,0 +1,41 @@
+# Generated manually to create new Organization model and clear dab_rbac data
+
+from django.db import migrations
+
+
+def clear_dab_rbac_data(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


AlanCoding:
Turned out this can't work.

PR #2039, opened 2026-05-18
AAP-74211 / CVE-2026-44188: revoke user's OAuth tokens on logout

omaciel:
## Summary

- Hooks `user_logged_out` so the user's `AccessToken` rows are deleted and live `RefreshToken` rows are revoked at logout time, closing **CVE-2026-44188** (tracker **AAP-74211**).
- Without this, a stolen-but-still-valid OAuth bearer token survives logout for the full TTL — 10h access / 14h refresh — and lets an attacker keep calling `/api/v1/wca/apikey/`, `/api/v0/ai/completions/`, etc.
- `LogoutView` is **unchanged**; redirect semantics for OIDC/AAP/GH are preserved (existing `LogoutTest` continues to pass).
- Also bumps `ujson` 5.12.0 → 5.12.1 (**CVE-2026-44660**, ujson memory leak on failing `write()`). This is required to unblock the `pip-audit` selftest gate on this branch — `ujson==5.12.0` on `main` is flagged on every PR until the pin moves.

## Why a signal handler (not a `LogoutView.dispatch` override)

Django 4.2's `LogoutView.post` calls `auth.logout(request)` which fires `user_logged_out` while `request.user` is still attached, so the handler has everything it needs. All other auth-signal handling already lives in `users/signals.py`, so this keeps logout side effects in one place.

## Handler behavior

- No-op for `None` / anonymous users (`is_authenticated` guard).
- Iterates live refresh tokens and calls `rt.revoke()` — django-oauth-toolkit's `revoke()` deletes the linked access token and stamps `revoked=now()` on the refresh token, so a bulk `update()` would skip the cascade.
- Then deletes any remaining `AccessToken` rows for the user (covers tokens minted without a refresh token, e.g. via `wisdom-manage createtoken`).
- Errors propagate — better a noisy 500 on logout than a silent failure that leaves the very tokens this CVE is about still valid.

## Relationship to AAP-68128

AAP-68128 (closed) deployed the `revoke_expired_refreshtokens` and `cleartokens` crons. Those only touch tokens whose TTL has **already passed** (`created__lt=now - REFRESH_TOKEN_EXPIRE_SECONDS`), so they do nothing about a stolen-but-still-fresh token. The cron remains useful as a complementary hygiene job — once tokens are revoked at logout, `cleartokens` is what eventually deletes the revoked rows from the DB.

## Out of scope (recommended follow-up)

Lowering `ACCESS_TOKEN_EXPIRE_SECONDS` from the django-oauth-toolkit default of 36000s to something like 3600s. That's a config-only mitigation that should land as a separate PR so the security fix here can ship without bikeshedding the TTL value.

## ujson bump details

- `pyproject.toml`: constraint `ujson>=5.12.0` → `ujson>=5.12.1`.
- `requirements.txt`: regenerated with `uv export --format requirements-txt --no-hashes --no-emit-project -o requirements.txt`. Only delta is `ujson==5.12.0` → `ujson==5.12.1`.
- `uv.lock`: refreshed via `uv lock --upgrade-package ujson`. Constraint line and package entry both updated; new sha256 + wheel URLs for 5.12.1.
- Upstream fix: ultrajson/ultrajson@82af1d0a, released as [5.12.1](https://github.com/ultrajson/ultrajson/releases/tag/5.12.1).

## Test plan

- [x] New `LogoutTokenRevocationTest` (5 cases): happy path, cross-user isolation, no-tokens no-op, anonymous-logout no-op, access-token-without-refresh-token branch.
- [x] Existing `LogoutTest` (4 cases) still passes — redirect behavior unchanged.
- [x] Adjacent suites pass: `test_revoke_expired_refreshtokens`, `test_pipeline`, `test_auth` (25 tests).
- [ ] CI run on this PR (pre-commit, selftest/pip-audit, full unit-test job, Konflux).
- [ ] Optional manual smoke: `wisdom-manage createtoken --username foo --create-user --password foo --token-name probe`, log in as `foo` in a browser, `curl -H 'Authorization: Bearer probe' http://localhost:8000/api/v1/me/` → 200; POST `/logout/`; same curl → 401.

```
wisdom-manage test ansible_ai_connect.main.tests.test_views.LogoutTokenRevocationTest
wisdom-manage test ansible_ai_connect.main.tests.test_views.LogoutTest
```


ansible_ai_connect/users/signals.py
```diff
@@ -52,6 +53,29 @@ def user_logout_log(sender, user, **kwargs):
...
+
+    # Refresh tokens first: rt.revoke() in django-oauth-toolkit also
+    # revokes (deletes) the linked AccessToken, so handle that cascade
+    # before deleting any orphan access tokens below.
+    for rt in RefreshToken.objects.filter(user=user, revoked__isnull=True):
+        rt.revoke()
+
+    # Catch access tokens that were never paired with a refresh token,
+    # e.g. tokens minted by `wisdom-manage createtoken`.
+    AccessToken.objects.filter(user=user).delete()
```
ldjebran: is this going to delete all the user tokens or only the oauth related tokens ?
note: wisdom service has tokens that are not related to oauth.

ansible_ai_connect/main/tests/test_views.py
```diff
@@ -79,6 +84,100 @@ def test_logout_without_login(self):
         self.client.post(reverse("logout"))
 
 
+class LogoutTokenRevocationTest(TestCase):
```
ldjebran, line 87: can we add a test to make sure user tokens are not deleted ?

ansible_ai_connect/users/signals.py
omaciel: Good catch — confirmed: `createtoken` creates rows with `application=NULL` (createtoken.py:99-104), and the FK is nullable. The previous fallback would have deleted those non-OAuth admin tokens.

Refined in b1c37eae: the fallback delete now filters `application__isnull=False`, so only OAuth-issued tokens (rows bound to an Application) are revoked. The first-pass `rt.revoke()` loop is already inherently OAuth-only — refresh tokens always have an Application.

Scope per CVE-2026-44188 is OAuth session tokens. Admin tokens minted via `wisdom-manage createtoken` survive logout untouched.

ansible_ai_connect/main/tests/test_views.py
omaciel: Added in b1c37eae as `test_logout_does_not_delete_non_oauth_tokens`: creates an `application=None` admin token plus a companion OAuth token for the same user, logs out, asserts the admin token survives while the OAuth token is gone.

Also retitled the comment on `test_logout_revokes_access_token_without_refresh_token` — it was previously labelled as a `createtoken` case, but the test actually covers an OAuth token issued without a refresh token (e.g. client_credentials), which is correct. The new test is the createtoken-style coverage.

ldjebran (approved):
LGTM, Thank you @omaciel

jameswnl (approved):

jameswnl (commented):
**Nice to have / future improvement:** Consider hard-deleting tokens from the DB instead of soft-revoking.

Currently `rt.revoke()` stamps a `revoked` timestamp on RefreshTokens but keeps the rows. For a CVE fix, removing all token material from the DB is arguably a stronger security posture. The loop + second-pass delete could be replaced with bulk deletes:

```python
RefreshToken.objects.filter(user=user, application__isnull=False).delete()
AccessToken.objects.filter(user=user, application__isnull=False).delete()
```

This would also simplify the code (no N+1 loop, no overlapping deletes). The `application__isnull=False` filter still preserves non-OAuth admin tokens, and the log line already provides the audit trail.

Not a blocker for this PR -- just something to consider for a follow-up.

omaciel:
@goneri mind also taking a peek?

hasys (approved):
Looks good to me, thank you @omaciel

goneri (approved):

PR #2045, opened 2026-05-22
Harden cookie security: __Host- prefix, Secure flag, CSRF rotation

ldjebran:
Jira Issue: https://redhat.atlassian.net/browse/AAP-61711

Assisted-by: Claude Code

## Description

Harden cookie security: __Host- prefix, Secure flag, CSRF middleware
====================================================================
Changes
1. Cookie hardening settings (base.py)
--------------------------------------
- __Host- prefix on session and CSRF cookies (__Host-sessionid, __Host-csrftoken)
  — prevents subdomain cookie tossing
- Secure flag on both cookies — HTTPS-only transmission
- SameSite=Lax — allows OAuth/OIDC redirects while blocking cross-site POST
- CSRF_COOKIE_AGE reduced from Django's 1-year default to 2 weeks, aligned
  with SESSION_COOKIE_AGE
- CSRF_COOKIE_HTTPONLY=False — JS must read the cookie for X-CSRFToken headers
- Gateway cookie settings — GATEWAY_SESSION_COOKIE_NAME and
  GATEWAY_CSRF_COOKIE_NAME (env-configurable) for detecting gateway-proxied
  requests

2. Development overrides (development.py)
-----------------------------------------
- Removed CSRF_USE_SESSIONS=True (was a workaround for gateway stripping
  cookies)
- Added CSRF_COOKIE_SECURE=False and CSRF_COOKIE_NAME="csrftoken" — plain HTTP
  dev servers can't use __Host- prefix (requires Secure)
- Same for session: SESSION_COOKIE_SECURE=False, SESSION_COOKIE_NAME="sessionid"

3. EnsureCsrfCookieMiddleware (middleware.py)
---------------------------------------------
New middleware placed before CsrfViewMiddleware. Handles two deployment modes:

Standalone mode — 4 cases for cross-origin proxy / OAuth scenarios:
  1. Browser sent CSRF cookie — reuse existing secret
  2. No cookie but X-CSRFToken header — use header value (OAuth via gateway,
     cookie domain mismatch)
  3. No cookie, no header, POST with csrfmiddlewaretoken — inject form token
     (logout form)
  4. None of the above — generate fresh secret

Gateway mode — aliases the gateway's csrftoken cookie into Django's
CSRF_COOKIE_NAME so Django's CSRF validation passes

4. Console templates (console.html, denied.html)
-------------------------------------------------
- Added <div id="csrf_token" hidden>{{ csrf_token }}</div> — Django embeds the
  CSRF token in the DOM so JS can read it when cookies aren't available
  (cross-origin proxy)

5. Admin portal (ansible_ai_connect_admin_portal)
-------------------------------------------------
- readCsrfCookie() — 3-tier fallback: __Host-csrftoken cookie -> DOM csrf_token
  element -> csrftoken cookie
- X-CSRFToken header — changed from { "X-CSRFToken": csrfToken } (sends null)
  to csrfToken ? { ... } : {} (omits header entirely when no token)
- webpack dev server — migrated from deprecated
  onBeforeSetupMiddleware/onAfterSetupMiddleware to setupMiddlewares, and https
  to server option (webpack-dev-server v4+)

6. Standalone chatbot (ansible_ai_connect_chatbot)
--------------------------------------------------
- readCsrfCookie() — __Host-csrftoken with csrftoken fallback
- X-CSRFToken header — same csrfToken ? { ... } : {} pattern
- chatbot/index.html asset hash change is auto-generated by npm run build
  in the ansible_ai_connect_chatbot directory

7. Gateway chatbot (aap_chatbot)
--------------------------------
- isGatewayRequest() — mirrors middleware's gateway detection (gateway session
  cookie present, Django session cookie absent)
- readCsrfCookie() — behind gateway uses csrftoken; standalone uses
  __Host-csrftoken with csrftoken fallback
- X-CSRFToken header — same pattern

8. Tests
--------
Python (test_cookie_security.py — 35 tests, all new):
  - 14 base cookie settings + __Host- invariants
  - 4 development overrides
  - 2 gateway settings
  - 9 EnsureCsrfCookieMiddleware tests (all 4 code paths + precedence + edge
    cases)
  - 2 middleware stack position
  - 4 gateway middleware tests

TypeScript (across 3 test files):
  - Admin portal: cookie spy mock, __Host- extraction, partial name rejection,
    DOM fallback, null case, headers: {} for null token
  - Standalone chatbot: cookie spy mock, __Host- present, csrftoken fallback,
    preference, null
  - Gateway chatbot: all above + gateway detection (uses csrftoken),
    both-sessions (uses __Host-), null

9. API schema unchanged
-----------------------
No changes to OpenAPI/schema files. The API schema is only generated in
development mode (SPECTACULAR_SETTINGS in development.py), so the cookie
hardening settings in base.py do not affect schema generation.

10. Security audit
==================

Strong protections
------------------
- __Host- prefix — strongest cookie binding available. Prevents subdomain
  cookie tossing, forces Secure, forces Path=/. No bypass possible.

- SameSite=Lax — blocks cross-site POST requests from sending cookies. This is
  the primary CSRF defense in modern browsers. Lax (not Strict) is correct
  because Strict would break OAuth/OIDC redirect flows.

- Secure flag — cookies never transmitted over HTTP. Combined with __Host-,
  can't be downgraded.

- HttpOnly on session cookie — JS can't steal session tokens via XSS.

- CSRF_COOKIE_AGE reduced to 2 weeks — limits the window for token fixation.

- CSRF token rotation on login — Django calls rotate_token() in login() by
  default, invalidating pre-authentication tokens.

Acceptable trade-offs
---------------------
- CSRF_COOKIE_HTTPONLY=False — required for the double-submit pattern (JS reads
  cookie, echoes in X-CSRFToken header). An XSS attacker could read it, but XSS
  already allows making requests directly, so HttpOnly on the CSRF cookie adds
  no real protection.

- DOM-embedded {{ csrf_token }} — same reasoning. XSS makes CSRF protection
  moot regardless.

- Dev mode relaxation — removes __Host- and Secure for HTTP dev servers.
  Acceptable as long as production always uses base.py settings.

One area to flag
----------------
Case 3 (form token injection) — when no cookie and no header exist, we inject
the POST body's csrfmiddlewaretoken into request.COOKIES and let
CsrfViewMiddleware validate it. This effectively makes the cookie-side match the
form-side, weakening the double-submit pattern to a single-submit check.

Could an attacker exploit this? A cross-site attacker would craft a form POST
with a known csrfmiddlewaretoken. However:

  - SameSite=Lax blocks the session cookie on cross-site POST — user not
    authenticated — request fails before CSRF matters
  - Django's CsrfViewMiddleware also checks the Referer header on HTTPS
  - The form token in legitimate use was generated by {% csrf_token %} and
    embedded in an authenticated page the attacker can't read

Case 3 relies on SameSite=Lax as the primary defense rather than double-submit.
This is a slight weakening of defense-in-depth but not exploitable in practice.

Verdict
-------
The implementation is secure. The layered defenses (__Host- + Secure +
SameSite=Lax + CSRF double-submit + token rotation) are industry best practice.
The one trade-off (case 3) is mitigated by SameSite=Lax blocking the session
cookie on cross-site POST, making the CSRF check unreachable by an attacker.

### Steps to test
1. Pull down the PR
2. deploy AAP locally with lightspeed integration, with lightspeed configured with chatbot. 
3. The default scenario is with lightspeed running http 
4. login to AAP and try to make some requests to chatbot 
5. open localhost:7080 login with oauth 
6. open localhost:7080/chatbot/ ensure you can use the chatbot , put some question
7. go back to localhost:7080/ and logout 
8. open localhost:7080/console, you will be redirected to login , login
9. the console is displayed , logout 
10. no error page should be displayed

testing with aap-ui

1. cd to aap_chatboot and run npm run build and npm pack
git clone aap-ui
in package.json change ansible-ai-connect-chatbot to 
```"@ansible/ansible-ai-connect-chatbot": "file:/Users/djlezzou/projects/ansible/lightspeed/ansible-ai-connect-service/aap_chatbot/package",
2. run npm ci
3. cd platform
4. export PLATFORM_SERVER=https://localhost
5. run npm start
6. the browser open to the served ui , login make sure the chat bot is functional


for https case comment lines 86-89 and restart the lightspeed
repeat all the above tests you can ensure that the right cookie and sessionid are created with names __Host-csrftoken and __Host-sessionid
 
 

## Type of Change
- [ ] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality)
- [ ] Breaking change (fix or feature causing existing functionality to break)
- [X] Security fix
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] CI/CD update

## Backport Policy

This change should be:
- [ ] **Not backported** - main/master only
- [X] **Backported** to specific releases (add labels after merge)

### Automated Backport Instructions

**After this PR is merged**, add one or more labels to automatically create backport PRs:

- `backport/stable-2.4` - Backport to stable-2.4 branch
- `backport/stable-2.5` - Backport to stable-2.5 branch
- `backport/stable-2.6` - Backport to stable-2.6 branch
- `backport/all` - Backport to all active stable branches
- `no-backport` - Explicitly mark as not needing backport

### Backport Justification
<!-- Required if backporting. Explain why this needs to be in older releases -->
<!-- Examples: -->
<!-- - Critical bug affecting production -->
<!-- - Security vulnerability fix (include CVE if applicable) -->
<!-- - Regression introduced in X.Y.Z -->
<!-- - Customer-blocking issue -->
<!-- - Data corruption/loss prevention -->

**Special backport considerations:**
<!-- Any conflicts, dependencies, or modifications needed for backport -->
<!-- Will this apply cleanly to older releases? -->
<!-- Does this depend on other commits that also need backporting? -->

### Scenarios tested
<!-- Describe the scenarios you've already manually verified, if applicable. -->

## Production deployment
<!-- Check the appropriate box. Document any pre-reqs, co-reqs, secrets, configmaps, etc that need to be considered or prepared ahead of a deployment to production. Include links to any related PRs, e.g. in ansible-wisdom-ops. -->

### DIFF
--- a/aap_chatbot/src/useChatbot/useChatbot.test.ts
+++ b/aap_chatbot/src/useChatbot/useChatbot.test.ts
@@ -1,6 +1,11 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { afterEach, describe, it, expect, vi, beforeEach } from "vitest";
 import { renderHook, waitFor } from "@testing-library/react";
-import { feedbackMessage, useChatbot } from "./useChatbot";
+import {
+  feedbackMessage,
+  readCookie,
+  readCsrfCookie,
+  useChatbot,
+} from "./useChatbot";
 import type { MessageProps } from "@patternfly/chatbot/dist/dynamic/Message";
 import { Sentiment } from "../Constants";
 import type { ChatFeedback } from "../types/Message";
@@ -12,6 +17,90 @@ const CONVERSATION_ID = "123e4567-e89b-12d3-a456-426614174000";
 vi.mock("@microsoft/fetch-event-source", () => ({
   fetchEventSource: vi.fn(),
 }));
+
+const clearCookies = () => {
+  for (const cookie of document.cookie.split(";")) {
+    const name = cookie.split("=")[0].trim();
+    if (name) {
+      document.cookie = `${name}=;expires=Thu, 01 Jan 1970 00:00:00 GMT;path=/`;
+    }
+  }
+};
+
+describe("readCookie", () => {
+  beforeEach(clearCookies);
+  afterEach(clearCookies);
+
+  it("returns null when cookie is not present", () => {
+    expect(readCookie("csrftoken")).toBeNull();
+  });
+
+  it("extracts cookie value", () => {
+    document.cookie = "csrftoken=12345";
+    expect(readCookie("csrftoken")).toEqual("12345");
+  });
+
+  it("extracts correct cookie from multiple cookies", () => {
+    document.cookie = "other=abc";
+    document.cookie = "csrftoken=12345";
+    expect(readCookie("csrftoken")).toEqual("12345");
+  });
+
+  it("does not match partial cookie names", () => {
+    document.cookie = "csrftoken_extra=wrong";
+    expect(readCookie("csrftoken")).toBeNull();
+  });
+});
+
+// Browsers reject __Host- prefixed cookies on non-HTTPS pages, so we mock
+// document.cookie getter instead of setting real cookies.
+describe("readCsrfCookie", () => {
+  let cookieSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    cookieSpy = vi.spyOn(document, "cookie", "get");
+  });
+
+  afterEach(() => {
+    cookieSpy.mockRestore();
+  });
+
+  it("returns __Host-csrftoken when standalone", () => {
+    cookieSpy.mockReturnValue("__Host-csrftoken=prod_token");
+    expect(readCsrfCookie()).toEqual("prod_token");
+  });
+
+  it("falls back to csrftoken when __Host-csrftoken is absent", () => {
+    cookieSpy.mockReturnValue("csrftoken=dev_token");
+    expect(readCsrfCookie()).toEqual("dev_token");
+  });
+
+  it("prefers __Host-csrftoken over csrftoken without gateway", () => {
+    cookieSpy.mockReturnValue(
+      "csrftoken=dev_token; __Host-csrftoken=prod_token",
+    );
+    expect(readCsrfCookie()).toEqual("prod_token");
+  });
+
+  it("uses csrftoken when behind gateway", () => {
+    cookieSpy.mockReturnValue(
+      "csrftoken=gw_token; __Host-csrftoken=django_token; gateway_sessionid=abc123",
+    );
+    expect(readCsrfCookie()).toEqual("gw_token");
+  });
+
+  it("uses __Host-csrftoken when both gateway and django sessions exist", () => {
+    cookieSpy.mockReturnValue(
+      "csrftoken=gw_token; __Host-csrftoken=django_token; gateway_sessionid=abc123; __Host-sessionid=xyz789",
+    );
+    expect(readCsrfCookie()).toEqual("django_token");
+  });
+
+  it("returns null when neither cookie is present", () => {
+    cookieSpy.mockReturnValue("");
+    expect(readCsrfCookie()).toBeNull();
+  });
+});
 
 describe("feedbackMessage", () => {
   it("should return a message with a thank you note for positive feedback", () => {
--- a/ansible_ai_connect/main/settings/base.py
+++ b/ansible_ai_connect/main/settings/base.py
@@ -89,6 +89,7 @@ MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
+    "ansible_ai_connect.main.middleware.EnsureCsrfCookieMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "oauth2_provider.middleware.OAuth2TokenMiddleware",
@@ -104,6 +105,30 @@ if os.environ.get("CSRF_TRUSTED_ORIGINS"):
     CSRF_TRUSTED_ORIGINS = os.environ.get("CSRF_TRUSTED_ORIGINS").split(",")
 else:
     CSRF_TRUSTED_ORIGINS = ["http://localhost:8000"]
+
+# SameSite=Lax (not Strict) because OAuth/OIDC login flows redirect back from
+# external identity providers (Red Hat SSO, GitHub, AAP). Strict would drop the
+# session cookie on that redirect, making the user appear unauthenticated.
+SESSION_COOKIE_SECURE = True
+SESSION_COOKIE_HTTPONLY = True
+SESSION_COOKIE_SAMESITE = "Lax"
+SESSION_COOKIE_NAME = "__Host-sessionid"
+SESSION_COOKIE_AGE = 1209600  # 2 weeks, aligned with CSRF_COOKIE_AGE
+
+CSRF_COOKIE_SECURE = True
+# JS must read the CSRF cookie to echo it in the X-CSRFToken header
+CSRF_COOKIE_HTTPONLY = False
+# SameSite=Lax (not Strict) so the CSRF cookie survives OAuth/OIDC redirects
+# from external identity providers, matching the session cookie policy above.
+CSRF_COOKIE_SAMESITE = "Lax"
+CSRF_COOKIE_NAME = "__Host-csrftoken"
+CSRF_COOKIE_AGE = 1209600  # 2 weeks; reduced from Django's 1-year default
+
+# When behind AAP gateway, the gateway owns CSRF validation using its own
+# cookie. These settings let the middleware detect gateway-proxied requests
+# and alias the gateway's CSRF cookie so Django's validation passes.
+GATEWAY_SESSION_COOKIE_NAME = os.environ.get("GATEWAY_SESSION_COOKIE_NAME", "gateway_sessionid")
+GATEWAY_CSRF_COOKIE_NAME = os.environ.get("GATEWAY_CSRF_COOKIE_NAME", "csrftoken")
 
 # Allow Prometheus to scrape metrics
 ALLOWED_CIDR_NETS = [os.environ.get("ALLOWED_CIDR_NETS", "10.0.0.0/8")]
--- a/ansible_ai_connect/main/settings/development.py
+++ b/ansible_ai_connect/main/settings/development.py
@@ -83,6 +83,11 @@ if DEBUG:
             "ansible_ai_connect.main.middleware.WisdomSocialAuthExceptionMiddleware"  # noqa: F405
         )
 
+SESSION_COOKIE_SECURE = False
+SESSION_COOKIE_NAME = "sessionid"
+CSRF_COOKIE_SECURE = False
+CSRF_COOKIE_NAME = "csrftoken"
+
 CSP_REPORT_ONLY = True
 AUTHZ_BACKEND_TYPE = os.getenv("AUTHZ_BACKEND_TYPE") or "dummy"
 # e.g:
--- a/ansible_ai_connect/main/templates/chatbot/index.html
+++ b/ansible_ai_connect/main/templates/chatbot/index.html
@@ -9,8 +9,8 @@
       content="{{bot_name}}"
     />
     <title>{{bot_name}}</title>
-    <script type="module" crossorigin src="/static/chatbot/index-8zRWvsD-.js"></script>
-    <link rel="stylesheet" crossorigin href="/static/chatbot/index-CN-YAjNY.css">
+    <script type="module" crossorigin src="/static/chatbot/index-xKaMtcOq.js"></script>
+    <link rel="stylesheet" crossorigin href="/static/chatbot/index-BCBeJGFG.css">
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
--- a/ansible_ai_connect/users/signals.py
+++ b/ansible_ai_connect/users/signals.py
@@ -21,6 +21,7 @@ from django.contrib.auth.signals import (
     user_login_failed,
 )
 from django.dispatch import Signal, receiver
+from django.middleware.csrf import rotate_token
 from oauth2_provider.models import get_access_token_model, get_refresh_token_model
 
 logger = logging.getLogger(__name__)
@@ -33,9 +34,11 @@ user_set_telemetry_settings = Signal()
 
 
 @receiver(user_logged_in)
-def user_login_log(sender, user, **kwargs):
+def user_login_log(sender, user, request=None, **kwargs):
     """Successful user login log to user log"""
     logger.info(f"User: {user} LOGIN successful")
+    if request:
+        rotate_token(request)
 
 
 @receiver(user_login_failed)
--- a/ansible_ai_connect/users/tests/test_signals.py
+++ b/ansible_ai_connect/users/tests/test_signals.py
@@ -12,6 +12,10 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from unittest.mock import Mock, patch
+
+from django.contrib.auth.signals import user_logged_in
+
 from ansible_ai_connect.test_utils import WisdomServiceLogAwareTestCase
 from ansible_ai_connect.users.signals import _obfuscate
 
@@ -25,3 +29,19 @@ class TestSignals(WisdomServiceLogAwareTestCase):
         self.assertEqual("h****n", _obfuscate("hidden"))
         self.assertEqual("to******et", _obfuscate("top-secret"))
         self.assertEqual("a-lo***************alue", _obfuscate("a-long-top-secret-value"))
+
+
+class TestCSRFTokenRotationOnLogin(WisdomServiceLogAwareTestCase):
+
+    @patch("ansible_ai_connect.users.signals.rotate_token")
+    def test_csrf_token_rotated_on_login(self, mock_rotate_token):
+        request = Mock()
+        user = Mock()
+        user_logged_in.send(sender=self.__class__, user=user, request=request)
+        mock_rotate_token.assert_called_once_with(request)
+
+    @patch("ansible_ai_connect.users.signals.rotate_token")
+    def test_csrf_token_not_rotated_without_request(self, mock_rotate_token):
+        user = Mock()
+        user_logged_in.send(sender=self.__class__, user=user)
+        mock_rotate_token.assert_not_called()
--- a/ansible_ai_connect_admin_portal/src/api/__tests__/api.test.ts
+++ b/ansible_ai_connect_admin_portal/src/api/__tests__/api.test.ts
@@ -1,34 +1,87 @@
 import "@testing-library/jest-dom";
-import { readCookie } from "../api";
+import { readCookie, readCsrfCookie } from "../api";
 
+// Browsers reject __Host- prefixed cookies on non-HTTPS pages, so we mock
+// document.cookie getter instead of setting real cookies.
 describe("API", () => {
-  it("Cookie extraction::Empty string", () => {
-    document.cookie = "";
-    const cookie = readCookie("csrftoken");
-    expect(cookie).toBeNull();
+  let cookieSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    cookieSpy = jest.spyOn(document, "cookie", "get");
   });
 
-  it("Cookie extraction::Single::With whitespace", () => {
-    document.cookie = "csrftoken = 12345   ";
-    const cookie = readCookie("csrftoken");
-    expect(cookie).toEqual("12345");
+  afterEach(() => {
+    cookieSpy.mockRestore();
   });
 
-  it("Cookie extraction::Single::Without whitespace", () => {
-    document.cookie = "csrftoken=12345";
-    const cookie = readCookie("csrftoken");
-    expect(cookie).toEqual("12345");
+  describe("readCookie", () => {
+    it("Cookie extraction::Empty string", () => {
+      cookieSpy.mockReturnValue("");
+      const cookie = readCookie("__Host-csrftoken");
+      expect(cookie).toBeNull();
+    });
+
+    it("Cookie extraction::Single::With whitespace", () => {
+      cookieSpy.mockReturnValue("__Host-csrftoken=12345   ");
+      const cookie = readCookie("__Host-csrftoken");
+      expect(cookie).toEqual("12345");
+    });
+
+    it("Cookie extraction::Single::Without whitespace", () => {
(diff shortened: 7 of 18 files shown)

### CODE REVIEW

File: ansible_ai_connect/main/settings/base.py [hunk 1]
```diff
@@ -105,6 +105,20 @@
...
+# SameSite=Lax (not Strict) because OAuth/OIDC login flows redirect back from
+# external identity providers (Red Hat SSO, GitHub, AAP). Strict would drop the
+# session cookie on that redirect, making the user appear unauthenticated.
+SESSION_COOKIE_SECURE = True
+SESSION_COOKIE_HTTPONLY = True
+SESSION_COOKIE_SAMESITE = "Lax"
+SESSION_COOKIE_NAME = "__Host-sessionid"
+
+CSRF_COOKIE_SECURE = True
+CSRF_COOKIE_HTTPONLY = False
```
omaciel, line 120: Consider adding a one-line comment explaining why CSRF is `HTTPONLY=False` while session is `HTTPONLY=True`. On a casual read this looks like a security regression; the real reason is that the SPA frontends need to read it via JS to populate the `X-CSRFToken` header. Suggested:

```python
# CSRF must be readable by JS so the SPA can echo it in the X-CSRFToken header.
CSRF_COOKIE_HTTPONLY = False
```

This also documents intent against a future "harden everything" patch that would break the frontend.

ansible_ai_connect/main/settings/base.py
```diff
@@ -105,6 +105,20 @@
...
+CSRF_COOKIE_SAMESITE = "Lax"
+CSRF_COOKIE_NAME = "__Host-csrftoken"
+CSRF_COOKIE_AGE = 1209600
```
omaciel: The 1-year→2-weeks rationale is in the PR description but not in code. Magic-number `1209600` survives PR bitrot better with an inline note:

```python
CSRF_COOKIE_AGE = 1209600  # 2 weeks; reduced from Django's 1-year default to shrink the CSRF replay window
```

ansible_ai_connect/main/settings/base.py
```diff
@@ -105,6 +105,20 @@
 else:
     CSRF_TRUSTED_ORIGINS = ["http://localhost:8000"]
 
+# SameSite=Lax (not Strict) because OAuth/OIDC login flows redirect back from
+# external identity providers (Red Hat SSO, GitHub, AAP). Strict would drop the
+# session cookie on that redirect, making the user appear unauthenticated.
+SESSION_COOKIE_SECURE = True
+SESSION_COOKIE_HTTPONLY = True
+SESSION_COOKIE_SAMESITE = "Lax"
+SESSION_COOKIE_NAME = "__Host-sessionid"
```
omaciel, line 115: Worth being explicit about `SESSION_COOKIE_AGE` too. Django's default (1209600 = 2 weeks) happens to match the new `CSRF_COOKIE_AGE`, but pinning it makes the intent reviewable and pairs the two values:

```python
SESSION_COOKIE_NAME = "__Host-sessionid"
SESSION_COOKIE_AGE = 1209600  # 2 weeks; matches CSRF_COOKIE_AGE
```

ansible_ai_connect/main/settings/base.py
```diff
@@ -105,6 +105,20 @@
...
+
+CSRF_COOKIE_SECURE = True
+CSRF_COOKIE_HTTPONLY = False
+CSRF_COOKIE_SAMESITE = "Lax"
+CSRF_COOKIE_NAME = "__Host-csrftoken"
```
omaciel, line 124: Optional: add a small settings smoke test that pins the `__Host-` invariants — `*_COOKIE_DOMAIN is None`, `*_COOKIE_PATH == "/"`, `*_COOKIE_SECURE is True`. The `__Host-` prefix is browser-enforced: if a future change sets `SESSION_COOKIE_DOMAIN` or flips `SECURE` off, the browser will silently reject the cookie and users will lose their session with no obvious failure mode. A 5-line test prevents that drift.

ansible_ai_connect/main/settings/development.py
```diff
@@ -83,6 +83,10 @@
             "ansible_ai_connect.main.middleware.WisdomSocialAuthExceptionMiddleware"  # noqa: F405
         )
 
+SESSION_COOKIE_SECURE = False
+SESSION_COOKIE_NAME = "sessionid"
+CSRF_USE_SESSIONS = True
```
omaciel: **High-priority verification ask.** With `CSRF_USE_SESSIONS=True` and no CSRF cookie in dev, `readCsrfCookie()` will return `null` in the browser, and AJAX POSTs will ship `X-CSRFToken: "null"` (axios) or empty header. That will 403 against DRF's `SessionAuthentication`. Affected dev flows include:

- `ansible_ai_connect_admin_portal` → `saveWcaKey`, `deleteWcaKey`, `saveWcaModelId`, `saveTelemetrySettings`
- `ansible_ai_connect_chatbot` / `aap_chatbot` → `/ai/feedback/`, `/ai/streaming_chat/`, `/ai/chat/`

The PR description handwaves this with "endpoints get the token from the form context or are not used in the dev gateway flow," but nothing in the diff verifies that. Please confirm at least the admin portal Save buttons still succeed end-to-end in dev, or document the workaround (e.g. set `CSRF_USE_SESSIONS=False` when not behind the AAP gateway).

omaciel (commented):
Thorough review — the security design is sound (`__Host-` prefix, SameSite=Lax for OAuth, `rotate_token` on login). Inline comments below cover suggested clarifications and a couple of higher-priority verification asks before merge.

**Top-level items that don't map to a single line:**

1. **Empty Testing section.** The PR description still has the `1. ... 2. ... 3. ...` stub. Given this touches settings + signals + two frontends and is a security fix, please add concrete repro steps (at minimum: admin portal Save flow in dev, and a prod-like check that POSTs work with `__Host-csrftoken`).
2. **OpenAPI schema drift.** `tools/openapi-schema/ansible-ai-connect-service.yaml` (around line 1651) still declares `cookieAuth.name: sessionid`. In production this is now `__Host-sessionid`, so generated clients will use the wrong cookie name. Either update to `__Host-sessionid` or note the env-dependence.
3. **Backport labels not yet set.** The PR ticks "Backported to specific releases" but no `backport/stable-2.x` labels are applied. For a security fix please confirm which stable branches need it and verify clean cherry-pick — settings layout has historically diverged across stable-2.5/2.6.
4. **Committed bundle hashes vs source.** `templates/chatbot/index.html` ships rebuilt `index-HltiUewU.js` / `index-BCBeJGFG.css`. Reviewers can't validate these match the .ts source without rebuilding — if CI doesn't already rebuild-and-diff, consider adding that gate.

Nice work on the rotation test split (with/without request) and the `__Host-csrftoken ?? csrftoken` migration shape — those are exactly right.

ansible_ai_connect/users/signals.py
```diff
@@ -33,9 +34,11 @@
 
 
 @receiver(user_logged_in)
-def user_login_log(sender, user, **kwargs):
+def user_login_log(sender, user, request=None, **kwargs):
     """Successful user login log to user log"""
     logger.info(f"User: {user} LOGIN successful")
```
omaciel, line 39: Minor: consider rotating the token before the log line, so the "LOGIN successful" log only fires once the rotation has actually completed. If `rotate_token` ever raises, today we'd log success and then fail, which is misleading on incident triage.

```python
if request:
    rotate_token(request)
logger.info(f"User: {user} LOGIN successful")
```

Low-impact — happy to leave as-is if the existing ordering is intentional.

ansible_ai_connect/users/tests/test_signals.py
```diff
@@ -25,3 +29,19 @@ def test_obfuscate(self):
...
+
+
+class TestCSRFTokenRotationOnLogin(WisdomServiceLogAwareTestCase):
+
+    @patch("ansible_ai_connect.users.signals.rotate_token")
+    def test_csrf_token_rotated_on_login(self, mock_rotate_token):
+        request = Mock()
+        user = Mock()
+        user_logged_in.send(sender=self.__class__, user=user, request=request)
+        mock_rotate_token.assert_called_once_with(request)
```
omaciel, line 41: Optional: this test only asserts `rotate_token` was called. Wrapping the `send()` in `self.assertLogs("ansible_ai_connect.users.signals", level="INFO")` and asserting the "LOGIN successful" message still appears would lock in that the existing logging behavior wasn't lost when the signature was widened. Same idea for `test_csrf_token_not_rotated_without_request`.

ansible_ai_connect_admin_portal/src/api/api.ts
```diff
@@ -19,19 +19,22 @@ export const readCookie = (name: string): string | null => {
...
+
 export const getWcaKey = () => {
   return axios.get(API_WCA_KEY_PATH);
 };
 
 export const saveWcaKey = (wcaKey: WcaKeyRequest) => {
-  const csrfToken = readCookie("csrftoken");
+  const csrfToken = readCsrfCookie();
   return axios.post(API_WCA_KEY_PATH, wcaKey, {
     headers: { "X-CSRFToken": csrfToken },
```
omaciel: Pre-existing but worth fixing while we're in here: when `csrfToken` is `null` (now the common case in dev under `CSRF_USE_SESSIONS=True`), axios serializes the header value to the literal string `"null"`. Better to omit the header entirely so the failure surfaces as a clean missing-CSRF rather than a value-mismatch:

```ts
const csrfToken = readCsrfCookie();
return axios.post(API_WCA_KEY_PATH, wcaKey, {
  headers: csrfToken ? { "X-CSRFToken": csrfToken } : {},
});
```

Applies to all five POST/DELETE call sites in this file.

ansible_ai_connect_admin_portal/src/api/__tests__/api.test.ts
```diff
@@ -1,34 +1,60 @@
...
 
-  it("Cookie extraction::Multiple::With whitespace", () => {
-    document.cookie = "smurf = abcdef; csrftoken = 12345   ";
-    const cookie = readCookie("csrftoken");
-    expect(cookie).toEqual("12345");
+    it("Cookie extraction::Multiple::Without whitespace", () => {
+      document.cookie = "smurf=abcdef;__Host-csrftoken=12345";
+      const cookie = readCookie("__Host-csrftoken");
+      expect(cookie).toEqual("12345");
+    });
```
omaciel, line 46: Test parity: the chatbot test file added a `"does not match partial cookie names"` case (e.g. `csrftoken_extra=wrong` must not match `csrftoken`). Worth adding here too — `readCookie` uses `startsWith(name + "=")` which is the right behavior, and this is the canonical regression guard. Suggestion:

```ts
it("Cookie extraction::Partial-name mismatch", () => {
  document.cookie = "__Host-csrftoken_extra=wrong";
  const cookie = readCookie("__Host-csrftoken");
  expect(cookie).toBeNull();
});
```

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.test.ts
```diff
@@ -13,6 +18,54 @@ vi.mock("@microsoft/fetch-event-source", () => ({
...
+  afterEach(clearCookies);
+
+  it("falls back to csrftoken when __Host-csrftoken is absent", () => {
+    document.cookie = "csrftoken=dev_token";
+    expect(readCsrfCookie()).toEqual("dev_token");
+  });
+
+  it("returns null when neither cookie is present", () => {
+    expect(readCsrfCookie()).toBeNull();
+  });
```
omaciel, line 88: Test parity: the admin portal test added a `"prefers __Host-csrftoken over csrftoken"` case to lock in that the `??` chain falls through correctly when both cookies exist. Worth mirroring here:

```ts
it("prefers __Host-csrftoken over csrftoken", () => {
  document.cookie = "csrftoken=dev_token";
  document.cookie = "__Host-csrftoken=prod_token";
  expect(readCsrfCookie()).toEqual("prod_token");
});

it("returns __Host-csrftoken when present", () => {
  document.cookie = "__Host-csrftoken=prod_token";
  expect(readCsrfCookie()).toEqual("prod_token");
});
```

(And the same in `aap_chatbot/src/useChatbot/useChatbot.test.ts` — the two files are kept in lockstep.)

aap_chatbot/src/useChatbot/useChatbot.test.ts
```diff
@@ -13,6 +18,54 @@ vi.mock("@microsoft/fetch-event-source", () => ({
...
+  afterEach(clearCookies);
+
+  it("falls back to csrftoken when __Host-csrftoken is absent", () => {
+    document.cookie = "csrftoken=dev_token";
+    expect(readCsrfCookie()).toEqual("dev_token");
+  });
+
+  it("returns null when neither cookie is present", () => {
+    expect(readCsrfCookie()).toBeNull();
+  });
```
omaciel, line 102: Same suggestion as in `ansible_ai_connect_chatbot/.../useChatbot.test.ts`: add `"prefers __Host-csrftoken over csrftoken"` and `"returns __Host-csrftoken when present"` cases for parity with the admin portal tests.

ansible_ai_connect/main/templates/chatbot/index.html
```diff
@@ -9,8 +9,8 @@
       content="{{bot_name}}"
     />
     <title>{{bot_name}}</title>
-    <script type="module" crossorigin src="/static/chatbot/index-8zRWvsD-.js"></script>
-    <link rel="stylesheet" crossorigin href="/static/chatbot/index-CN-YAjNY.css">
+    <script type="module" crossorigin src="/static/chatbot/index-HltiUewU.js"></script>
+    <link rel="stylesheet" crossorigin href="/static/chatbot/index-BCBeJGFG.css">
```
omaciel, line 13: These hashed bundle filenames change every rebuild and aren't independently verifiable in review — if CI doesn't already rebuild the chatbot and confirm the committed `index-*.js`/`.css` match the .ts source, please run the build locally and double-check the committed bundle was produced from this branch (not a stale rebuild). Otherwise the production runtime could diverge from what the diff shows.

ldjebran:
Put the main PR in draft , we need to update containerized installer and lightspeed operator with with CSRF_TRUSTED_ORIGINS

a new approach with a lot of testing in progress while integrated with AAP, AAP-UI dev env, local dev environment and cloud with lightspeed http and https

ansible_ai_connect/main/settings/development.py
ldjebran: done

ansible_ai_connect/main/settings/base.py
ldjebran: done

ansible_ai_connect/main/settings/base.py
ldjebran: done

ansible_ai_connect_admin_portal/src/api/api.ts
ldjebran: done

ansible_ai_connect_admin_portal/src/api/__tests__/api.test.ts
ldjebran: done

ansible_ai_connect_chatbot/src/useChatbot/useChatbot.test.ts
ldjebran: done

ansible_ai_connect/main/templates/chatbot/index.html
ldjebran: yes it was produced by this branch, put a comment in commit

aap_chatbot/src/useChatbot/useChatbot.test.ts
ldjebran: done

ldjebran:
many changes has been from last version @omaciel can relaunch the review process.

ldjebran:
do not want to touch the npm audit fix for the moment, it has conflicting versions, in this PR solve the main issue only.
deleting some old npm reports to have only the most relevant from last push.

ldjebran:
rebased to have the latest npm packages fixes

omaciel (approved):
Approving — the inline feedback round landed thoroughly (10 of 12 asks addressed cleanly, the 2 skipped were ones I explicitly flagged as optional/low-impact). Verified against the latest diff:

- ✅ `CSRF_COOKIE_HTTPONLY=False` now has the "JS must read" rationale comment
- ✅ Magic-number `1209600` on both `SESSION_COOKIE_AGE` and `CSRF_COOKIE_AGE` is documented and explicitly aligned
- ✅ Four `__Host-` invariant tests landed (domain unset, path=/, etc.) — exactly the regression guard I asked for
- ✅ `CSRF_USE_SESSIONS=True` removed from dev settings with plain-name cookie overrides — closes the dev-flow concern
- ✅ All 5 admin portal POST/DELETE call sites use `csrfToken ? { "X-CSRFToken": csrfToken } : {}` instead of emitting `"null"` headers
- ✅ Test parity across the two chatbot test files and admin portal, including the bonus `"uses csrftoken when behind gateway"` case
- ✅ Bundle hash provenance attested in commit

Skipped (acceptable):
- Reorder `rotate_token` before log line — I marked as "happy to leave as-is if intentional"
- Optional `assertLogs` in test_signals — I marked as optional

The security design is right: `__Host-` prefix, `SameSite=Lax` for OAuth, `rotate_token` on login, the 4-case `_ensure_csrf_cookie` logic is intricate but well-commented and thoroughly tested (the `test_cookie_security.py` suite covers all cases including gateway mode variants).

### Three remaining items from my original top-level review, non-blocking but please address before/after merge

1. **PR description Testing section is still empty.** Given this is a security fix touching settings + signals + two frontends, even minimal repro steps ("admin portal Save in dev succeeds", "POST with `__Host-csrftoken` works in prod-like env") would help future bisects and security audit trails.

2. **OpenAPI schema drift.** `tools/openapi-schema/ansible-ai-connect-service.yaml` still declares `cookieAuth.name: sessionid` — production is now `__Host-sessionid`, so generated clients consuming this schema will use the wrong cookie name. Either update or annotate the env-dependence.

3. **Backport labels.** PR description ticks "Backported to specific releases" but no `backport/stable-2.x` labels are applied. For a security fix this should be explicit — particularly because `base.py` / `development.py` layout has historically diverged across `stable-2.5` / `2.6`, so cherry-picks may need touch-up.

### One CI note, not this PR's fix to make

The two failing `lint` jobs are blocked on `npm audit` finding `js-cookie <=3.0.5` (GHSA-qjx8-664m-686j) via `@patternfly/chatbot → @segment/analytics-next → js-cookie`. That's a pre-existing chatbot dep vulnerability, not introduced by this PR. The fix would require bumping `@patternfly/chatbot` to 2.1.0 (breaking change). Worth splitting that into a separate dep-bump PR rather than gating this security fix on it.

Nice work overall. 🚀

ldjebran:
all dependent PR/MR has been merged, put in ready for review.

PR #2046, opened 2026-05-25
ci: add immediate repo sync workflow to downstream

SiddharthRajaraman:
## Summary

Adds a push-triggered GitHub Actions workflow that mirrors commits to `ansible-automation-platform/ansible-wisdom-service` immediately on push to `main`, instead of relying on the hourly scheduled sync from `cicd-ops`.

Uses the shared `ansible-cicd` [repo-sync template](https://github.com/ansible/ansible-cicd/blob/main/.github/workflows/i-want-it-now.yaml) with `TARGET_REPO` overridden to `ansible-wisdom-service` (since the downstream repo name differs from upstream).

## Prerequisites

- [ ] `#ansible-github-admins` must grant visibility of the org secret `ANSIBLE_CICD_BOT_ORG_RW_TOKEN_PRIVATE_KEY` to this repo

## Test plan

- Verify the workflow appears in the Actions tab after merge
- Once the org secret is granted, confirm that pushing to `main` triggers an immediate sync to `ansible-automation-platform/ansible-wisdom-service`

### DIFF
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -1,0 +1,68 @@
+name: "Sync to downstream"
+
+on:
+  push:
+    branches:
+      - main
+    # Tags are not mirrored by this workflow — branch sync only.
+
+permissions:
+  contents: read
+
+concurrency:
+  group: repo-sync-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  SOURCE_ORG: "ansible"
+  TARGET_ORG: "ansible-automation-platform"
+  TARGET_REPO: "ansible-wisdom-service"
+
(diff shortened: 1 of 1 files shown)


omaciel (approved):
Same fleet rollout as ansible/ansible-chatbot-stack#262 and ansible/ansible-mcp-tools#78 — identical diff modulo the per-repo settings. Good call hardcoding the downstream as `ansible-wisdom-service` (different name from upstream) — that's the renamed-downstream case the `TARGET_REPO` portability concern was about, and it's handled correctly. LGTM! 🚀

Same nudge as on the others: once verified working, please retire the corresponding hourly scheduled sync.

PR #2047, opened 2026-05-25
add automerge for dependabotprs based on checks validation

acosferreira:
create workflow to alto-merge dependabot prs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
     open-pull-requests-limit: 3
     labels:
       - "dependencies"
-      - "python"
+      - "python:uv"
       - "security"
     commit-message:
       prefix: "deps"
@@ -27,7 +27,6 @@ updates:
     labels:
       - "dependencies"
       - "javascript"
-      - "admin-portal"
       - "security"
     commit-message:
       prefix: "deps"
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,40 @@
+---
+name: Auto-merge Dependabot PRs
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  contents: write
+  pull-requests: write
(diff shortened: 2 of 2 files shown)


omaciel:
@acosferreira — closing this, and I want to be clear it is because the ground moved underneath it, not because of anything wrong with the work.

**What changed.** As of yesterday, Dependabot version updates are disabled in this repository: #2182 set `open-pull-requests-limit: 0` on all four ecosystems, and Dependabot *security* updates were already disabled at the repository level (`security_and_analysis.dependabot_security_updates: disabled`). MintMaker (Renovate) is now the single source of dependency updates — running both was producing duplicate PRs for the same bumps across four ecosystems.

This workflow is gated on:

```yaml
if: github.event.pull_request.user.login == 'dependabot[bot]'
```

so it can no longer fire. The nine open Dependabot PRs were closed at the same time.

**The idea was right, and the implementation was careful** — the action is SHA-pinned, majors are excluded from automerge and commented instead of silently skipped, and failures leave a link to the run. None of that is why this is closing.

**If you still want automerge, it now belongs in `renovate.json` rather than a workflow.** Renovate has it natively, and the fleet already relies on it: the shared `aap-konflux-pipelines//renovate/default.json` sets `lockFileMaintenance.automerge: true`, and `ansible-application-code.json` automerges AAP submodule patch bumps. The equivalent of what this PR does is roughly:

```json
{
  "matchUpdateTypes": ["patch", "pin", "digest"],
  "automerge": true
}
```

That also gets you the major-version exclusion for free, without needing `fetch-metadata` to determine the update type.

One caveat worth carrying over if you do: this repo now has required checks that a bot cannot satisfy on its own — see #2181 for why `requirements.txt` is generated and must not be edited directly. Any automerge rule should sit behind those checks passing, which Renovate's `automerge` does respect.

Happy to help draft that rule if it is useful. The label tidy-up in `.github/dependabot.yml` here (`python` → `python:uv`) is a genuine improvement that got lost — it just no longer has an effect while the ecosystems are paused.

PR #2051, opened 2026-05-26
ci: skip repo-sync workflow in downstream org

SiddharthRajaraman:
## Summary

Adds an `if: github.repository_owner == 'ansible'` guard to the repo-sync workflow job so it is silently skipped when the workflow file is mirrored to the downstream `ansible-automation-platform` org.

Without this guard, every sync push to downstream `main` would re-trigger the workflow there — either failing (if the upstream/downstream repo names differ) or causing an infinite loop (if the names match).

## Test plan

- [ ] Verify the workflow still triggers normally on push to `main` in the upstream `ansible` org
- [ ] Confirm the job shows as "skipped" in the downstream org's Actions tab after the next sync

### DIFF
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -21,6 +21,7 @@ env:
 jobs:
   sync:
     name: "Sync → ansible-wisdom-service"
+    if: github.repository_owner == 'ansible'
     runs-on: ubuntu-latest
 
     steps:


omaciel (approved):
Same fleet rollout as ansible/ansible-chatbot-stack#264. Extra important here since the downstream name (`ansible-wisdom-service`) differs from upstream — without this guard the mirrored workflow fails loudly on every sync trying to push to a nonexistent TARGET_REPO from the downstream side. The org guard fixes that cleanly. LGTM 🚀

PR #2055, opened 2026-05-29
Fix polynomial-runtime ReDoS in Ollama answer parsers (SonarCloud S5852)

omaciel:
## What

Rewrites the three fenced-block extraction regexes in `ollama/pipelines.py` (`unwrap_message_with_yaml_answer`, `unwrap_playbook_answer`, `unwrap_task_answer`) from a backtracking-prone form to a linear one, resolving the three **`python:S5852`** Security Hotspots SonarCloud flagged in the new-code period:

> Make sure the regex used here, which is vulnerable to polynomial runtime due to backtracking, cannot lead to denial of service.

## Why

Each regex combined two overlapping unbounded quantifiers, the classic polynomial-backtracking shape:

- a leading `.*?` (redundant under `re.search`, which already scans) followed by a greedy `(.+)` — two unbounded `.`-matchers separated by a literal (` ``` `) that `.` can itself match;
- a `\n+` separator that overlaps the `(.+)` content match under `re.DOTALL` (`\n` is matched by both).

## The change

```diff
- re.search(r".*?```(yaml|yml|)\n+(.+)```", task, re.MULTILINE | re.DOTALL)
+ re.search(r"```(yaml|yml|)\n(.+?)```", task, re.MULTILINE | re.DOTALL)
```

(and the analogous two). Drop the redundant `.*?`, collapse `\n+` → `\n`, make the content capture lazy `(.+?)`.

## Behavior

For a **single** fenced block — the normal model output and every existing test — behavior is identical. As a side effect, the parsers now extract only the **first** fenced block instead of greedily spanning from the first fence to the *last* one (the old behavior merged multiple code blocks and their literal ` ``` ` markers into the returned YAML, which was never valid). New `TestUnwrapAnswerFirstBlock` cases lock this in.

## Testing

- Added `TestUnwrapAnswerFirstBlock` (3 cases) — fail on the old regexes, pass on the new ones.
- Existing `TestUnwrapTaskAnswer` / `TestUnwrapPlaybookAnswer` / `TestUnwrapRoleAnswer` / pipeline tests are unaffected (all single-block).

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/ollama/pipelines.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/ollama/pipelines.py
@@ -131,14 +131,14 @@ def unwrap_role_answer(message: str, create_outline: bool) -> tuple[str, list[di
 def unwrap_message_with_yaml_answer(message: str) -> str:
     task: str = message_to_string(message)
 
-    m = re.search(r".*?```(yaml|yml|)\n+(.+)```", task, re.MULTILINE | re.DOTALL)
+    m = re.search(r"```(yaml|yml|)\n(.+?)```", task, re.MULTILINE | re.DOTALL)
     return m.group(2).strip() if m else ""
 
 
 def unwrap_playbook_answer(message: str) -> tuple[str, str]:
     task: str = message_to_string(message)
 
-    m = re.search(r".*?```(yaml|)\n+(.+)```(.*)", task, re.MULTILINE | re.DOTALL)
+    m = re.search(r"```(yaml|)\n(.+?)```(.*)", task, re.MULTILINE | re.DOTALL)
     if m:
         playbook = m.group(2).strip()
         outline = m.group(3).lstrip().strip()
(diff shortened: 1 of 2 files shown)


omaciel:
This is a fix based on SonarCloud scan report.

hasys (approved):
LGTM - Ship it!

  Summary

  Clean fix for SonarCloud S5852. Transforms three regexes from polynomial-backtracking form to linear:

  - r".*?```(yaml|yml|)\n+(.+)```"
  + r"```(yaml|yml|)\n(.+?)```"

  Eliminates overlapping quantifiers (.*? + .+ with overlapping \n+) that caused polynomial runtime.

  Verified
  
  - ✅ All existing tests pass (single-block case unchanged)
  - ✅ New TestUnwrapAnswerFirstBlock tests prove first-block extraction works
  - ✅ Validated locally: 50-130x speedup on pathological input (no closing backticks)

  Bonus Fix

  Old regex greedily spanned first to *last*, merging multiple blocks into invalid YAML. New regex correctly extracts only the first block. Since Ollama models can be inconsistent, this actually improves resilience.

  Minor Note
  
  Regex now requires exactly \n after opening fence - won't match ˋˋˋyaml<no-newline>code. Given this is dev tooling with flaky models, consider adding a fallback or logging when extraction fails, but not a blocker.

PR #2074, opened 2026-06-15
Increase chatbot readiness health check timeout from 1s to 5s

TamiTakamiya:
## Summary
- Increases the `/readiness` call timeout in `HttpChatBotMetaData.self_test()` from 1 second to 5 seconds
- The 1s timeout caused intermittent false-negative health check failures when external LLM providers (Azure OpenAI, RHOAI) were slow to respond to health probes
- Normal-case checks complete in ~0.2s, so 5s provides comfortable margin without slowing down the happy path

Closes: [AAP-77500](https://issues.redhat.com/browse/AAP-77500)

## Test plan
- [ ] Verify health check passes consistently with Azure OpenAI provider
- [ ] Verify health check passes consistently with RHOAI provider
- [ ] Confirm no regression in normal-case health check response time

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/http/pipelines.py
@@ -154,7 +154,7 @@ class HttpChatBotMetaData(HttpMetaData):
             r = self.session.get(
                 self.config.inference_url + "/readiness",
                 headers=headers,
-                timeout=1,
+                timeout=5,
             )
             r.raise_for_status()
 


de1987 (approved):

omaciel (approved):
I approve it and want to ask: would a process of tries (maybe 5 times) until a maximum timeout or criteria is met be more elegant and robust?

TamiTakamiya:
> I approve it and want to ask: would a process of tries (maybe 5 times) until a maximum timeout or criteria is met be more elegant and robust?

The Jira ticket was opened for intermittent failures on ATF runs. I would like to try it out with this simple fix (increasing the timeout). 

We observed that other chatbot test cases completed successfully even when the health check test failed. The lightspeed-stack/llama-stack should be working and I think this change on the timeout value suggested by Claude makes sense.

However, if this does not fix the inttermittent failure on the chatbot health check. I will re-visit and spend more time on investigation.

PR #2083, opened 2026-07-07
Bump transitive deps to fix pip-audit CVEs

resoluteCoder:
## Summary
- Bump `cryptography` constraint to `>=48.0.1` (resolves to 49.0.0). This is a 2-major-version jump from the previously pinned 46.x, but the cryptography project follows a rapid major versioning scheme — breaking changes are minimal and largely affect low-level API users. Our usage is via pyopenssl/autobahn/django-ansible-base which all support 49.x.
- Bump `daphne` to `~=4.2.2`
- Bump `ujson` constraint to `>=5.13.0`
- Add `msgpack>=1.2.1` constraint to fix CVE in transitive dep from autobahn (uv export was resolving to 1.1.2 without the explicit constraint due to a platform marker resolution issue)

All four are transitive/constraint-dependency bumps in `pyproject.toml` with regenerated `uv.lock` and `requirements.txt`.

## Test plan
- [ ] pip-audit CI passes
- [ ] Existing tests unaffected (no API changes)


SiddharthRajaraman (approved):
lgtm

PR #2104, opened 2026-07-22
AAP-83188 AAP-83186: Bump js-yaml to 4.3.0 in ansible_ai_connect_chatbot (CVE-2026-59869)

SiddharthRajaraman:
## Summary
- Bumps transitive `js-yaml` to 4.3.0 in `ansible_ai_connect_chatbot` via npm overrides to address CVE-2026-59869
- Admin portal / aap_chatbot were already fixed at 4.3.0 in #2103; chatbot lock still resolved 4.2.0

## Security
- [x] This is a security fix

## Jira
- https://redhat.atlassian.net/browse/AAP-83188
- https://redhat.atlassian.net/browse/AAP-83186

## Exposure
- `js-yaml` is a transitive build/tooling dependency (eslint/cosmiconfig); no direct app imports
- Not used for runtime YAML parsing in production Lightspeed images

## Test plan
- [ ] Verify lockfile resolves only js-yaml@4.3.0 under ansible_ai_connect_chatbot
- [ ] CI passes

## Backport
- Downstream stables still vulnerable; separate backports as needed (no dedicated container tickets currently)

PR #2102, opened 2026-07-22
AAP-80970 AAP-80967: Bump fast-uri to >=3.1.3 (CVE-2026-13676)

SiddharthRajaraman:
## Summary
- Security fix for CVE-2026-13676 (`fast-uri` Unicode hostname canonicalization / host policy bypass)
- Bumps `overrides.fast-uri` from `>=3.1.1` to `>=3.1.3` in `aap_chatbot` and `ansible_ai_connect_admin_portal`
- Updates `ansible_ai_connect_admin_portal` lockfile from `fast-uri` 3.1.2 → 3.1.4 (`aap_chatbot` already at 4.1.1)

## Jira
- https://issues.redhat.com/browse/AAP-80970
- https://issues.redhat.com/browse/AAP-80967

## Exposure
Build-time transitive dependency via `ajv`. Not used for HTTP host-based policy enforcement. Risk assessed as Low.

## Test plan
- [ ] Verify lockfiles resolve `fast-uri` >= 3.1.3
- [ ] CI passes (lint/unit/build for admin portal and aap_chatbot)

## Checklist
- [x] This is a security fix
- [ ] Backports will be handled separately

Made with [Cursor](https://cursor.com)

PR #2105, opened 2026-07-23
fix: resolve chatbot npm audit vulnerabilities

SiddharthRajaraman:
## Summary
- Bump `vitest` / `@vitest/browser` / `@vitest/coverage-v8` to ≥3.2.7 (GHSA-p63j-vcc4-9vmv)
- Tighten npm overrides for `dompurify` (≥3.4.12), `immutable` (≥5.1.8), and `brace-expansion` so Chatbot npm-audit passes with `audit_level: low`

## Context
These findings are currently blocking unrelated PRs (e.g. #2104) via the Chatbot - npm-audit workflow.

## Test plan
- [ ] Chatbot npm-audit CI green
- [ ] Local: `cd ansible_ai_connect_chatbot && npm audit --audit-level=low`

### DIFF
--- a/ansible_ai_connect_chatbot/package-lock.json
+++ b/ansible_ai_connect_chatbot/package-lock.json
@@ -25,8 +25,8 @@
         "@eslint/compat": "^1.2.1",
         "@eslint/js": "^9.13.0",
         "@testing-library/react": "^16.2.0",
-        "@vitest/browser": "^3.2.4",
-        "@vitest/coverage-v8": "^3.2.4",
+        "@vitest/browser": "^3.2.7",
+        "@vitest/coverage-v8": "^3.2.7",
         "buffer": "^6.0.3",
         "empty-module": "^0.0.2",
         "eslint": "^9.13.0",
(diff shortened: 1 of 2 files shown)


omaciel (approved):
Approving — verified this clears the findings rather than just moving version strings. The `Chatbot - npm-audit` workflow runs at `audit_level: low` and is green on the head, and each floor resolves to a real published version (dompurify 3.4.12, immutable 5.1.8, js-yaml 4.3.0 — 4.2.1 doesn't exist, 4.3.0 is the patched 4.x). No file overlap with #2102/#2103/#2106 (all in different package dirs), so this lands independently and unblocks the chatbot audit for #2104. Nice that the overrides are all `>=` floors here — more maintainable than mixing exact pins.

One non-blocking cleanup before you merge: the description says it tightens `brace-expansion`, but the diff doesn't touch it — it adds `js-yaml >=4.3.0` and `immutable >=5.1.8` instead (looks like text carried over from the admin-portal PR). Worth correcting the description so the record matches the diff, and the test-plan checkboxes are safe to tick since the audit + build/test both passed.

🤖 Drafted with AI assistance

PR #2109, opened 2026-07-28
fix: retry WCA API calls on HTTP 422 responses

TamiTakamiya:
## Summary
- Add HTTP 422 to the list of non-fatal (retryable) status codes in the WCA pipeline's `fatal_exception` method
- Recent WCA changes occasionally (~1% of the time) return 422 for code generation API calls from ARI (Ansible Risk Insight)
- With this change, 422 responses will be retried with exponential backoff (up to 4 retries by default), matching the existing behavior for 429 (rate limited) responses

## Test plan
- [ ] Verify that WCA API calls returning 422 are retried with exponential backoff
- [ ] Verify that other 4xx errors (e.g. 400, 403, 404) are still treated as fatal (not retried)
- [ ] Verify that 429 and 5xx errors continue to be retried as before

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
@@ -236,9 +236,9 @@ class WCABaseMetaData(
         if isinstance(exc, requests.RequestException):
             status_code = getattr(getattr(exc, "response", None), "status_code", None)
             # retry on server errors and client errors
-            # with 429 status code (rate limited),
+            # with 429 status code (rate limited) and 422 status code,
             # don't retry on other client errors
-            return bool(status_code and (400 <= status_code < 500) and status_code != 429)
+            return bool(status_code and (400 <= status_code < 500) and status_code not in (422, 429))
         else:
             # retry on all other errors (e.g. network)
             return False
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/pipelines_base.py
@@ -238,7 +238,9 @@ class WCABaseMetaData(
             # retry on server errors and client errors
             # with 429 status code (rate limited) and 422 status code,
             # don't retry on other client errors
-            return bool(status_code and (400 <= status_code < 500) and status_code not in (422, 429))
+            return bool(
+                status_code and (400 <= status_code < 500) and status_code not in (422, 429)
+            )
         else:
             # retry on all other errors (e.g. network)
             return False
--- a/ansible_ai_connect/ai/api/model_pipelines/wca/tests/test_pipelines_base_ssl.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/wca/tests/test_pipelines_base_ssl.py
@@ -27,9 +27,7 @@ from unittest.mock import Mock, patch
 import requests
 from django.test import SimpleTestCase, override_settings
 
-from ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_base import (
-    WCABaseMetaData,
-)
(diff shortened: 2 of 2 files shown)


de1987 (approved):

omaciel:
## I don't think `fatal_exception` is reachable for HTTP status codes

Thanks for chasing the ~1% ARI failures — that's a real problem worth fixing. But I traced the retry path and I believe this change can't take effect, because `fatal_exception` never sees an HTTP status code in production. Happy to be shown wrong if I've missed a path.

### The retried function doesn't raise on 4xx

At all six `giveup=self.fatal_exception` sites, the decorated function returns the response without raising:

```python
@backoff.on_exception(backoff.expo, Exception, max_tries=self.retries + 1,
                      giveup=self.fatal_exception, ...)
def post_request():
    return self.session.post(prediction_url, ...)   # no raise_for_status here

response = post_request()
context = Context(model_id, response, task_count > 1)
InferenceResponseChecks().run_checks(context)
response.raise_for_status()                          # outside the retry scope
```

`raise_for_status()` is outside the `@backoff` wrapper in every case — `pipelines_base.py` lines 434, 496, 582, 669, 756 and 821.

I checked the alternatives that would make a 4xx raise inside `post_request()` and didn't find them:

- no `hooks={'response': ...}` on the session
- no `urllib3.Retry` / `status_forcelist` / `raise_on_status` on any mounted adapter — the only one is `AllowBrokenSSLContextHTTPAdapter`

So a 422 is an ordinary `requests` outcome, nothing raises inside `post_request()`, and `backoff` is never entered. The exceptions that *do* reach `fatal_exception` are transport-level (`ConnectionError`, `Timeout`, `SSLError`), which carry `response=None` → `status_code=None` → `bool(None and ...)` → `False` → retry. The `400 <= status_code < 500` branch looks unreachable.

The new unit tests pass because they call `fatal_exception` directly, which bypasses the path where this would show up.

### This implies 429 isn't retried today either

The description says this matches "the existing behavior for 429 (rate limited) responses". If the above is right, there is no such behavior — WCA rate-limiting has never actually been retried. That may be the more valuable bug of the two.

### Fastest way to confirm

Item 1 of your own test plan is the decisive check: *"Verify that WCA API calls returning 422 are retried with exponential backoff."* Written as an integration test with a mocked 422 response and an assertion on the call count, I'd expect it to fail against this branch. If it passes, my analysis is wrong and I'd genuinely like to see where.

### If you do make the retry reachable

Moving `raise_for_status()` inside `post_request()` would work, but it needs care: `InferenceResponseChecks().run_checks(context)` currently runs *before* `raise_for_status()` and converts specific statuses into typed errors.

422 in particular is already overloaded — `wca_utils.py:205`:

```python
class ResponseStatusCode422WCAValidationFailure(Check[Context]):
    def check(self, context: Context):
        if context.result.status_code == 422:
            payload_detail = payload_json.get("detail")
            if payload_detail and "validation failed" in payload_detail.lower():
                raise WcaValidationFailure(...)
```

So 422 means both a genuine deterministic validation failure (when `detail` contains "validation failed") and the spurious ARI case you're targeting. A blanket status-code retry can't tell them apart — a real validation failure would get retried four times with exponential backoff before surfacing the same 400 to the user. Discriminating on the payload rather than the status alone would avoid that.

Also worth noting `fatal_exception` is shared across all six pipelines (codegen, content match, playbook generation, role generation, playbook explanation, role explanation), so any change here is global even though the reported problem is codegen-specific.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

TamiTakamiya:
Closing based on Og's review — the `fatal_exception` status code logic is unreachable because `raise_for_status()` is called outside the `@backoff` scope. Will rethink the approach to properly handle the sporadic 422s from ARI code generation calls.

PR #2110, opened 2026-07-29
fix: retry WCA API calls on transient HTTP 422 and 429 responses

TamiTakamiya:
## Summary
- Recent WCA changes occasionally (~1% of the time) return HTTP 422 for code generation API calls from ARI (Ansible Risk Insight)
- The existing retry logic via `@backoff.on_exception` was unreachable for HTTP status code errors because `raise_for_status()` was called outside the backoff scope (as identified in [PR #2109 review](https://github.com/ansible/ansible-ai-connect-service/pull/2109#issuecomment-5104797514))
- This PR adds `_raise_for_retryable_status()` inside each `post_request()` function so that transient 422 and 429 responses actually trigger exponential backoff retries
- Genuine "validation failed" 422 responses are excluded from retry by inspecting the response body, preserving existing `WcaValidationFailure` behavior

## Changes
- **`exceptions.py`**: Added `WcaRetryableHttpError` exception (extends `WcaException`, not `requests.RequestException`, so `fatal_exception()` returns `False` and backoff retries)
- **`pipelines_base.py`**: Added `_raise_for_retryable_status()` static method and call it inside all six pipeline `post_request()` functions; catch `WcaRetryableHttpError` in caller exception handlers
- **`test_pipelines_base_ssl.py`**: Unit tests for `fatal_exception` and `_raise_for_retryable_status` (11 test cases covering 200, 400, 403, 422-transient, 422-validation-failed, 429, 500)
- **`test_wca_client.py`**: Integration tests verifying transient 422 is retried and succeeds on second attempt, and that exhausted retries raise `WcaInferenceFailure`

## Test plan
- [ ] Verify that WCA API calls returning transient 422 are retried with exponential backoff
- [ ] Verify that genuine "validation failed" 422 responses are NOT retried
- [ ] Verify that 429 responses are retried
- [ ] Verify that other 4xx errors (e.g. 400, 403, 404) are still treated as fatal (not retried)
- [ ] Verify existing tests pass without regressions

### DIFF
--- a/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
@@ -119,7 +119,7 @@ class WcaRequestIdCorrelationFailure(WcaException):
 
 @dataclass
 class WcaRetryableHttpError(WcaException):
-    """A retryable HTTP error from WCA (transient 422, 429, 5xx)."""
+    """A retryable HTTP error from WCA (transient 422, 429)."""
 
     status_code: int = 0
 
--- a/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
@@ -104,6 +104,16 @@ class WcaValidationFailure(WcaException):
 
 
 @dataclass
+class WcaGenerationFailure(WcaException):
+    """An attempt to run a WCA playbook or role generation failed."""
+
+
+@dataclass
+class WcaExplanationFailure(WcaException):
+    """An attempt to run a WCA playbook or role explanation failed."""
+
+
+@dataclass
 class WcaCodeMatchFailure(WcaException):
     """An attempt to run a WCA code match failed."""
 
--- a/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/exceptions.py
@@ -118,5 +118,12 @@ class WcaRequestIdCorrelationFailure(WcaException):
 
 
 @dataclass
+class WcaRetryableHttpError(WcaException):
+    """A retryable HTTP error from WCA (transient 422, 429, 5xx)."""
+
+    status_code: int = 0
+
+
+@dataclass
 class WcaInstanceDeleted(WcaException):
     """WCA Instance associated with the Model ID has been deleted."""
--- a/ansible_ai_connect/ai/api/model_pipelines/tests/test_wca_client.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/tests/test_wca_client.py
@@ -37,6 +37,8 @@ from ansible_ai_connect.ai.api.model_pipelines.exceptions import (
     WcaBadRequest,
     WcaCodeMatchFailure,
     WcaEmptyResponse,
+    WcaExplanationFailure,
+    WcaGenerationFailure,
     WcaInferenceFailure,
     WcaInvalidModelId,
     WcaKeyNotFound,
@@ -378,7 +380,7 @@ class TestWCAClientPlaybookGeneration(WisdomAppsBackendMocking, WisdomServiceLog
         model_client.session = Mock()
         model_client.session.post = Mock(side_effect=HTTPError(500))
         with (
-            self.assertRaises(WcaInferenceFailure),
+            self.assertRaises(WcaGenerationFailure),
             self.assertLogs(
                 logger="ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_base", level="INFO"
             ) as log,
@@ -620,7 +622,7 @@ class TestWCAClientExplanation(WisdomAppsBackendMocking, WisdomServiceLogAwareTe
         model_client.session = Mock()
         model_client.session.post = Mock(side_effect=HTTPError(500))
         with (
-            self.assertRaises(WcaInferenceFailure),
+            self.assertRaises(WcaExplanationFailure),
             self.assertLogs(
                 logger="ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_base", level="INFO"
             ) as log,
--- a/ansible_ai_connect/ai/api/model_pipelines/tests/test_wca_client.py
+++ b/ansible_ai_connect/ai/api/model_pipelines/tests/test_wca_client.py
@@ -67,10 +67,13 @@ from ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_base import (
     wca_codegen_playbook_retry_counter,
     wca_codegen_retry_counter,
     wca_codegen_role_hist,
+    wca_codegen_role_retry_counter,
     wca_codematch_hist,
     wca_codematch_retry_counter,
     wca_explain_playbook_hist,
     wca_explain_playbook_retry_counter,
+    wca_explain_role_hist,
+    wca_explain_role_retry_counter,
 )
 from ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_onprem import (
     WCAOnPremCompletionsPipeline,
@@ -546,6 +549,29 @@ class TestWCAClientRoleGeneration(WisdomAppsBackendMocking, WisdomServiceLogAwar
                 continue
             self.assertEqual(file["content"], "---\n- ansible.builtin.package:\n    name: emacs\n")
 
+    @assert_call_count_metrics(metric=wca_codegen_role_hist)
+    @assert_call_count_metrics(metric=wca_codegen_role_retry_counter)
+    def test_role_gen_error(self):
+        request = Mock()
+        model_client = WCASaaSRoleGenerationPipeline(mock_pipeline_config("wca"))
+        model_client.get_api_key = Mock(return_value="some-key")
+        model_client.get_token = Mock(return_value={"access_token": "a-token"})
+        model_client.get_model_id = Mock(return_value="a-random-model")
+        model_client.session = Mock()
+        model_client.session.post = Mock(side_effect=HTTPError(500))
+        with (
+            self.assertRaises(WcaGenerationFailure),
+            self.assertLogs(
+                logger="ansible_ai_connect.ai.api.model_pipelines.wca.pipelines_base", level="INFO"
+            ) as log,
+        ):
+            model_client.invoke(
+                RoleGenerationParameters.init(
+                    request=request, text="Install Wordpress", create_outline=True
+                )
+            )
+            self.assertInLog("Caught retryable error after 1 tries.", log)
+
 
 @override_settings(WCA_SECRET_BACKEND_TYPE="dummy")
 @override_settings(ENABLE_ANSIBLE_LINT_POSTPROCESS=False)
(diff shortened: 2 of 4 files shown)


de1987 (dismissed):

omaciel (dismissed):
## Approving — this correctly fixes the unreachability from #2109

Thanks for turning that around so quickly, and for going after the harder version of the fix rather than the one-line one.

### The core problem is genuinely solved

`_raise_for_retryable_status()` is called *inside* the `@backoff`-decorated `post_request()` at all six sites, so retryable statuses now actually enter the retry loop. The integration test proves it rather than asserting on a predicate:

```python
model_client.session.post = Mock(side_effect=[transient_422, success_200])
...
self.assertEqual(model_client.session.post.call_count, 2)
```

That's exactly the check that would have failed against #2109, and it's the reason I'm confident this one works.

`test_infer_transient_422_exhausts_retries` asserting `call_count == self.config.retry_count + 1` is a good complement — it pins the exhaustion path too.

### The 422 overloading is handled correctly

The exclusion matches `ResponseStatusCode422WCAValidationFailure` in `wca_utils.py` exactly — same `detail` key, same `"validation failed" in detail.lower()` test — so a genuine validation failure still returns early and still surfaces as `WcaValidationFailure`. `test_422_validation_failed_case_insensitive` covers the casing, and `test_422_non_json_raises` / `test_422_empty_body_raises` cover the degenerate bodies. This was the part I was most worried about and it's tighter than I expected.

### 429 now actually works

Worth calling out explicitly: the pre-existing 429 handling was dead for the same reason 422 was. This revives it. That was arguably the more valuable half of the bug.

Keeping `fatal_exception` unchanged is right, since `WcaRetryableHttpError` extends `WcaException` rather than `requests.RequestException` and so takes the `else` branch. `test_unprocessable_entity_is_fatal` asserting `True` documents that deliberately, which is a nice touch.

### Two non-blocking notes

**1. Typo in the comment** — `pipelines_base.py:264`:

```python
#    This should bThae retried via backoff.
```

**2. Four of the six pipelines don't catch `WcaRetryableHttpError`.**

`infer_from_parameters` (L465) and content match (L534) were updated to `except (HTTPError, WcaRetryableHttpError)`. Playbook generation, role generation, playbook explanation and role explanation call `_raise_for_retryable_status()` but have no `try`/`except` at all, and `WcaRetryableHttpError` appears nowhere outside `pipelines_base.py` and the tests — `views.py` catches a long list of specific `Wca*` types and this is not among them.

To be fair, those four were *already* unguarded: `result.raise_for_status()` would have raised an uncaught `HTTPError` just the same. So this is not a new class of bug. But two things do change for them:

- the uncaught exception type changes from `requests.HTTPError` to `WcaRetryableHttpError`, so anything upstream keying on `HTTPError` would stop matching
- a persistent 429 on those endpoints now burns four backoff intervals before failing with the same unmapped error

Not blocking, and arguably out of scope for a fix targeting codegen-from-ARI. But if the intent is that all six behave consistently, those four want handlers too — or a follow-up issue so it doesn't get lost.

**Minor observability note:** `@wca_codegen_hist.time()` sits inside the `@backoff` decorator, so a retried call now records one histogram observation per attempt rather than per call. Consistent with how network-error retries already behaved, so probably what you want — just worth knowing before someone reads a latency graph.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

omaciel (dismissed):
## Re-approving — `1a697daa` addresses both notes

Verified against the new head:

```
6 × _raise_for_retryable_status(result)          call sites
6 × except (HTTPError, WcaRetryableHttpError)    handlers  (L465, 534, 620, 716, 812, 886)
```

All six pipelines are now symmetric, and the `bThae` typo is gone.

### The behavior change on the four newly-guarded pipelines is safe

Worth spelling out what changed, since the `try` block grew rather than just gaining an `except`:

Previously, playbook generation / role generation / playbook explanation / role explanation had no handler at all — `result.raise_for_status()` raised an uncaught `requests.HTTPError`. Now that becomes `WcaInferenceFailure`, with a log line.

I checked where each lands at the API boundary: `views.py` catches ten specific `Wca*` types (L706–773) and neither `HTTPError` nor `WcaInferenceFailure` is among them, so **both** fall through to the same catch-all at L781. So this is not a regression and not a user-visible change — it's a consistency and logging improvement. Fine either way, just noting it so nobody expects a new error mapping that isn't there.

I also checked the widened `try` doesn't swallow anything it shouldn't. `InferenceResponseChecks().run_checks()` raises `WcaValidationFailure`, `WcaBadRequest`, `WcaInvalidModelId`, `WcaInstanceDeleted` and friends, and `WcaRequestIdCorrelationFailure` is raised directly inside the block — none are in the `except` tuple, so they all still propagate to their existing handlers. Behaviour preserved.

### One remaining note, non-blocking

**The four new handlers have no test coverage.** `1a697daa` is code-only — 69 insertions in `pipelines_base.py`, no test changes. The codegen path got genuinely good tests in the earlier commits (`test_infer_transient_422_is_retried` asserting `call_count == 2` is what convinced me the fix was real), so it's a little asymmetric that four newly-added handlers ship untested.

Not worth blocking on — they're four copies of a pattern already proven on codegen. But one test on, say, playbook generation exercising retry-then-fail would close the gap cheaply, and would catch it if one of the four copies drifts later.

### CI

`Unit Test and SonarCloud Scan` and both Konflux checks were still `pending` when I reviewed. Everything else is green. Please confirm the unit-test run completes before merging — it's the check that would catch a regression in the four pipelines this commit touched.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

omaciel (dismissed):
## Re-approving — and correcting one thing I got wrong last round

Two new commits since my last pass: `27dbddea` (tests expect `WcaInferenceFailure` rather than `HTTPError`) and `ec4360f7` (new `WcaGenerationFailure` / `WcaExplanationFailure` types).

### The new exception types are a real improvement

Raising `WcaInferenceFailure` from a *playbook explanation* path was always slightly wrong semantically. Splitting into `WcaGenerationFailure` and `WcaExplanationFailure` fixes that, and it's more than cosmetic — it means these paths can be mapped individually later without disturbing codegen.

### I checked the retype doesn't break existing handlers

This was the risk worth verifying: two places catch `WcaInferenceFailure`, and if either sat on a generation or explanation path, silently changing the raised type would break it.

- `pipelines_saas.py:278` — inside `self_test()`, which calls `infer_from_parameters`
- `inference.py:207` — the completion stage

Both are on the **codegen** path, which still raises `WcaInferenceFailure`. `ec4360f7` only touched the four generation/explanation pipelines. So nothing that previously caught it stops catching it.

The two new types aren't mapped in `views.py`, so they land on the catch-all at L781 exactly as `WcaInferenceFailure` would have — no user-visible change, just correct semantics and a hook for future mapping.

### Correction: I was wrong that the four handlers were untested

Last round I said the new handlers shipped without test coverage. That was only half right, and I should have checked more carefully before saying it.

`main` already had two `assertRaises(HTTPError)` assertions covering the previously-uncaught error paths — which is precisely why `27dbddea` and `ec4360f7` had to update them. Playbook generation (`test_wca_client.py:383`) and explanation (`:625`) were covered all along; the tests just asserted the old type.

The narrower, accurate version of my note: **role generation and role explanation error paths still have no assertion** on the failure type, while their playbook counterparts do. Worth closing that gap for symmetry, but it's pre-existing rather than something this PR introduced. Apologies for the overstatement.

### CI

`Unit Test and SonarCloud Scan` now passes — that was the check I asked to see complete last round, and it's the one that would have caught a regression in the retype. Everything green except the two Konflux checks, still pending.

Nothing outstanding from my side.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

de1987:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ansible_ansible-ai-connect-service&pullRequest=2110) **Quality Gate failed**
> 
> Failed conditions  [4 New issues](https://sonarcloud.io/project/issues?id=ansible_ansible-ai-connect-service&pullRequest=2110&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [87.1% Coverage on New Code](https://sonarcloud.io/component_measures?id=ansible_ansible-ai-connect-service&pullRequest=2110&metric=new_coverage&view=list) (required ≥ 90%)  [5.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=ansible_ansible-ai-connect-service&pullRequest=2110&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=ansible_ansible-ai-connect-service&pullRequest=2110)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

you still need to increase the coverage and create reusable functions to avoid the duplications

omaciel (approved):
## Re-approving — `84893502` is correct

Four-line change, `logger.error` → `logger.exception` in the new except blocks. Right call: inside an `except`, `logger.exception` logs at ERROR *and* attaches the traceback, which is what you want when converting a caught exception into a typed one.

### One observation, entirely optional

The commit makes the four new handlers consistent with each other, but the six now split three ways:

```
L467  codegen (inference)     logger.error       <- no traceback
L536  content match           (logs nothing)
L622  playbook generation     logger.exception
L718  role generation         logger.exception
L814  playbook explanation    logger.exception
L888  role explanation        logger.exception
```

Both odd ones out are pre-existing, not introduced here. But given this commit's stated purpose is exactly "use `logger.exception()` in except blocks", it's a little surprising that codegen — the highest-traffic path of the six — is the one still without a traceback, and content match logs nothing at all before raising.

Two more lines would finish the job. Equally fine to leave as-is and keep this PR scoped; just flagging so it's a choice rather than an oversight.

Tiny nit while I'm here: with `logger.exception`, the `{e}` in the message is now duplicated by the traceback. Harmless, and consistent with the existing codegen line, so no need to change it.

### CI

`Unit Test and SonarCloud Scan` has gone back to pending on the new commit, along with the two Konflux checks. Everything else green. Same ask as before — let the unit-test run finish before merging.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

omaciel (approved):
## Re-approving — and this time I've actually read `5c7fa96a`

Housekeeping first: my previous approval is still showing as current, but only by accident of timing — it landed 35 seconds after `5c7fa96a` was pushed, so GitHub counted it against a commit I hadn't looked at. Now reviewed properly.

### The gap I flagged is genuinely closed

`test_role_gen_error` and `test_role_exp_error` assert `WcaGenerationFailure` and `WcaExplanationFailure` respectively, so all four generation/explanation pipelines now have error-path coverage rather than two of four. The `@assert_call_count_metrics(..._retry_counter)` decorators also exercise the retry accounting. That's the right shape and it does what I asked for.

### One thing worth knowing: the `assertInLog` line never runs

```python
with (
    self.assertRaises(WcaGenerationFailure),
    self.assertLogs(...) as log,
):
    model_client.invoke(...)                                    # raises
    self.assertInLog("Caught retryable error after 1 tries.", log)   # unreachable
```

`invoke()` raises, so control leaves the block immediately and `assertRaises` catches it on exit. The `assertInLog` after it is dead — it never executes, and would not fail even if the log line were missing entirely.

**This is not something this PR introduced.** The pattern already appears 5 times on `main`, and these two tests copy it faithfully. So the new tests are consistent with the surrounding code, which is a reasonable thing for them to be.

But it does mean the log assertions in these tests are decorative rather than load-bearing. Worth a separate cleanup across all 7 instances — moving `assertInLog` outside the `with` block would make them real:

```python
with self.assertLogs(...) as log:
    with self.assertRaises(WcaGenerationFailure):
        model_client.invoke(...)
self.assertInLog("Caught retryable error after 1 tries.", log)
```

Not blocking, and I'd argue it belongs in its own PR rather than expanding this one further — the exception-type assertions are what actually close the gap, and those work.

### CI

`Unit Test and SonarCloud Scan` and both Konflux checks are pending again on the new commit. Same ask: let the unit-test run finish before merging, particularly since this commit is entirely tests.

---
🤖 Analysis assisted by AI (Claude Code); reviewed by @omaciel.

PR #2115, opened 2026-08-12
fix: resolve npm audit vulnerabilities across chatbot and admin portal

SiddharthRajaraman:
## Summary
- Clear blocking **Chatbot / AAP Chatbot / Admin Portal npm-audit** CI failures (`audit_level: low`) by updating npm overrides and regenerating lockfiles
- Pins / bumps: `brace-expansion` 1.1.18, `dompurify` 3.4.13, `js-yaml` 4.3.1, `nanoid` 3.3.18, `postcss` 8.5.23, `fast-uri` 4.1.2, `react-router` / `react-router-dom` 7.18.2
- Admin portal: `react-router-dom` major bump **6 → 7** (APIs in use — `createBrowserRouter`, `RouterProvider`, `BrowserRouter`, `MemoryRouter` — are compatible)

## Context
These findings currently block unrelated PRs that touch JS lockfiles (e.g. #2113). Local verification:

```bash
cd ansible_ai_connect_chatbot && npm audit --audit-level=low   # 0 vulnerabilities
cd aap_chatbot && npm audit --audit-level=low                  # 0 vulnerabilities
cd ansible_ai_connect_admin_portal && npm audit --audit-level=low  # 0 vulnerabilities
```

**Note:** This PR includes the **postcss 8.5.23** bump (CVE-2026-69153). After this merges, #2113 can rebase to Django-only (CVE-2026-15307).

## Test plan
- [ ] Chatbot - npm-audit CI green
- [ ] AAP Chatbot - npm-audit CI green
- [ ] Admin Portal - npm-audit CI green
- [ ] Chatbot / AAP Chatbot / Admin Portal Build and Test green
- [ ] Smoke admin portal routing (react-router v7 bump)

Made with [Cursor](https://cursor.com)

PR #2116, opened 2026-08-12
fix: surgically bump vulnerable npm resolutions for audit CI

SiddharthRajaraman:
## Summary
Minimal alternative to #2115 for clearing Chatbot / AAP Chatbot / Admin Portal **npm-audit** failures.

- **Surgically** bump vulnerable resolutions in each `package-lock.json`, **preserving majors** (e.g. brace-expansion `1.1.16→1.1.18`, `2.1.2→2.1.4`, `5.0.7→5.0.9` — not a global force to 1.x)
- Tighten **existing** override floors only (`postcss`, `dompurify`, `js-yaml`, `fast-uri`) where the floor still allowed a vulnerable version
- Admin portal: direct `react-router-dom` **6 → 7.18.2** (advisories have no fixed v6) + Jest `TextEncoder` polyfill

## Why not #2115-style global overrides?
A global `"brace-expansion": "1.1.18"` override forced Vitest coverage’s newer `minimatch` onto brace-expansion 1.x and broke `@vitest/coverage-v8` (`expand is not a function`). This PR avoids that by bumping each lockfile entry in-place within its own major.

## Compare with
- #2115 — broader overrides + full lockfile regenerations

## Local verification
```bash
cd ansible_ai_connect_chatbot && npm ci && npm audit --audit-level=low && npm run coverage
cd aap_chatbot && npm ci && npm audit --audit-level=low
cd ansible_ai_connect_admin_portal && npm ci && npm audit --audit-level=low
# admin router suites
npm test -- --watchAll=false --testPathPattern='(AppHeader|PageSidebar|App\\.test|AppDenied)'
```

## Test plan
- [ ] Chatbot / AAP Chatbot / Admin Portal npm-audit green
- [ ] Unit Test and SonarCloud Scan green (incl. chatbot coverage)
- [ ] Build and Test green for all three JS trees
- [ ] Smoke admin portal routing (react-router v7)

### DIFF
--- a/aap_chatbot/package-lock.json
+++ b/aap_chatbot/package-lock.json
@@ -3894,9 +3894,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.16",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.16.tgz",
-      "integrity": "sha512-IDw48K2/2kRkg9LdJxurvq3lV3aBgq0REY89duEqFRthjlPdXHKMj7EnQOXVckxzgisinf3nHfrcE2FufFLXMw==",
+      "version": "1.1.18",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+      "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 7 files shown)


omaciel (commented):
## Review — surgical audit fix (vs #2115)

This is the right version of the fix. Verified the mechanics, not just the description:

**The per-major claim holds in the lockfiles.** `ansible_ai_connect_chatbot`'s lock shows brace-expansion bumped *within each of its three lines* — `1.1.16→1.1.18`, `2.1.2→2.1.4` (both instances), `5.0.7→5.0.9` (both instances). That's exactly the per-line pinning shape that avoids #2115's failure: the vitest-coverage path keeps its 2.x/5.x brace-expansion with the `expand` named export intact, so the `@vitest/coverage-v8` breakage structurally cannot recur.

**Override deltas are floor-tightening only** — every `package.json` change raises a *pre-existing* override floor (`dompurify`, `postcss`, `fast-uri`, `js-yaml`) or bumps the admin portal's existing exact 1.x pin; no new tree-wide overrides introduced. The admin portal's global `brace-expansion 1.1.18` pin is safe there because its tree only carries the 1.x line (jest→minimatch v3).

**react-router** handling matches #2115 (admin portal 6→7.18.2 with the `TextEncoder` polyfill; chatbots converge on 7.18.2) — no intra-repo skew.

One non-blocking observation: the surgical brace-expansion/nanoid bumps have no corresponding override floors, so they live only in the lockfiles. That's fine — the semver ranges already resolve ≥ these versions on any regen, and the npm-audit gate backstops regressions — just noting it's lockfile-pinned rather than policy-pinned.

**Verdict:** prefer this over #2115 and close the other once green. The one thing standing between this and approval is the `Unit Test and SonarCloud Scan` check finishing — it's the exact check that failed on #2115, so its result here is the discriminator. Will approve when it lands green.

*Review performed with AI assistance.*

omaciel (approved):
Approving on the strength of the verified mechanics (see my earlier comment): per-major surgical bumps confirmed in all three lockfiles — including brace-expansion's three coexisting lines each staying within its own major — overrides limited to tightening pre-existing floors, and react-router converged at 7.18.2 across the apps with the required Jest polyfill.

Note: the `Unit Test and SonarCloud Scan` check (the one that failed on #2115) is still running as I approve; branch protection gates the merge on it regardless. Structurally the #2115 failure cannot recur here — the vitest coverage path keeps its 2.x/5.x brace-expansion with the `expand` named export — and local `npm run coverage` corroborates.

Suggest closing #2115 in favor of this PR once this goes green, so the two don't race each other in the merge queue.

*Review performed with AI assistance.*

de1987 (approved):

PR #2119, opened 2026-08-17
AAP-88125 AAP-88123 AAP-87486 AAP-87484 AAP-87465 AAP-87467: Bump baseline-browser-mapping and browserslist (CVE-2026-45819, CVE-2026-73089, CVE-2026-73088)

SiddharthRajaraman:
## Summary

- **Security fix:** yes (CVE-2026-45819, CVE-2026-73089, CVE-2026-73088)
- Pin transitive `baseline-browser-mapping` via npm `overrides` to `>=2.11.0`, lockfile **2.11.15**.
- Pin transitive `browserslist` via npm `overrides` to `>=4.28.7`, lockfile **4.28.8**. This single pin covers both CVE-2026-73089 (unbounded cache) and CVE-2026-73088 (prototype pollution in `normalizeStats()`).
- Both packages are build-time frontend deps (webpack/Vite/babel). No application imports. Runtime is Django plus static assets.

## Backport policy

No backport in this PR. Upstream `main` only.

## Jira

- https://issues.redhat.com/browse/AAP-88125
- https://issues.redhat.com/browse/AAP-88123
- https://issues.redhat.com/browse/AAP-87486
- https://issues.redhat.com/browse/AAP-87484
- https://issues.redhat.com/browse/AAP-87465
- https://issues.redhat.com/browse/AAP-87467

## Testing

- Lockfile versions verified: every `node_modules/baseline-browser-mapping` is `2.11.15`; every `node_modules/browserslist` is `4.28.8`.
- `package.json` `overrides` include both pins in all three frontend workspaces.
- `npm ci` was **not** run locally to avoid Node version skew; surgical lockfile patch only.

### DIFF
--- a/aap_chatbot/package-lock.json
+++ b/aap_chatbot/package-lock.json
@@ -3868,9 +3868,9 @@
       "license": "MIT"
     },
     "node_modules/baseline-browser-mapping": {
-      "version": "2.10.38",
-      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.10.38.tgz",
-      "integrity": "sha512-31/02mVB4yuQU6adKk5SlY6m+mxDwUq5KZkyYgnLrrKl7TEm1+3PyDtDBz2kOv/wxZz41GHsvV1A/u6RmiyBvw==",
+      "version": "2.11.15",
+      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.11.15.tgz",
+      "integrity": "sha512-FwMjJJ7HnyZpWe+oWxegG0fezZyBZUagI5LZEoO3GCbtbKNwRfMH9Ue5d5v01PNePBy1QSfPSDTTeVL0Hb9EzA==",
       "dev": true,
       "license": "Apache-2.0",
       "bin": {
@@ -3918,9 +3918,9 @@
       }
     },
     "node_modules/browserslist": {
-      "version": "4.28.2",
-      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.28.2.tgz",
-      "integrity": "sha512-48xSriZYYg+8qXna9kwqjIVzuQxi+KYWp2+5nCYnYKPTr0LvD89Jqk2Or5ogxz0NUMfIjhh2lIUX/LyX9B4oIg==",
+      "version": "4.28.8",
+      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.28.8.tgz",
+      "integrity": "sha512-V2NpofLblG64mfOtSgDhOJESZEGogzDMBv/q+W6oc4LXWP/q75eOXoOaaOu1EOadB9U4Bwx/e0yzbvwKH8zalA==",
       "dev": true,
       "funding": [
         {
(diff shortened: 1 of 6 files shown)


omaciel (approved):
Approving — verified safe.

CVE bump of two build-time frontend deps across all three workspaces (aap_chatbot, admin_portal, ansible_ai_connect_chatbot):
- `baseline-browser-mapping` → 2.11.15 (clears CVE-2026-45819)
- `browserslist` → 4.28.8 (clears CVE-2026-73089 + CVE-2026-73088)

Verification:
- The `>=` override floors (`>=2.11.0`, `>=4.28.7`) resolved within-major to the pinned lockfile versions 2.11.15 / 4.28.8, both ≥ the CVE fix floors. No major-version overshoot.
- Lockfile is internally consistent: browserslist 4.28.8's subtree was bumped alongside (caniuse-lite 30001809, electron-to-chromium 1.5.408, node-releases 2.0.53, update-browserslist-db 1.3.1), and its `baseline-browser-mapping: ^2.11.12` range is satisfied by 2.11.15. All 17 checks pass — the lint + Unit Test jobs run `npm ci`, which hard-fails on any package.json/lock mismatch, so the surgical patch is proven in-sync.
- Build-time/dev deps only; no app imports (runtime is Django + static assets). Coverage unchanged (95.23%), SonarCloud gate green.

Non-blocking forward note: these are unbounded `>=` overrides. Harmless here since the lockfile pins exact within-major versions and CI validated them, but if the lock is ever regenerated via `npm install`, `>=4.28.7` could later pull a future browserslist major. Worth standardizing on major-capped pins fleet-wide, but nothing to change in this PR.

🤖 Reviewed with the assistance of AI (Claude Code); conclusions verified against the diff and green CI by me.

omaciel:
Heads-up: this one has a delayed side effect, and it isn't something review would have caught.

The `"baseline-browser-mapping": ">=2.11.0"` override added here now collides with `wisdom-service.Containerfile:69`:

```
RUN cd /tmp/ansible_ai_connect_admin_portal && npx update-browserslist-db@latest
```

That tool shells out to `npm install caniuse-lite baseline-browser-mapping`, and npm refuses when the package is both overridden and directly installed:

```
npm error code EOVERRIDE
npm error Override for baseline-browser-mapping@* conflicts with direct dependency
```

`build-container` fails there, so `main` is currently unbuildable.

**Nothing was wrong with this PR when it merged.** `update-browserslist-db` only runs that install when the pinned `caniuse-lite` is stale — otherwise it exits early and never touches npm. `caniuse-lite@1.0.30001810` was published **2026-08-24 20:04Z**, six days after this merged. Before that, the override sat here harmlessly. This PR's own Konflux run passed on 08-18 with the override already in place (verified against head `c8757bd1`) — same files, same tool version, green then, red now.

So a third-party npm publish broke the build. The `npx ...@latest` has been in the Containerfile since 2024-09-09; it was simply never load-bearing until an override met it.

Two candidate fixes, either about one line:

- drop `baseline-browser-mapping` from `overrides` — it comes in transitively via `browserslist`, which this PR already pinned `>=4.28.7`, so the override may be redundant for the advisory it was added for (worth confirming against the audit before removing), or
- remove the `npx update-browserslist-db@latest` step.

No open PR does either — I checked all 39.

Full write-up with the logs and timestamps is on #2120, where this first surfaced.

Flagging rather than fixing since it's your change and the first option needs a call on whether the override is still doing work for the CVE.

---

*Investigation and this comment were produced with AI assistance; findings verified against the PipelineRun logs, the npm registry, and the repository history.*

PR #1161, opened 2024-06-27
Fix admin portal paths

manstis:
This PR does not need a corresponding Jira item. 

This PR moves the Admin Portal to a folder named consistently with the revised Python service module name.

PR #2122, opened 2026-08-25
Format two admin portal files so pre-commit passes

omaciel:
## What

Formats two admin portal files. No behaviour change — union type members are joined onto one line where they fit.

- `ansible_ai_connect_admin_portal/src/api/types.ts`
- `ansible_ai_connect_admin_portal/src/ErrorModal.tsx`

## Why

`pre-commit` has failed on **every PR since 2026-08-19**, including PRs containing no TypeScript at all — it first went red on `konflux/mintmaker/main/charset-normalizer-3.x`, a dependency-only branch.

Neither file has been touched since #1161, so nothing in this repo made it start. They are simply unformatted under the prettier the hook declares: `mirrors-prettier` v4.0.0-alpha.8 pins `prettier@4.0.0-alpha.8` via `additional_dependencies`, and that version rewrites both. CI had been passing on a cached hook environment; once it was rebuilt, the declared config started being enforced and the existing drift surfaced.

So this isn't a new problem — it's an old one that was invisible while the cache held.

## Verification

- Formatted with **exactly** the pinned version, so this is what CI produces — the diff matches the CI failure output line for line
- `prettier 3.9.6` produces **byte-identical** output on both files, so this is not an artifact of the alpha
- All **109** tracked `ts`/`tsx`/`js` files now report `All matched files use Prettier code style!`

## Worth a follow-up, not in this PR

The hook is pinned to a **prettier 4 alpha**. That's a slightly odd thing to gate CI on, and since 3.9.6 formats these files identically, moving the hook to a stable 3.x looks like a no-op for the tree. I'd rather that be a deliberate decision than a drive-by change here.

## Test plan

- [ ] `pre-commit` passes
- [ ] No behavioural diff — types only

---

*Authored with AI assistance; the reformat was produced by the pinned prettier and cross-checked against a second version.*

### DIFF
--- a/ansible_ai_connect_admin_portal/src/ErrorModal.tsx
+++ b/ansible_ai_connect_admin_portal/src/ErrorModal.tsx
@@ -17,8 +17,7 @@ interface ErrorModalProps {
 }
 
 export type HasError =
-  | { inError: false }
-  | { inError: true; message: string; detail: string };
+  { inError: false } | { inError: true; message: string; detail: string };
 export const NO_ERROR: HasError = { inError: false };
 
 export const ErrorModal = (props: ErrorModalProps) => {
--- a/ansible_ai_connect_admin_portal/src/api/types.ts
+++ b/ansible_ai_connect_admin_portal/src/api/types.ts
@@ -1,10 +1,6 @@
 // Loading statuses and payloads
 export type Status =
-  | "NOT_ASKED"
-  | "LOADING"
-  | "FAILURE"
-  | "SUCCESS"
-  | "SUCCESS_NOT_FOUND";
+  "NOT_ASKED" | "LOADING" | "FAILURE" | "SUCCESS" | "SUCCESS_NOT_FOUND";
 export type NotAsked = { status: "NOT_ASKED" };
 export type Loading = { status: "LOADING" };
 export type Failure = { status: "FAILURE"; error: Error };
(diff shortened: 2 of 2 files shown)


omaciel:
## Verification notes

Reproduced the reformat locally with the exact pinned toolchain. All claims hold.

**The pin is real.** `mirrors-prettier` at `rev: v4.0.0-alpha.8` carries `additional_dependencies: ["prettier@4.0.0-alpha.8"]` in its own `.pre-commit-hooks.yaml` — the repo's `.pre-commit-config.yaml` doesn't restate it, so it's worth saying out loud where the version actually comes from.

**Both formatters agree.** Ran each against `main`'s copies of the two files:

| prettier | output vs. this PR |
|---|---|
| 4.0.0-alpha.8 (the pinned one) | byte-identical |
| 3.9.6 (current stable) | byte-identical |

One trap worth recording: `prettier --version` from the 4.0.0-alpha.8 install **prints `3.9.6`**. The package metadata is correct (`node_modules/prettier/package.json` → `4.0.0-alpha.8`), but anyone sanity-checking via `--version` will conclude they're on stable when they aren't. I verified through the package manifest instead.

**Necessary and sufficient**, checked both directions across all 109 tracked `ts`/`tsx`/`js` files:

- this branch → `All matched files use Prettier code style!`
- `origin/main` (control) → `Code style issues found in 2 files` — exactly `ErrorModal.tsx` and `api/types.ts`, no others

So this closes the hook completely; there's no third file waiting to surface next run.

Agreed on leaving the alpha-pin question to a follow-up. Since 3.9.6 produces identical output on the whole tree, moving the hook to stable 3.x looks like a genuine no-op — but that's a config decision, not a drive-by.

**Do not merge in isolation.** `build-container` is red here (fails at 4 min) for an unrelated reason fixed by #2121, whose build passes at 17m27s. The two PRs are blocked on each other's problem; both need to land for `main` to be fully green.

---
🤖 Verification assisted by AI; both prettier versions installed and diffed locally, with a control run against main.

SiddharthRajaraman (approved):
lgtm

PR #2124, opened 2026-08-28
Add idna>=3.15 constraint floor

omaciel:
## What

Adds `idna>=3.15` to `[tool.uv] constraint-dependencies`. `uv.lock` gains the manifest entry; the resolved version is already 3.15, so `requirements.txt` is unchanged and no package versions move.

## Why

`stable-2.7` has carried this floor since d828b76e; `main` never got it. Recording it keeps the constraint table consistent across release lines and prevents `uv lock` / `uv export` from regressing below 3.15.

## Scope

`uv_export.yml` reproduced locally (`uv lock` + `uv export`, `git diff --exit-code` clean).

Assisted-by: Claude Code — drafted by one model, independently validated by a second before push.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ constraint-dependencies = [
   'pyjwt>=2.13.0',
   'ujson>=5.13.0',
   'msgpack>=1.2.1',
+  'idna>=3.15',
   'twisted>=26.4.0',
   'pygments>=2.20.0',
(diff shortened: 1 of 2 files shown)


hasys (approved):
Looks good to me, thank you @omaciel

SiddharthRajaraman (approved):
lgtm

PR #2125, opened 2026-08-28
ci: make the SonarCloud skip guard actually skip the scan

omaciel:
## What

Turn the `workflow_run` staleness check in `sonar-pr.yaml` into a real skip: it now records `skip=true` as a step output and every subsequent step is gated on it. Adds a fail-fast step that refuses to scan when `PR_NUMBER` / `PR_BASE` / `PR_HEAD` are empty.

## Why

The guard ends with `exit 0`, which only ends *that step* — the job keeps going. When it fires, `PR_NUMBER`/`PR_BASE`/`PR_HEAD` are never exported and `gh pr checkout` never runs, so the scan is invoked as:

```
-Dsonar.pullrequest.key= -Dsonar.pullrequest.branch= -Dsonar.pullrequest.base=
```

With an empty pull-request key the scanner falls back to **branch** mode and publishes an analysis of the `workflow_run` checkout as `main`, stamped with the stale PR SHA via `sonar.scm.revision`. It then evaluates main's quality gate and exits 3.

Seen in [run 33199399627](https://github.com/ansible/ansible-ai-connect-service/actions/runs/33199399627): the guard correctly logged `Skipping SonarCloud analysis - PR has been updated since workflow started`, then scanned anyway and failed on main's gate (`new_violations: 2`). The PR that triggered it was never analyzed, and main's Sonar state was overwritten from an unrelated checkout.

## Scope

`.github/workflows/sonar-pr.yaml` only. No change to what gets scanned on a healthy run — the added `if:` conditions are all no-ops when the guard passes.

## Note

#2042 (Dependabot, `sonarqube-scan-action` 4 → 8) edits the same `uses:` line this PR adds an `if:` above, so whichever lands second will need a rebase. That bump is worth landing regardless — the action currently emits an unsupported-version/security warning on every run.

### DIFF
--- a/.github/workflows/sonar-pr.yaml
+++ b/.github/workflows/sonar-pr.yaml
@@ -45,6 +45,7 @@ jobs:
           run-id: ${{ github.event.workflow_run.id }}
 
       - name: Extract PR metadata and validate workflow_run
+        id: pr_meta
         env:
           REPO_NAME: ${{ github.event.repository.full_name }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


SiddharthRajaraman (approved):
lgtm

PR #2134, opened 2026-08-30
deps(deps): bump propcache from 0.4.1 to 0.5.2 (via uv)

omaciel:
Supersedes #2065.

### Why a new PR instead of #2065

Dependabot PR #2065 edited only the generated `requirements.txt`. In this repo `requirements.txt` is an artifact exported from `uv.lock` + `pyproject.toml`, and the `wisdom-service - uv-export` gate regenerates it:

```sh
uv lock
uv export --format requirements-txt --no-hashes --no-emit-project -o requirements.txt
git diff --exit-code -- uv.lock requirements.txt
```

Because #2065 never touched `uv.lock` (which still pinned `propcache==0.4.1`), `uv lock` kept that still-valid pin and the export reverted the bump — so #2065 could never pass CI, and a rebase/recreate wouldn't help.

### This PR

Re-resolves propcache through uv so both files move together:

```sh
uv lock --upgrade-package propcache
uv export --format requirements-txt --no-hashes --no-emit-project -o requirements.txt
```

Result: `propcache 0.4.1 -> 0.5.2` in both `uv.lock` and `requirements.txt`. `propcache` is a transitive dep (`via aiohttp, yarl`); nothing caps it and `0.5.2` is compatible with `requires-python = ">=3.12"`.


hasys (approved):
Looks good to me

SiddharthRajaraman (approved):
lgtm

PR #2150, opened 2026-09-03
deps: batch Dependabot updates (actions, npm, Python)

SiddharthRajaraman:
## Summary

Batches open Dependabot PRs into a single branch to avoid rebase conflicts between overlapping lockfile/workflow updates.

Supersedes: #2138, #2139, #2140, #2141, #2143, #2145, #2146, #2067

## Major version bumps

| Package | From → To | Area | Original PR |
|---------|-----------|------|-------------|
| `actions/checkout` | 3/4/`master` → **7** | GitHub Actions (most workflows; starting tag varies per file) | #2139 |
| `actions/setup-python` | 4 → **7** | GitHub Actions (8 workflows) | #2138 |
| `oke-py/npm-audit-action` | 4 → **5** | GitHub Actions (3 npm audit workflows) | #2140 |
| `css-loader` | 6 → **7** | admin portal (dev) | #2145 |
| `@testing-library/react` | 14 → **16** | admin portal (dev) | #2146 |
| `peter-evans/create-pull-request` | 6 → **8** | `ui_release_aap_chatbot.yml` (pairs with checkout v7) | review follow-up |

## Constraint / policy changes

| Package | Change | Why |
|---------|--------|-----|
| `black` | `~=26.3.1` → `~=26.5.1` | Not just a patch bump: `~=26.3.1` means `<26.4.0`, so this is a deliberate constraint widening to admit 26.5.1 and keep the `~=` convention. Original PR #2067 used a compound range; this PR matches sibling specifiers instead. |
| `@testing-library/dom` | declared `^10.4.0` | Required peer of `@testing-library/react` 16; was only auto-installed. |

## Security (not version bumps)

These landed in the CI-fix commit / review follow-up, not in the eight superseded Dependabot PRs.

| Change | What | Why |
|--------|------|-----|
| pip-audit ignore `CVE-2026-73228`, `CVE-2026-73229` | Suppress DRF advisories until 3.17.2 is installable | Fixed in DRF 3.17.2, but blocked by **two** caps: this repo's `djangorestframework~=3.15.2` **and** django-ansible-base's `djangorestframework<3.16`. Tracker: [AAP-91663](https://issues.redhat.com/browse/AAP-91663) (review by 2026-12-03). |
| `qs` override `>=6.16.0 <7` | Floor + major cap | 6.16.0 is the first patched version for the open qs advisories; `<7` keeps express/body-parser on the 6.x parse API. |
| `postcss-selector-parser@^6`: `>=6.1.3 <7` / `@^7`: `>=7.1.3` | Range-keyed overrides | GHSA-w9m9-85wc-3x92 has two disjoint windows; this tree has both `^6` and `^7` consumers. |
| `body-parser` override `>=1.20.6 <2` | Cap the existing floor | Unbounded `>=1.20.6` had resolved to 2.3.0 while express expects 1.x. |

## Not included

| Package | Original PR | Reason |
|---------|-------------|--------|
| `referencing` 0.36.2 → 0.37.0 | #2142 | **Blocked** — `django-ansible-base` requires `referencing<0.37.0`; `uv lock` is unsatisfiable at 0.37.0 |
| `propcache` 0.4.1 → 0.5.2 | #2065 | **Already on `main`** at 0.5.2 — no change needed |

## Scope / backport

Landing as one squash. Intended to stay on `main` unless a specific piece is needed elsewhere:

- **CI actions majors** (`checkout`/`setup-python`/`npm-audit-action`) — only needed on a stable branch if that branch still runs those workflows.
- **Admin-portal npm majors + overrides** — frontend-only; no Python runtime impact.
- **DRF pip-audit ignore** — needed wherever pip-audit runs against the same DRF pin.
- **Python lock (`ansible-compat`, `zope-interface`, `black`)** — lockfile-heavy; expect conflicts if cherry-picked.

## Untested on this PR

These workflow hunks are not exercised by pull_request checks (first real run is after merge / on dispatch):

- `sonar-pr.yaml` (`workflow_run` uses the default-branch copy)
- `backport.yml` (`pull_request_target`)
- `sync-openapi-spec.yml`, `security_report.yml` (push to main)
- `ui_publish_aap_chatbot.yml`, `ui_release_aap_chatbot.yml` (`workflow_dispatch`)

## Notes

- `uv.lock` and `requirements.txt` regenerated via `uv export` (source of truth is `pyproject.toml` + `uv.lock`).
- Individual Dependabot PRs were closed in favor of this batch.

## Test plan

- [ ] CI green (unit tests, selftest, npm audit, Konflux)
- [ ] Admin portal build/tests pass with `@testing-library/react` 16 and `css-loader` 7
- [ ] Watch the first post-merge push to main for the untested workflows above


omaciel (commented):
Nice batching — the supersede list and the "Not included" table with reasons make this much easier to review than eight separate PRs would have been. Verified the mechanical parts and they hold up; two things I'd like changed before approving, and both are small.

## Two changes here are not dependency bumps, and aren't in the body

**1. `pip_audit.yml` adds two CVE suppressions:**

```
# DRF 3.17.2 fixes CVE-2026-73228/73229, but django-ansible-base pins
# djangorestframework<3.16 — no satisfiable upgrade path yet.
CVE-2026-73228
CVE-2026-73229
```

**2. `package.json` adds two new `overrides`:** `qs >=6.16.0` and `postcss-selector-parser >=6.1.3`.

Both are security decisions, and neither appears in the three tables above. A reviewer reading the body would approve this believing it's version bumps only. Please add a "Security" section listing them — the suppressions especially.

**The suppression rationale is correct, and I checked it rather than taking it on faith:** `django-ansible-base 2026.1.26` really does declare `djangorestframework<3.16` in its PyPI metadata, and both CVEs are first patched in 3.17.2, so there is genuinely no satisfiable upgrade. No objection to the suppression itself.

Two follow-ups on it though:
- **It has no review-by date or tracking ticket.** Ignored CVEs without an expiry become permanent silently. The pattern used in the sibling repos is a dated justification plus a Jira link.
- **There are two constraints, not one.** This repo's own `pyproject.toml` also pins `djangorestframework~=3.15.2`. When django-ansible-base relaxes its cap, that local pin still blocks 3.17.2 — so whoever revisits this needs to move both, and the comment currently only names the dab one.

## `postcss-selector-parser: ">=6.1.3"` leaves the second vulnerable window open

GHSA-w9m9-85wc-3x92 is **non-contiguous** — it has two disjoint windows:

| range | first patched |
|---|---|
| `>=6.1.0 <6.1.3` | 6.1.3 |
| `>=7.1.0 <7.1.3` | 7.1.3 |

An unbounded `>=6.1.3` closes the first and leaves the second wide open: 7.1.0–7.1.2 satisfy the floor and are vulnerable.

This is **latent, not live** — today it resolves to 6.1.4 at the root and 7.0.0 under `postcss-modules-local-by-default` / `postcss-modules-scope`, all outside both windows. But the declared range permits the bad window on any future resolution, and `npm audit` won't catch it because audit reads the resolved tree, not the range.

Suggest `">=6.1.3 <7.1.0 || >=7.1.3"`.

`qs: ">=6.16.0"` is correct as written — 6.16.0 is the first patched version for all four open qs advisories, including the widest (`>=2.2.5 <6.16.0`).

## Smaller notes

- **`black` is a constraint widening, not just a bump:** `~=26.3.1` means `<26.4.0`; the new `>=26.3.1,<26.6.0` is a deliberate policy relaxation to admit 26.5.1. Worth naming as such rather than sitting in the "minor/patch" table.
- **`actions/checkout` is 3 → 7 in the body, but `pip_audit.yml` goes v4 → v7** — the starting version varies per workflow. Cosmetic.

## Verified good

- **Lock delta is coherent:** 6 changed, 0 added, 6 removed. Every removal (`deep-equal`, `es-get-iterator`, `is-arguments`, `object-is`, and the nested `@testing-library/dom` / `aria-query`) is a transitive dep that `@testing-library/react` 16 dropped — which accounts for the -121 net lines rather than anything unexpected being pulled out.
- **The `oke-py/npm-audit-action` 4 → 5 major is actually exercised.** I checked the path filters, since two of the three npm-audit workflows are scoped to `ansible_ai_connect_chatbot/**` and `aap_chatbot/**` and correctly don't run here. `Admin Portal - npm-audit` does match and passes on the current head, so the major is covered.
- All other checks green, including Konflux, EC, CodeQL and SonarCloud.

Happy to approve once the `postcss-selector-parser` floor is bounded and the two security changes are called out in the body.

omaciel (commented):
**Follow-up to my review above (5105015955), not a replacement.** The asks there still stand — list the suppression and the two overrides in the body, name both DRF caps, add a tracker/expiry. This pass goes a level deeper and adds a few things; it also refines the `postcss-selector-parser` suggestion, so please take the inline note on line 205 as superseding the range I proposed earlier.

Thanks again for pulling these together, @SiddharthRajaraman — batching overlapping lockfile/workflow bumps is a sensible way to dodge the rebase treadmill. The mechanical core holds up under closer inspection: every +/− line from the eight superseded PRs is present and unaltered, `requirements.txt` is byte-identical to `uv export` on the new lock, the `referencing` block is real (django-ansible-base's `api-documentation` extra pins `<0.37.0`), `propcache` is indeed already 0.5.2, and the resolved-package diffs in both lockfiles contain only what's claimed.

**One addition to the before-merge list from my earlier review:**

- **`ui_release_aap_chatbot.yml:16`** — `checkout@v7` alongside `peter-evans/create-pull-request@v6` is a documented incompatibility (credentials moved out of `.git/config` in checkout v6+). That workflow is `workflow_dispatch`-only, so the green checks here never exercised it; it would surface on the next manual release run. Details inline.

The rest are smaller and non-blocking: a forward-compat cap on `qs` (the security floor itself is right), a missing peer declaration for `@testing-library/dom`, and some pinning consistency.

Two structural thoughts, offered as suggestions:

- **Test coverage of the workflow hunks.** Several of the changed workflows didn't run on this PR — `sonar-pr.yaml` (`workflow_run` executes the default-branch copy), `backport.yml` (`pull_request_target`), `sync-openapi-spec.yml` and `security_report.yml` (push to main), and both `ui_publish`/`ui_release` (dispatch). The first push to main after merge is the real test for those, so it'd be worth a line in the body naming them and keeping an eye on that run.
- **Squash granularity.** The repo is squash-merge-only and `backport.yml` cherry-picks the single squash. This one carries five majors, two security floors, a suppression, and two regenerated locks; a `backport/*` label will conflict on both lockfiles, and a css-loader-7 or testing-library-16 regression would bisect to a commit with nine unrelated changes. If it's not too late, splitting into CI-actions / npm / security-floors would make each half independently revertible and backportable. If it's easier to land as-is, a **Scope** section saying which pieces need to follow to other branches (and which don't) would go a long way.

Happy to help with any of it.

---
*Review assisted by AI (Claude Code); all findings reviewed by me before posting.*

.github/workflows/ui_release_aap_chatbot.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
```
omaciel, line 16: **Please pair this with a `create-pull-request` bump — v6 (line 46) is incompatible with checkout ≥ v6.**

From checkout v6 onward the token is stored via an `includeIf` in `$RUNNER_TEMP` rather than as an `extraheader` in `.git/config`. `peter-evans/create-pull-request@v6` (tag last moved 2024-06) still looks for the old location; it can't find/unset the header, ends up sending two `Authorization` headers, and `git remote prune origin` fails with `Duplicate header: Authorization` / HTTP 400, exit 128 — no release PR. This is tracked in peter-evans/create-pull-request#4228 and fixed in v7.0.9.

Because this workflow is `workflow_dispatch`-only, none of the green checks on this PR ran it — it would surface on the next manual "AAP Chatbot - Release" run.

Two ways to fix, either is fine:
- bump `peter-evans/create-pull-request` to `v8` (current) in this PR, or
- add `persist-credentials: false` to this checkout step (the workflow already passes its own token to create-pull-request).

.github/workflows/pip_audit.yml
```diff
@@ -61,3 +61,7 @@ jobs:
             # Disputed — pyjwt 2.12.1 is latest, no fix version available
             # See: https://osv.dev/vulnerability/PYSEC-2025-183
             PYSEC-2025-183
+            # DRF 3.17.2 fixes CVE-2026-73228/73229, but django-ansible-base pins
+            # djangorestframework<3.16 — no satisfiable upgrade path yet.
+            CVE-2026-73228
+            CVE-2026-73229
```
omaciel: **Suggestion: give this suppression the same shape as the three above it, and list it in the PR body.**

The ignore itself is defensible — there's no satisfiable upgrade path today. A few things would make it safe to leave in place:

1. **The comment is half the story.** `pyproject.toml:26` pins `djangorestframework~=3.15.2` *directly*, so lifting django-ansible-base's `<3.16` cap on its own wouldn't unblock this. Someone following the comment as written would bump d-a-b, remove the ignore, and still be red. Worth naming both caps.
2. **No `See:` link.** Each of the three entries above carries an advisory URL; matching that (e.g. the GHSA or osv.dev links for CVE-2026-73228 / CVE-2026-73229, and the minimal fixed version 3.17.2) keeps the list self-documenting.
3. **No unblock condition.** A one-liner like `# Remove once both caps allow DRF >= 3.17.2` turns this from a permanent exception into a tracked one.

Something like:

```yaml
            # DRF 3.17.2 fixes CVE-2026-73228/73229. Blocked by two caps: our own
            # djangorestframework~=3.15.2 in pyproject.toml, and django-ansible-base's
            # djangorestframework<3.16. Remove once both allow >=3.17.2.
            # See: https://osv.dev/vulnerability/CVE-2026-73228
            # See: https://osv.dev/vulnerability/CVE-2026-73229
            CVE-2026-73228
            CVE-2026-73229
```

And since this came in via the "fix CI" commit rather than any of the eight listed PRs, a row in the body (alongside the `qs` / `postcss-selector-parser` overrides) would keep the description complete.

ansible_ai_connect_admin_portal/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^6.1.4",
-    "@testing-library/react": "^14.1.0",
+    "@testing-library/react": "^16.3.3",
```
omaciel, line 28: **Suggestion: declare `@testing-library/dom` explicitly.**

`@testing-library/react` 16 moved `@testing-library/dom` to a *required* peer (`^10`). Right now it only exists in the lock as an auto-installed `"peer": true` entry at 10.4.0 — nothing in `package.json` asks for it.

That's fragile in two ways: this `package.json` currently only resolves under `--legacy-peer-deps` (the pre-existing `@patternfly/patternfly ^5` vs `@ansible/ansible-ui-framework` peer `^6` conflict), and in that mode a lock regeneration won't auto-install peers — so the next `npm ci && npm run test` in `code_coverage.yml` would fail with `Cannot find module '@testing-library/dom'`. And Dependabot can't bump a package that isn't declared.

```json
    "@testing-library/dom": "^10.4.0",
```

in `devDependencies` fixes both.

ansible_ai_connect_admin_portal/package.json
```diff
@@ -200,6 +200,8 @@
     "http-proxy-middleware": "2.0.10",
     "body-parser": ">=1.20.6",
     "js-yaml": "4.3.1",
-    "browserslist": ">=4.28.7"
+    "browserslist": ">=4.28.7",
+    "qs": ">=6.16.0",
```
omaciel: **Suggestion (forward-compat, not a security gap): cap the floor — `">=6.16.0 <7"`.**

To be clear up front: as I said in my earlier review, the floor itself is right — express 4.x hard-pins `qs ~6.15.1`, and 6.16.0 is the first patched version for all four open qs advisories. Nothing to fix on the security side.

The only change I'd make is bounding the open-ended range. When `qs` 7.0.0 ships, the next lock refresh resolves this override to 7.x for express and body-parser under webpack-dev-server, which expect the 6.x parse API, and `npm start` breaks for a reason nobody can trace to an advisory. There's already an example of that failure mode three lines up — `"body-parser": ">=1.20.6"` resolved to body-parser **2.3.0** while express declares `~1.20.5`. Worth capping that one too while you're here, if you don't mind the extra scope.

ansible_ai_connect_admin_portal/package.json
```diff
@@ -200,6 +200,8 @@
...
+    "postcss-selector-parser": ">=6.1.3"
```
omaciel: **Refining my earlier suggestion — please take this one instead of the `>=6.1.3 <7.1.0 || >=7.1.3` range I proposed in the first review.**

GHSA-w9m9-85wc-3x92 has two disjoint windows (`>=6.1.0 <6.1.3` and `>=7.1.0 <7.1.3`), and this tree has consumers on both majors: ~20 packages declaring `^6` at the root (cssnano-5 / postcss-preset-env-7 plugins), and css-loader's `postcss-modules-*` declaring `^7.0.0`. That combination is what makes a single bare-key override awkward:

- `">=6.1.3"` (as written) leaves the 7.1.0–7.1.2 window open for the `^7` consumers.
- `">=6.1.3 <7.1.0 || >=7.1.3"` (my earlier suggestion) closes both windows, but a bare-key override resolves to the *highest* match — 7.1.6 — so on the next fresh resolve it would move all the `^6` consumers across a major they were never tested against. My mistake for not thinking that through.

Range-keyed overrides fix both at once, because each only applies to the declarations it matches:

```json
"postcss-selector-parser@^6": ">=6.1.3 <7",
"postcss-selector-parser@^7": ">=7.1.3"
```

For what it's worth, I also checked that a plain `npm update postcss-selector-parser --package-lock-only` on main's lock already moves the root copy to 6.1.4 with no override at all — so if you'd rather keep `overrides` minimal, dropping this entry and committing that lock is also a clean answer; the `^7` floor is then the only thing you'd lose.

pyproject.toml
```diff
@@ -60,7 +60,7 @@ classifiers = [
 
 [project.optional-dependencies]
 dev = [
-  'black~=26.3.1',
+  'black>=26.3.1,<26.6.0',
```
omaciel: **Minor: consider `black~=26.5.1` to match the file's convention.**

This compound range is what Dependabot generated in #2067, so no fault of the batch — but it's now the only non-`~=` specifier among its dev siblings, and `<26.6.0` is an arbitrary CalVer cap that blocks the very next monthly black. Dependabot would then have to rewrite a compound range instead of bumping a `~=` operand.

`'black~=26.5.1'` keeps the shape consistent. Separately — and pre-existing, so no action needed here — `.pre-commit-config.yaml` still pins black 25.9.0, so `pip install .[dev]` and the hook format with different rules. Might be worth aligning in a follow-up.

.github/workflows/sonar-pr.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       pull-requests: read
       actions: read
     steps:
-      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
omaciel, line 35: The SHA is correct (`3d3c42e5…` == `v7.0.1` == current `v7`), and the default checkout is exempt from v7's new fork-PR guard, so this is fine as written — just flagging that it wasn't actually exercised here: `workflow_run` jobs always run the default-branch copy of the workflow, and the SonarCloud runs for this PR report `main`'s SHA. It'll get its first real run on the next push to main after merge. Same applies to `backport.yml`, `sync-openapi-spec.yml`, `security_report.yml`, and both `ui_publish`/`ui_release` workflows. Not a blocker, just worth knowing which hunks are still untested.

.github/workflows/security_report.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v7
```
omaciel, line 12: **Minor: while you're here, line 10's `actions/checkout@master` is worth pinning too.**

After this PR `actions/checkout` has four pinning strategies in the tree: SHA in `sonar-pr.yaml`, `@v7` in 17 files, `@v4` in `dependabot_uv_export.yml:34` (added after Dependabot opened #2139, so it missed the sweep), and `@master` here. `master` ships the next behaviour change (v6 moved credentials, v7 refused fork checkouts) straight into a push-to-main workflow with no PR to review or bisect.

Bumping both stragglers to the same `v7.0.1` in this PR would finish the job and save a follow-up Dependabot one-liner next week.

.github/workflows/ui_release_aap_chatbot.yml
SiddharthRajaraman: Bumped `peter-evans/create-pull-request` from v6 → v8 in this workflow (commit 58994220) so it finds credentials the way checkout v7 stores them. Left `persist-credentials` at the default since v8 is the actual fix for peter-evans/create-pull-request#4228.

.github/workflows/pip_audit.yml
SiddharthRajaraman: Updated the ignore comment to name both caps, added the OSV links and an unblock condition, and opened [AAP-91663](https://issues.redhat.com/browse/AAP-91663) (review by 2026-12-03) so this doesn't become a silent permanent exception. Also listed it in a new Security section in the PR body.

ansible_ai_connect_admin_portal/package.json
SiddharthRajaraman: Declared `@testing-library/dom` `^10.4.0` in `devDependencies` so the required peer is explicit and Dependabot can bump it. Lock now has it as a real dep at 10.4.0 rather than only a `peer: true` install.

ansible_ai_connect_admin_portal/package.json
SiddharthRajaraman: Capped to `>=6.16.0 <7`. Also bounded the existing `body-parser` floor to `>=1.20.6 <2` — the lock had resolved that one to 2.3.0, and it now sits at 1.20.6 under express.

ansible_ai_connect_admin_portal/package.json
SiddharthRajaraman: Switched to the range-keyed overrides you suggested:

```json
"postcss-selector-parser@^6": ">=6.1.3 <7",
"postcss-selector-parser@^7": ">=7.1.3"
```

Lock now has 6.1.4 for the `^6` consumers and 7.1.6 for the `^7` ones, so both GHSA windows are closed without dragging the cssnano/postcss-preset-env plugins across a major.

pyproject.toml
SiddharthRajaraman: Changed to `black~=26.5.1` so it matches the other `~=` specifiers. Left the pre-commit 25.9.0 pin alone as you noted — that's a follow-up, not this batch. Called out the constraint widening (not a patch bump) in the PR body.

.github/workflows/sonar-pr.yaml
SiddharthRajaraman: Acknowledged — listed `sonar-pr.yaml`, `backport.yml`, `sync-openapi-spec.yml`, `security_report.yml`, and both `ui_publish`/`ui_release` in an "Untested on this PR" section so we watch the first post-merge/dispatch runs. No code change here.

.github/workflows/security_report.yml
SiddharthRajaraman: Pinned both stragglers to `@v7` to match the rest of the sweep: `security_report.yml` (`@master` → `@v7`) and `dependabot_uv_export.yml` (`@v4` → `@v7`). Used the floating `v7` tag rather than the SHA so Dependabot can keep them in lockstep with the other 17 workflows.

SiddharthRajaraman:
Addressed the review comments in 58994220 and updated the PR body.

- **Security section** in the description for the DRF pip-audit ignore and the npm overrides; tracker is [AAP-91663](https://issues.redhat.com/browse/AAP-91663) (review by 2026-12-03).
- **Range-keyed** `postcss-selector-parser` overrides, `qs`/`body-parser` major caps, and an explicit `@testing-library/dom` peer.
- **`create-pull-request` v8** next to checkout v7; leftover `checkout@master` / `@v4` pinned to `@v7`.
- **`black~=26.5.1`** instead of the compound range.
- **Scope / untested-workflows** notes added rather than splitting the squash.

Inline replies are on the individual threads.

omaciel (approved):
LGTM — thanks @SiddharthRajaraman, this is a thorough turnaround. Everything from both earlier reviews is addressed, and I re-verified each item by command rather than taking the replies on faith. CI is fully green on `58994220` (Konflux, Enterprise Contract, CodeQL, SonarCloud, pre-commit, all lint/selftest/unit jobs).

## Resolution

| Ask | Evidence |
|---|---|
| `create-pull-request` v6 → v8 | `ui_release_aap_chatbot.yml:46` — all nine inputs the workflow passes exist in v8; no renames to worry about |
| DRF ignore: both caps, links, unblock condition, tracker | `pip_audit.yml:64-70` — names our own `~=3.15.2` *and* django-ansible-base's `<3.16`, two OSV links, an explicit removal condition, and AAP-91663 with a review-by date |
| `@testing-library/dom` declared | `package.json:26`; the lock entry flipped from `"peer": true` to a real dev dependency at 10.4.0 |
| `qs` capped | `">=6.16.0 <7"`, resolves 6.16.0 |
| `body-parser` capped | `">=1.20.6 <2"`; the lock moved **2.3.0 → 1.20.6**, now correctly nested under express |
| Range-keyed `postcss-selector-parser` | `@^6` → 6.1.4 at the root, `@^7` → **7.0.0 → 7.1.6** nested. Both GHSA-w9m9-85wc-3x92 windows closed, and the `^6` consumers stayed on 6.x |
| `black~=26.5.1` | `pyproject.toml`, `uv.lock` (26.5.1) and `requirements.txt:48` all agree |
| Action pin stragglers | `security_report.yml` `@master` → `@v7`, `dependabot_uv_export.yml` `@v4` → `@v7`; a tree-wide grep now shows every `actions/checkout` and `actions/setup-python` on v7 (one deliberately SHA-pinned at v7.0.1) |
| Body: Security / Scope / Untested sections, black as a widening | All four present and accurate |

## Lockfile verification

I diffed the **resolved package set** rather than line counts. 15 entries changed, every one explained:

- the `@testing-library/dom` peer→real flip plus its five transitives (`@types/aria-query`, `dom-accessibility-api`, `lz-string`, `pretty-format`, `react-is`);
- the body-parser 2→1 cascade: `raw-body` 3.0.2→2.5.3, `iconv-lite` 0.7.3→0.4.24, and the 2.x copies of `content-type` / `media-typer` / `mime-types` / `type-is` dropped.

Nothing unexplained, nothing unrelated. `npm audit` reports **0 findings**, and because a green audit only reflects the resolved tree, I also queried OSV directly for each version the cascade landed on — `iconv-lite@0.4.24`, `raw-body@2.5.3`, `body-parser@1.20.6`, `qs@6.16.0`, `postcss-selector-parser@6.1.4` and `@7.1.6` — **zero advisories each**, so the downgrades are not a security regression. `npm ci --dry-run` also succeeds cleanly on the new lock with no `--legacy-peer-deps`, which resolves a concern I raised earlier; that one was overstated on my part.

## One thing worth knowing before the next release run (no change requested)

`ui_release_aap_chatbot.yml:55` sets `sign-commits: true`, added back in #1616. That input **does not exist in create-pull-request v6** — it was introduced in v7 — so it has been an unrecognized no-op ever since, and this bump is what finally makes commit signing actually happen. Nice side effect, but two consequences to expect:

- the commit identity becomes `github-actions[bot]`, superseding the `committer: GitHub <noreply@github.com>` the workflow sets;
- API-created commits need `contents: write`, and this file has no `permissions:` block, so it inherits the repository default.

Since the workflow is `workflow_dispatch`-only, none of that is exercised by CI — worth a glance on the first release run, along with the other workflows in your "Untested on this PR" list.

Approving. Thanks for bounding the overrides properly and for documenting the security changes in the body — that section makes this much easier for whoever picks it up on another branch.

---
*Review assisted by AI (Claude Code); all findings reviewed by me before posting.*

PR #2164, opened 2026-09-04
Re-resolve fflate in the lockfile [lship]

omaciel:
## What

Adds an `overrides` floor for `fflate` in both chatbot packages and re-resolves the lockfiles onto it.

```json
"fflate": ">=0.4.9 <0.5.0"
```

| package | from | to |
|---|---|---|
| `aap_chatbot` | 0.4.8 | 0.4.9 |
| `ansible_ai_connect_chatbot` | 0.4.8 | 0.4.9 |

## Why the cap matters

GHSA-px8p-9vwx-vf98 / CVE-2026-45820 is **non-contiguous** — five disjoint vulnerable windows:

| vulnerable | first patched |
|---|---|
| `>=0.4.5 <0.4.9` | 0.4.9 |
| `>=0.5.0 <0.5.4` | 0.5.4 |
| `>=0.6.0 <0.6.11` | 0.6.11 |
| `>=0.7.0 <0.7.5` | 0.7.5 |
| `>=0.8.0 <0.8.3` | 0.8.3 |

An unbounded `>=0.4.9` would satisfy the audit today while re-admitting four later vulnerable ranges on any future resolution. The `<0.5.0` cap is load-bearing, not tidiness.

## Scope

Lockfile delta is one entry per package — `fflate` 0.4.8 → 0.4.9, nothing added or removed. `fflate` reaches the tree via `posthog-js` as a production dependency, so this ships in both chatbot bundles.

## Deferred, not addressed here

- `setuptools` 80.9.0 → 83.0.0 (CVE-2026-59890) — major bump
- `multer` (`ansible_ai_connect_admin_portal`) 1.4.5-lts.1 → 2.2.0 — major bump, **8 CVEs**; the largest open pile and worth its own ticket rather than recurring in every lship body

> **Removed from this list: `pydantic`.** lship reported `2.* -> 1.10.13` with "unreadable current". That is wrong and would be harmful if actioned. The repo is on `pydantic==2.12.5`, and neither cited CVE affects it — CVE-2021-29510 covers only 1.x ranges, and CVE-2024-3772 is fixed in **2.4.0**, which 2.12.5 is well past. lship took the `first_patched_version` from the *1.x* branch of a multi-branch advisory and proposed a two-major downgrade to fix nothing.

## Test plan
- [ ] CI passes
- [ ] Both `npm-audit` workflows pass (they are the gate that proves the floor took)

### DIFF
--- a/aap_chatbot/package-lock.json
+++ b/aap_chatbot/package-lock.json
@@ -3938,11 +3938,11 @@
       ],
       "license": "MIT",
       "dependencies": {
+        "baseline-browser-mapping": "^2.11.12",
         "caniuse-lite": "^1.0.30001809",
-        "node-releases": "^2.0.53",
         "electron-to-chromium": "^1.5.402",
-        "update-browserslist-db": "^1.3.0",
-        "baseline-browser-mapping": "^2.11.12"
+        "node-releases": "^2.0.53",
+        "update-browserslist-db": "^1.3.0"
       },
       "bin": {
         "browserslist": "cli.js"
@@ -6175,9 +6175,9 @@
       }
     },
     "node_modules/fflate": {
-      "version": "0.4.8",
-      "resolved": "https://registry.npmjs.org/fflate/-/fflate-0.4.8.tgz",
-      "integrity": "sha512-FJqqoDBR00Mdj9ppamLa/Y7vxm+PRmNWA67N846RvsoYVMKB4q3y/de5PA7gUmRMYK/8CMz2GDZQmCRN1wBcWA==",
+      "version": "0.4.9",
+      "resolved": "https://registry.npmjs.org/fflate/-/fflate-0.4.9.tgz",
+      "integrity": "sha512-zdxgIEddhfsyCaWpJ2SdXEP8ZMrKJ6+5jl4OupODcywU0IhRk6gdXuVGcPICyfx2H97hVK7xmJtRLPjkxAX8Vw==",
       "license": "MIT"
     },
     "node_modules/file-entry-cache": {
--- a/aap_chatbot/package.json
+++ b/aap_chatbot/package.json
@@ -86,6 +86,7 @@
     "fast-uri": ">=4.1.3",
     "js-cookie": ">=3.0.6",
     "baseline-browser-mapping": ">=2.11.0",
-    "browserslist": ">=4.28.7"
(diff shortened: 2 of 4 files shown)


omaciel:
Reviewed but deliberately not approving — @omaciel authored this, so it needs a second pair of eyes rather than mine.

**The change is correct, and the cap is the part worth defending in review.** GHSA-px8p-9vwx-vf98 is non-contiguous with five disjoint windows (`>=0.4.5 <0.4.9`, `>=0.5.0 <0.5.4`, `>=0.6.0 <0.6.11`, `>=0.7.0 <0.7.5`, `>=0.8.0 <0.8.3`). An unbounded `>=0.4.9` would pass the audit today and re-admit four later vulnerable ranges on any future resolution. `>=0.4.9 <0.5.0` is right.

Verified:

- lockfile delta is exactly one entry per package — `fflate` 0.4.8 → 0.4.9, **0 added, 0 removed** in both
- `fflate` reaches the tree via `posthog-js` with `dev` unset, so this is a production dependency and the fix ships
- both `AAP Chatbot - npm-audit` and `Chatbot - npm-audit` ran and passed, which is the gate that proves the floor actually took

**I edited the PR body**, for two reasons worth stating so the change is auditable:

1. It claimed *"2 packages that no `package.json` declares. There is no floor to raise, so the lockfile was re-resolved with a targeted upgrade."* The diff does the opposite and better — it adds a durable `overrides` floor to both `package.json` files. As written, a reviewer would have approved this believing the fix evaporates on the next unrelated `npm install`.

2. It listed `pydantic 2.* → 1.10.13` under "Fixes lship refused to apply". That recommendation is wrong and would be harmful if actioned: the repo is on `pydantic==2.12.5`, CVE-2021-29510 affects only 1.x ranges, and CVE-2024-3772 is fixed in **2.4.0** — which 2.12.5 is well past. lship took the `first_patched_version` from the *1.x* branch of a multi-branch advisory and proposed a two-major downgrade to fix nothing. I replaced the bullet with a note recording why, rather than deleting it silently.

`multer` in `ansible_ai_connect_admin_portal` is the one worth pulling out separately — 1.4.5-lts.1 against **eight** CVEs, deferred as a major. It will keep reappearing in every lship body until it has its own ticket.

de1987 (approved):

PR #2184, opened 2026-09-06
deps: raise fast-uri, js-yaml and webpack-dev-middleware security floors

omaciel:
## What

Raises npm `overrides` floors so the lockfiles resolve to patched releases, and adds two that were missing entirely.

| App | Package | Floor | Resolved |
|---|---|---|---|
| `ansible_ai_connect_admin_portal` | fast-uri | `>=4.1.3` → `>=4.1.4 <5` | 4.1.3 → **4.1.4** |
| `aap_chatbot` | fast-uri | `>=4.1.3` → `>=4.1.4 <5` | 4.1.3 → **4.1.4** |
| `aap_chatbot` | js-yaml | *(none)* → `>=4.3.2 <5` | 4.3.1 → **4.3.2** |
| `ansible_ai_connect_chatbot` | js-yaml | `>=4.3.1` → `>=4.3.2 <5` | 4.3.1 → **4.3.2** |
| `ansible_ai_connect_admin_portal` | webpack-dev-middleware | *(none)* → nested `>=7.4.6 <8` | 7.4.3 → **7.4.6** |

## Why

**fast-uri and js-yaml** — both floors were declared *at* a version the advisories name as affected, so they read as deliberate security pins while the resolved tree stayed exposed. fast-uri 4.1.3 is one of the three exact versions named by CVE-2026-84394 and also sits inside the `[4.0.0, 4.1.4)` range of CVE-2026-84292. js-yaml 4.3.1 is below the 4.3.2 that fixes CVE-2026-84375; the admin portal already had 4.3.2, the two chatbot apps did not.

**webpack-dev-middleware** — 7.4.6 was released 2026-09-03 with *"reject requests that resolve outside the output directory"*, the fix for CVE-2026-76844. The advisory predates that release and the GitHub advisory record still carries **no patched version**, so this fix is easy to miss by reading the advisory alone.

## Scope

All floors are capped below the next major, matching the convention already in these files (`qs: ">=6.16.0 <7"`). This matters: with an unbounded `>=4.3.2`, npm resolved js-yaml to **5.4.1** — a silent major bump. The cap keeps each dependency on its current major and takes only the minimum patched release.

The webpack-dev-middleware override is **nested** (`webpack-dev-server` → `webpack-dev-middleware`) rather than flat, scoped to the chain where the package actually appears.

### Transitive cost, disclosed

webpack-dev-middleware 7.4.6 requires a newer memfs, which upstream has restructured. Regeneration therefore **adds 8 `@jsonjoy.com/fs-*` packages** and moves memfs 4.43.0 → 4.71.0 plus several `@jsonjoy.com` helpers. Nothing removed, nothing downgraded.

Every entry in that cascade is marked `dev` in the lockfile (verified: 0 non-dev entries, against a control showing 1077 of 1405 entries do carry the flag). None of it reaches a shipped image — the container copies only the built static output and never `node_modules`.

## Verification

- All five targets resolve as intended.
- No downgrades in any of the three lockfiles.
- Regenerating a second time is a fixed point.


hasys (approved):
lgtm, thank you @omaciel

PR #2185, opened 2026-09-11
deps: move compression to 1.8.2 in the admin portal lockfile

omaciel:
## What

Moves the transitive `compression` pin in the admin portal lockfile from 1.8.1 to 1.8.2. One file, 8 insertions, 3 deletions, no `package.json` change.

## Why

CVE-2026-87776 / GHSA-vc2v-76pw-4v95: `compression` before 1.8.2 never destroys the zlib stream when a client aborts a compressed response, so each aborted response leaks native zlib memory until the process runs out. Fixed upstream in 1.8.2, published 2026-09-11.

**No runtime impact on this project.** `compression` is reached only as a transitive of `webpack-dev-server`, which is a `devDependency`:

- `scripts/start.js` references `webpack-dev-server` once, out of 18 `require()` calls
- `scripts/build.js` references it zero times, out of 15

`npm run build` runs `build.js`, so the middleware is never loaded and is not part of the production bundle. The reason to bump it anyway is that dependency scanners read the whole lockfile: the `npm_audit` workflow runs `npm audit` with no `--omit=dev`, so this would start failing on unrelated PRs once the advisory reaches npm's audit database. It had not propagated at the time of writing.

## Scope

- `webpack-dev-server` 5.2.6 requires `compression` `^1.8.1`, which already admits 1.8.2 — no `overrides` entry needed.
- `destroy` 1.2.0, the single new dependency 1.8.2 introduces (it is what destroys the stream), is already present in the lockfile.
- Package count unchanged at 1405.


SiddharthRajaraman (approved):
lgtm

PR #2186, opened 2026-09-11
deps: add a colord security floor to the admin portal overrides

omaciel:
## What

Adds `"colord": ">=2.9.4 <3"` to the admin portal's existing `overrides` block and lets the lockfile resolve against it. Two files, 5 changed lines total.

## Why

CVE-2026-85062 / GHSA-2wm5-q62r-hmrv — colord before 2.9.4 rejects oversized malformed colour strings slowly enough to be a denial-of-service vector. Moderate. Published **2026-09-08**.

**This is pre-existing, not new.** `main` carries colord 2.9.3 today. The advisory landed one day after the Admin Portal npm-audit job last ran clean (last green run 2026-09-07), so the first job to run afterwards reports it — which surfaces as a red check on whichever PR happens to run first, reading as that PR's problem when it is the branch's.

That is a gate-scheduling artefact, not a code change. The audit job runs only on push and pull_request for this path, so a newly published advisory cannot surface until someone touches the directory. Worth fixing separately; noted below.

## Scope

- colord is **transitive-only** — reached from `postcss-colormin` and `postcss-minify-gradients`, which both declare `^2.9.1`. Nothing imports it directly.
- That range already admits the fix, so a lockfile refresh alone would have resolved it. The floor is added anyway because it is how this project records security floors (24 entries already, e.g. `fast-uri`, `qs`, `js-yaml`), and it stops a future regeneration dropping back below the fix.
- Bounded at `<3` to match the neighbouring entries rather than admitting an unreviewed major. **Resolution moves to 2.10.0**, the current release inside that bound, not to 2.9.4.
- Verified locally: `audit-ci --low` goes from failing on this advisory to `Passed npm security audit`.

## Related

- ansible/ansible-ai-connect-service#2185 (compression 1.8.2) is the PR whose run surfaced this. Kept separate so that one stays a single-package change.
- The same advisory wave produced GHSA-82fw-gwwq-j7x9 against `vitest`/`@vitest/mocker` 3.2.7 in both chatbot projects. That fix is 4.1.11, a major bump of the test framework, so it needs its own decision rather than a floor.


SiddharthRajaraman (approved):
lgtm

omaciel:
> **AI-assisted review** (opencode). Follow-up to #2185.

**Content: verified.**
- Adds `colord: ">=2.9.4 <3"` as the 25th entry in the admin portal `overrides` block. `colord` is transitive-only (not a direct dep), so the floor is safe.
- The `<3` cap bounds it to the 2.x line, matching house style (`fast-uri: ">=4.1.4 <5"`, `qs: ">=6.16.0 <7"`). Resolution lands on **2.10.0**, not 2.9.4 — confirmed in the lockfile.
- Lockfile diff is a single package: `colord 2.9.3 → 2.10.0`, no additions or removals.

**Gates:** all green. The `Admin Portal - npm-audit` lint job that failed on #2185 (GHSA-2wm5-q62r-hmrv) now passes, so the floor resolves the flagged vulnerability. Konflux build + enterprise contract green.

Remediation for the pre-existing advisory that ambushed #2185; the compression bump itself was a clean one-package change and unaffected. (Reviewing as a comment, not an approval — this is my own PR.)

